PR #29, opened 2020-10-05
Fix powerlevel10k_classic background

nwykes:


JanDeDobbeleer:
@nwykes thanks for this one, can't believe I missed this.
For completeness, can you change it to the following?

```json
"foreground": "#546E7A",
"background": "transparent",
```

That way we make the background explicit (same effect, but better for people trying to see what it does).
As for your validation error, the commits have to follow the conventional commits example, meaning for this one specifically it needs to start with `fix: powerlevel10k_classic background` (mind the case). I still need to add a contributing guide, although the action itself does list that context.

Also, you might want to rebase as well, I'm merging another PR in first.

Thanks!

nwykes:
Updated to use transparent and rebased.  Thanks - I had missed the commitlint check.

JanDeDobbeleer (approved):

PR #35, opened 2020-10-07
feat: os segment

nwykes:
### Prerequisites

### Description

Add osicon segment for displaying an OS specific icon


segment.go
```diff
@@ -56,6 +56,8 @@ const (
 	ShellInfo SegmentType = "shell"
 	//Node writes which node version is currently active
 	Node SegmentType = "node"
+	//OS write os specific icon
+	OsIcon SegmentType = "osicon"
```
JanDeDobbeleer: Can we rename this to Os? As people will be able to set any string (which could also be plain `"windows"` for example), it feels more natural that way. I would leave `WindowsIcon` etc as is as the other segments use the same context and most likely people will be adding an icon.

segment_osicon_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOsIcon(t *testing.T) {
+	env := new(MockedEnvironment)
+	env.On("getRuntimeGOOS", nil).Return("windows")
+	osicon := &osicon{env: env}
```
JanDeDobbeleer: You can also assign properties to validate the exact icon, so differ between set and not set.
See [here](https://github.com/JanDeDobbeleer/oh-my-posh3/blob/main/segment_exit_test.go#L30) for example.

JanDeDobbeleer (changes requested):

segment.go
nwykes: I originally had it as "os", but had a naming collision with the os pkg.  I'll try changing the segment name to os and call the internal struct something else.

segment_osicon_test.go
nwykes: wil do, thanks.

segment.go
JanDeDobbeleer: That's the best solution indeed 👍🏻

JanDeDobbeleer:
@nwykes first off, thanks for the contributions and coffee. You made my day yesterday yesterday and today with that 🙏🏻

A final request, can you rebase to main and remove that merge commit? I have a personal dislike against those things 😅

nwykes:
> @nwykes first off, thanks for the contributions and coffee. You made my day yesterday yesterday and today with that 🙏🏻
> 
> A final request, can you rebase to main and remove that merge commit? I have a personal dislike against those things 😅

least I can do, since I'm now using it in powershell in windows and zsh in Mac and linux everywhere, replacing my old powerlevel10k setup.  More coffees to come I hope.

Still working on the changes, but I'll get it rebased before submitting again.

PR #39, opened 2020-10-07
feat: add os segment

nwykes:
### Prerequisites

### Description

Add new os segment that display an OS specific string which defaults to an icon.

### DIFF
--- a/segment_os.go
+++ b/segment_os.go
@@ -1,0 +1,38 @@
+package main
+
+type osInfo struct {
+	props          *properties
+	env            environmentInfo
+}
+
+const (
+	//Macos the string to use for macOS
+	macOS Property = "macos"
+	//LinuxIcon the string to use for linux
+	Linux Property = "linux"
+	//WindowsIcon the icon to use for windows
+	Windows Property = "windows"
+)
+
+func (n *osInfo) enabled() bool {
+	return true
+}
+
+func (n *osInfo) string() string {
+	goos := n.env.getRuntimeGOOS()
+	switch goos {
+	case "windows":
+		return n.props.getString(Windows, "\uE62A")
+	case "darwin":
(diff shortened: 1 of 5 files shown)


nwykes:
Sorry for the new PR i totally botched the history on the other one.

JanDeDobbeleer:
> Sorry for the new PR i totally botched the history on the other one.

Shit happens, tell me when you're ready!

segment_os.go
```diff
@@ -0,0 +1,38 @@
...
+	env            environmentInfo
+}
+
+const (
+	//Macos the string to use for macOS
+	macOS Property = "macos"
+	//LinuxIcon the string to use for linux
+	Linux Property = "linux"
+	//WindowsIcon the icon to use for windows
+	Windows Property = "windows"
```
JanDeDobbeleer, line 14: this no longer references the `icon` notion, on propose?

nwykes: That's correct.  It defaults to the icon for the property, but I wanted to remove "icon" from the name so folks wouldn't think it could only be an icon.  Maybe that makes it more obvious that "win" would be a valid value, and not just an icon.

nwykes:
If you're good with the property name change, I think this is good to go as a first cut.  I want to add the WSL support to it, as well as more distro icon support instead of just the generic linux Tux, but those can come in future PRs I think.

segment_os.go
JanDeDobbeleer: Alright. Makes sense. I might have to update that everywhere to align (could break custom themes).

JanDeDobbeleer:
Let's rebase and ship this one.

segment_os.go
nwykes: are those in use somewhere already?  I don't want to break anything.

PR #62, opened 2020-10-12
feat: add wsl capability and linux distro icons to os segment

### LINKED ISSUE #28: Add WSL indication segment
Goal: a segmnet that shows you're inside the WSL. There should be an evironment variable we can use: `WSL_DISTRO_NAME`. If that one's set, you know you're inside of the WSL.

nwykes:
### Prerequisites

### Description

adds wsl handling to the OS Segment
adds linux distro icons to OS Segment

Resolves #28


environment.go
```diff
@@ -110,6 +112,15 @@ func (env *environment) getRuntimeGOOS() string {
 	return runtime.GOOS
 }
 
+func (env *environment) getPlatform() string {
+	if runtime.GOOS == "windows" {
+		return "windows"
+	}
+	p, _, _, _ := host.PlatformInformation()
```
JanDeDobbeleer, line 129: Is there a reason we do this rather than `lsb_release -a` and parse the content?
I feel like this is overkill to include a full lib for limited functionality

JanDeDobbeleer: Never min, figured it out. This will have to be done.

segment_os.go
```diff
@@ -26,12 +34,64 @@ func (n *osInfo) string() string {
...
+				n.props.getString(WSLSeparator, " - "),
+				n.props.getString(Linux, getLinuxIcon(p)))
+		}
+		return n.props.getString(Linux, getLinuxIcon(p))
 	default:
 		return ""
 	}
 }
 
+func getLinuxIcon(p string) string {
```
JanDeDobbeleer: I would remove this. We have idea if the icon set here will be available in the patched font the user want to use. And adding each and everyone of these as a `Property` is a bit of a bridge too far imho.

nwykes: this assumes nerd fonts, which most of the icons seem to be geared for.  we could get even fancier and do something like p10k where there's a global config setting which type of font is used and sets the default icons appropriately.  https://github.com/romkatv/powerlevel10k/blob/master/internal/icons.zsh

nwykes: I added the distro icon since I thought that had been discussed previously on one of the other PRs

environment.go
nwykes: using that library opens up some other possibilities for other segments - like number of running jobs or memory in use.

segment_os.go
JanDeDobbeleer: If we do go down this path, I feel like we should add them to the Properties as well otherwise you can never override.

On top of that, if so, it might be a good idea to default to icons everywhere and propose a nerd font patched font for max enjoyment.

JanDeDobbeleer: For completeness, that last remark shouldn't be resolved here 😁

nwykes: how about a use simple font vs nerd font setting?  if simple is used, then it defaults to a single linux icon/string property and if nerd font it branches into the distro specific icons?

nwykes: because i agree, having a property for every distro is overload.

JanDeDobbeleer: > how about a use simple font vs nerd font setting?  if simple is used, then it defaults to a single linux icon/string property and if nerd font it branches into the distro specific icons?  

We'd have to be sure all nerd fonts render it consistently then. You're going to slap me, but the more I think about it, the more I feel like adding then as a Property is the least annoying thing to do going forward. And still default to the icons.

JanDeDobbeleer: @nwykes did I demotivate you?

nwykes: lol not at all. Work got crazy and I haven't been able to finish this up.  I've converted the distro icons to properties and just need to put a bow on it and resubmit.

docs/docs/segment-os.md
```diff
@@ -25,6 +25,8 @@ Display OS specific info. Defaults to Icon.
 
 ## Properties
 
-- macos: `string` - the string to use for macOS - defaults to macOS icon
-- linux: `string` - the icon to use for Linux - defaults to Linux icon
-- windows: `string` - the icon to use for Windows - defaults to Windows icon
+- macos: `string` - the string/icon to use for macOS - defaults to macOS icon
+- linux: `string` - the string/icon to use for Linux - defaults to Linux icon
+- windows: `string` - the string/icon to use for Windows - defaults to Windows icon
+- wsl: `string` - the string/icon to use for WSL - defaults to WSL
```
JanDeDobbeleer: can we change this to ``defaults to `WSL` ``, same for adding the default below.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@nwykes this looks ready for prime time, no? If you rebase I'll publish ;-)

nwykes:
> @nwykes this looks ready for prime time, no? If you rebase I'll publish ;-)

been double checking on a few different distros.  I'll try to rebase tonight my time.

nwykes:
Should be ready to go.  sorry for the delay.

PR #75, opened 2020-10-16
feat: add terraform workspace segment

matsest:
### Prerequisites

### Description

This adds support for showing the [current terraform workspace name](https://www.terraform.io/docs/commands/workspace/show.html).

Inspired by a bash prompt I've used to show the current workspace:

```
function terraform_prompt()
  {
     if [ -d .terraform ]; then
       workspace="$(command terraform workspace show 2>/dev/null)"
       echo " (${workspace})"
     fi
  }
```

### DIFF
--- a/docs/docs/segment-terraform.md
+++ b/docs/docs/segment-terraform.md
@@ -1,0 +1,25 @@
+---
+id: terraform
+title: Terraform Context
+sidebar_label: Terraform
+---
+
+## What
+
+Display the currently active Terraform Workspace name.
+
+Note:
+- Will need a terraform binary in your PATH
+- Will only be displayed in directories that contain a `.terraform` subdirectory
+
+## Sample Configuration
+
+```json
+{
+  "type": "terraform",
+  "style": "powerline",
+  "powerline_symbol": "\uE0B0",
+  "foreground": "#000000",
+  "background": "#ebcc34"
+}
+```
--- a/segment_terraform.go
+++ b/segment_terraform.go
@@ -1,0 +1,24 @@
+package main
+
+type terraform struct {
+	props         *properties
+	env           environmentInfo
+	workspaceName string
+}
+
+func (tf *terraform) string() string {
+	return tf.workspaceName
+}
+
+func (tf *terraform) init(props *properties, env environmentInfo) {
+	tf.props = props
+	tf.env = env
+}
+
+func (tf *terraform) enabled() bool {
+	if !tf.env.hasCommand("terraform") || !tf.env.hasFolder(".terraform") {
+		return false
+	}
+	tf.workspaceName, _ = tf.env.runCommand("terraform", "workspace", "show")
+	return true
+}
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -30,6 +30,7 @@ module.exports = {
         "session",
         "shell",
         "spotify",
+        "terraform",
         "text",
         "time",
       ]
--- a/segment.go
+++ b/segment.go
@@ -69,6 +69,8 @@ const (
 	Kubectl SegmentType = "kubectl"
 	//Dotnet writes which dotnet version is currently active
 	Dotnet SegmentType = "dotnet"
+	//Terraform writes the terraform workspace we're currently in
+	Terraform SegmentType = "terraform"
 	//Powerline writes it Powerline style
 	Powerline SegmentStyle = "powerline"
 	//Plain writes it without ornaments
@@ -126,6 +128,7 @@ func (segment *Segment) mapSegmentWithWriter(env environmentInfo) error {
 		Az:        &az{},
 		Kubectl:   &kubectl{},
 		Dotnet:    &dotnet{},
+		Terraform: &terraform{},
 	}
 	if writer, ok := functions[segment.Type]; ok {
 		props := &properties{
--- a/segment_terraform_test.go
+++ b/segment_terraform_test.go
@@ -1,0 +1,64 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type terraformArgs struct {
+	hasTfCommand  bool
+	hasTfFolder   bool
+	workspaceName string
+}
+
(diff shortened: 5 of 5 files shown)


matsest:
The existence of the .terraform dir and test for it is the WIP part I've identified and not too sure about (not very familiar with golang)

Edit: I think this is working now, after testing locally and running the gh actions.  But very open to improvement suggestions :)

docs/docs/segment-terraform.md
```diff
@@ -0,0 +1,24 @@
...
+
+Display the currently active Terraform Workspace name.
+
+## Sample Configuration
+
+```json
+{
+  "type": "terraform",
+  "style": "powerline",
+  "powerline_symbol": "",
```
JanDeDobbeleer: Can you change this to `"powerline_symbol": "\uE0B0",`, I changed all docs to use UTF32 text representations rather than icons which can't always be rendered

segment_terraform.go
```diff
@@ -0,0 +1,24 @@
...
+	return tf.workspaceName
+}
+
+func (tf *terraform) init(props *properties, env environmentInfo) {
+	tf.props = props
+	tf.env = env
+}
+
+func (tf *terraform) enabled() bool {
+	if !(tf.env.hasCommand("terraform") && tf.env.hasFolder(".terraform")) {
```
JanDeDobbeleer: Can we adjust this to `!tf.env.hasCommand("terraform") && !tf.env.hasFolder(".terraform")`, when I read this I didn't see the initial `!` and I'm sure I won't be the only one :-)

segment_terraform.go
```diff
@@ -0,0 +1,24 @@
...
+		return false
+	}
+	tf.workspaceName = tf.env.runCommand("terraform", "workspace", "show")
```
JanDeDobbeleer: Can this fail when the previous conditions passes?

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@matsest feel free to reach out for help when you need it, I'm here to assist. Great work overall, looking forward to merging this 👍🏻

segment_terraform.go
matsest: Done :)

segment_terraform.go
matsest: Hm, no. The only requirement for the command to run is to have the terraform binary. It can run in all directories, even those who have nothing to do with Terraform. Even in directories that have not been subject to "terraform init" (i.e. they will not have a .terraform subdir) the command will output "default".

But it does not make sense to show this output in directories without the .terraform subdir, so that's why that check is added.

segment_terraform.go
matsest: Actually, it seems like if !(A && B) and  if !A && !B does not evaluate in the same manner.

With the first implementation (how it was) it did not show the prompt if there was not a .terraform directory, but in the suggested implementation it actually shows the prompt despite not having a .terraform directory. I've tested it with the current build.

Is this a go-ism? @JanDeDobbeleer  Agree that for readability the suggested implementation is better.

Consider this example where the third statement is equivalent to what happens now:
```
func main() {
        if !false && !false {
                fmt.Println("!false && !false works")
        }

        if !(false && false) {
                fmt.Println("!(false && false) works")
        }

        if !true && !false {
                fmt.Println("!true && !false works") <------ this output will not be printed.
        }

        if !(true && false) {
                fmt.Println("!(true && false) works")
        }
}
``` 

This is however fixable with an || statement.. (if !A || !B). But I think that is a bit less clear. Let me know what you think.

segment_terraform.go
matsest: A suggestion for how to include this test into segment_terraform_test.go (actually evaluate both cases together) would be of great help! :)

segment_terraform.go
matsest: A suggestion for how to include this test into segment_terraform_test.go (actually evaluate both cases together) would be of great help! :) I started - see b413a7c

docs/docs/segment-terraform.md
```diff
@@ -0,0 +1,28 @@
...
+
+```json
+{
+  "type": "terraform",
+  "style": "powerline",
+  "powerline_symbol": "\uE0B0",
+  "foreground": "#000000",
+  "background": "#ebcc34",
+  "properties": {
+    "prefix": " "
```
JanDeDobbeleer: This is the default so that can be omitted :-)

segment_terraform.go
JanDeDobbeleer: It should be `!tf.env.hasCommand("terraform") || !tf.env.hasFolder(".terraform")`, when either of one condition isn't met, it's disabled.

JanDeDobbeleer:
2 comments and a general remark, can you squash all these commits into 1? It's small enough to be in a single commit context.

matsest:
@JanDeDobbeleer Sounds good! Will fix this when I'm back home tomorrow evening

docs/docs/segment-terraform.md
matsest: Which part? Remove the properties :  {} statement? @JanDeDobbeleer

JanDeDobbeleer: No, prefix defaults to `" "`

matsest:
Thanks for the helpful review. I've gone through your review comments, rebased and squashed the PR into one commit. Hope it looks good now @JanDeDobbeleer  :)

docs/docs/segment-terraform.md
```diff
@@ -0,0 +1,27 @@
...
+## Sample Configuration
+
+```json
+{
+  "type": "terraform",
+  "style": "powerline",
+  "powerline_symbol": "\uE0B0",
+  "foreground": "#000000",
+  "background": "#ebcc34",
+  "properties": {
```
JanDeDobbeleer: in this case this can be removed entirely :-)

JanDeDobbeleer:
> Hope it looks good now

Only need to remove the properties in the docs and it can be merged!

PR #76, opened 2020-10-16
feat: dotnet segment for .NET SDK display

tillig:
### Prerequisites

### Description

New segment for .NET SDK version (or unsupported version) display.

Includes update for handling command execution errors so segments
can act differently based on exit codes. Using a custom error
type to make it testable rather than passing the OS error directly
to the segment.

### DIFF
--- a/docs/docs/segment-dotnet.md
+++ b/docs/docs/segment-dotnet.md
@@ -1,0 +1,31 @@
+---
+id: dotnet
+title: Dotnet
+sidebar_label: Dotnet
+---
+
+## What
+
+Display the currently active .NET SDK version.
+
+## Sample Configuration
+
+```json
+{
+  "type": "dotnet",
+  "style": "powerline",
+  "powerline_symbol": "",
+  "foreground": "#000000",
+  "background": "#00ffff",
+  "properties": {
+    "prefix": " \uE77F "
+  }
+}
+```
+
+## Properties
+
+- display_version: `boolean` - display the active version or not; useful if all you need is an icon indicating `dotnet`
+  is present - defaults to `true`
+- unsupported_version_icon: `string` - text/icon that is displayed when the active .NET SDK version (e.g., one specified
+  by `global.json`) is not installed/supported - defaults to `\u2327` (X in a rectangle box)
--- a/segment_dotnet_test.go
+++ b/segment_dotnet_test.go
@@ -1,0 +1,81 @@
+package main
+
+import (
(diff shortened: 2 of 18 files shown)


docs/docs/segment-dotnet.md
```diff
@@ -0,0 +1,31 @@
...
+
+```json
+{
+  "type": "dotnet",
+  "style": "powerline",
+  "powerline_symbol": "",
+  "foreground": "#000000",
+  "background": "#00ffff",
+  "properties": {
+    "prefix": "  "
```
JanDeDobbeleer: Can you add the textual utf32 representation of the icon? Like `\uF321` for example.

docs/docs/segment-dotnet.md
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+```
+
+## Properties
+
+- display_version: `boolean` - display the active version or not; useful if all you need is an icon indicating `dotnet`
+  is present - defaults to `true`
+- unsupported_version_icon: `string` - text/icon that is displayed when the active .NET SDK version (e.g., one specified
+  by `global.json`) is not installed/supported - defaults to `⌧`
```
JanDeDobbeleer: Same comment as above, use the UTF32 character text representation

segment_dotnet_test.go
```diff
@@ -0,0 +1,83 @@
...
+func bootStrapDotnetTest(args *dotnetArgs) *dotnet {
+	env := new(MockedEnvironment)
+	env.On("hasCommand", "dotnet").Return(args.enabled)
+	if args.unsupported {
+		err := &commandError{exitCode: 145}
+		env.On("runCommand", "dotnet", []string{"--version"}).Return("", err)
+	} else {
+		env.On("runCommand", "dotnet", []string{"--version"}).Return(args.version, nil)
+	}
+
```
JanDeDobbeleer: not a fan of the spaces here myself :-)

JanDeDobbeleer (changes requested):

docs/docs/segment-dotnet.md
tillig: I can throw another PR in to fix up the other docs using extended characters if that's interesting.

JanDeDobbeleer:
Can you rebase? I can still merge this today (beware, lots of changes)

tillig:
Already rebased! I think I caught it like a minute before your comment. ;)

PR #125, opened 2020-11-05
feat: Spotify segment for windows

lnu:
Windows support added for spotify segment. AutoHotkey is used to get the informations. AutoHoykey script added to the project. Should be deployed in the same folder as the oh-my-posh executable.

Preview:
ohmyposh_spotify_windows


JanDeDobbeleer:
I'm reviewing this evening. Love this!

JanDeDobbeleer:
2 pointers I want to address in 1 comment as they'll restructure everything quite a bit.

- The layout here is very pragmatic but creates a bit of chaos when extending toward other platforms
- We need to use `environmentInfo` to interact with the system, otherwise adding tests becomes cumbersome

I'll address the first and solve the second one inline.
To solve the cross platform mess, golang has the ability to compile only for the selected platform.
We can use build tags to create logic that only runs on multiple platforms, and create platform specific files that run on a single platform (Windows, Darwin).

**segment-spotify.go**: contains all shared logic

```golang
package main

import (
	"fmt"
)

type spotify struct {
	props  *properties
	env    environmentInfo
	status string
	artist string
	track  string
}

const (
	//PlayingIcon indicates a song is playing
	PlayingIcon Property = "playing_icon"
	//PausedIcon indicates a song is paused
	PausedIcon Property = "paused_icon"
	//StoppedIcon indicates a song is stopped
	StoppedIcon Property = "stopped_icon"
	//TrackSeparator is put between the artist and the track
	TrackSeparator Property = "track_separator"
)

func (s *spotify) string() string {
	icon := ""
	switch s.status {
	case "stopped":
		icon = s.props.getString(StoppedIcon, "\uF04D ")
	case "paused":
		icon = s.props.getString(PausedIcon, "\uF8E3 ")
	case "playing":
		icon = s.props.getString(PlayingIcon, "\uE602 ")
	}
	separator := s.props.getString(TrackSeparator, " - ")
	return fmt.Sprintf("%s%s%s%s", icon, s.artist, separator, s.track)
}

func (s *spotify) init(props *properties, env environmentInfo) {
	s.props = props
	s.env = env
}

```
**segment_spotify_others.go**: contains the logic for everything not supported
Notice the build tags to exclude building this for Windows and Darwin.

```golang
// +build !darwin
// +build !windows

package main

func (s *spotify) enabled() bool {
	return false
}

```

**segment_spotify_darwin.go**: Darwin specific logic

```golang
package main

func (s *spotify) enabled() bool {
	var err error
	// Check if running
	running := s.runAppleScriptCommand("application \"Spotify\" is running")
	if running == "false" || running == "" {
		return false
	}
	s.status = s.runAppleScriptCommand("tell application \"Spotify\" to player state as string")
	if err != nil {
		return false
	}
	if s.status == "stopped" {
		return false
	}
	s.artist = s.runAppleScriptCommand("tell application \"Spotify\" to artist of current track as string")
	s.track = s.runAppleScriptCommand("tell application \"Spotify\" to name of current track as string")
	return true
}

func (s *spotify) runAppleScriptCommand(command string) string {
	val, _ := s.env.runCommand("osascript", "-e", command)
	return val
}

```

**segment-spotify_windows.go**: Windows specific logic

- In here I also addressed the second comment, use `s.env` to call system commands (just like the Darwin one does).
- I would also remove the autohotkey script from the codebase and add that to the docs, allowing the user to provide us with the location to the script instead.

```golang
package main

import (
	"strings"
)

const (
	//AutoHotkeyScript script to read spotify windows title
	AutoHotkeyScript Property = "autohotkey_script"
)

func (s *spotify) enabled() bool {
	if !s.env.hasCommand("AutoHotkey") {
		return false
	}
	spotifyWindowTitle, err := s.env.runCommand("AutoHotkey", s.props.getString(AutoHotkeyScript, ""))
	if err != nil || spotifyWindowTitle == "" {
		return false
	}
	if !strings.Contains(spotifyWindowTitle, " - ") {
		s.status = "stopped"
		return true
	}
	infos := strings.Split(spotifyWindowTitle, " - ")
	s.artist = infos[0]
	// remove first element
	s.track = strings.Join(infos[1:], " - ")
	s.status = "playing"
	return true
}

```

JanDeDobbeleer (changes requested):
2 pointers I want to address in 1 comment as they'll restructure everything quite a bit.

- The layout here is very pragmatic but creates a bit of chaos when extending toward other platforms
- We need to use `environmentInfo` to interact with the system, otherwise adding tests becomes cumbersome

I'll address the first and solve the second one inline.
To solve the cross platform mess, golang has the ability to compile only for the selected platform.
We can use build tags to create logic that only runs on multiple platforms, and create platform specific files that run on a single platform (Windows, Darwin).

**segment-spotify.go**: contains all shared logic

```golang
package main

import (
	"fmt"
)

type spotify struct {
	props  *properties
	env    environmentInfo
	status string
	artist string
	track  string
}

const (
	//PlayingIcon indicates a song is playing
	PlayingIcon Property = "playing_icon"
	//PausedIcon indicates a song is paused
	PausedIcon Property = "paused_icon"
	//StoppedIcon indicates a song is stopped
	StoppedIcon Property = "stopped_icon"
	//TrackSeparator is put between the artist and the track
	TrackSeparator Property = "track_separator"
)

func (s *spotify) string() string {
	icon := ""
	switch s.status {
	case "stopped":
		icon = s.props.getString(StoppedIcon, "\uF04D ")
	case "paused":
		icon = s.props.getString(PausedIcon, "\uF8E3 ")
	case "playing":
		icon = s.props.getString(PlayingIcon, "\uE602 ")
	}
	separator := s.props.getString(TrackSeparator, " - ")
	return fmt.Sprintf("%s%s%s%s", icon, s.artist, separator, s.track)
}

func (s *spotify) init(props *properties, env environmentInfo) {
	s.props = props
	s.env = env
}

```
**segment_spotify_others.go**: contains the logic for everything not supported
Notice the build tags to exclude building this for Windows and Darwin.

```golang
// +build !darwin
// +build !windows

package main

func (s *spotify) enabled() bool {
	return false
}

```

**segment_spotify_darwin.go**: Darwin specific logic

```golang
package main

func (s *spotify) enabled() bool {
	var err error
	// Check if running
	running := s.runAppleScriptCommand("application \"Spotify\" is running")
	if running == "false" || running == "" {
		return false
	}
	s.status = s.runAppleScriptCommand("tell application \"Spotify\" to player state as string")
	if err != nil {
		return false
	}
	if s.status == "stopped" {
		return false
	}
	s.artist = s.runAppleScriptCommand("tell application \"Spotify\" to artist of current track as string")
	s.track = s.runAppleScriptCommand("tell application \"Spotify\" to name of current track as string")
	return true
}

func (s *spotify) runAppleScriptCommand(command string) string {
	val, _ := s.env.runCommand("osascript", "-e", command)
	return val
}

```

**segment-spotify_windows.go**: Windows specific logic

- In here I also addressed the second comment, use `s.env` to call system commands (just like the Darwin one does).
- I would also remove the autohotkey script from the codebase and add that to the docs, allowing the user to provide us with the location to the script instead.

```golang
package main

import (
	"strings"
)

const (
	//AutoHotkeyScript script to read spotify windows title
	AutoHotkeyScript Property = "autohotkey_script"
)

func (s *spotify) enabled() bool {
	if !s.env.hasCommand("AutoHotkey") {
		return false
	}
	spotifyWindowTitle, err := s.env.runCommand("AutoHotkey", s.props.getString(AutoHotkeyScript, ""))
	if err != nil || spotifyWindowTitle == "" {
		return false
	}
	if !strings.Contains(spotifyWindowTitle, " - ") {
		s.status = "stopped"
		return true
	}
	infos := strings.Split(spotifyWindowTitle, " - ")
	s.artist = infos[0]
	// remove first element
	s.track = strings.Join(infos[1:], " - ")
	s.status = "playing"
	return true
}

```

lnu:
Thanks for your comments. I did all the changes.
I have a better understanding of the soft.
So the script path is passed in the json config file in the spotify section and the build has to be done using `-tags xxx` .

JanDeDobbeleer:
The build is done on each platform making the need for tags unnecessary. This should build correctly out of the box for the current platform when using `go build`

docs/docs/segment-spotify.ahk
```diff
@@ -0,0 +1,7 @@
+DetectHiddenWindows, On
```
JanDeDobbeleer: I see I wasn't 100% clear on this, we can move this context to ` docs/docs/segment-spotify.md` in a section that lists the Windows specific settings like `autohotkey_script`.

docs/docs/segment-spotify.md
```diff
@@ -6,11 +6,14 @@ sidebar_label: Spotify
...
-Show the currently playing song in the Spotify MacOS client. Only available on MacOS for obvious reasons.
-Be aware this can make the prompt a tad bit slower as it needs to get a response from the Spotify player using Applescript.
+Show the currently playing song in the Spotify MacOS/Windows client.
+AutoHotkey is needed for the windows client. You also have to add the
+`autohotkey_script` property with the path to this [script](./segment-spotify.ahk) in the spotify section of your config file.
+Be aware this can make the prompt a tad bit slower as it needs to get a response from the Spotify player using Applescript/AutoHotkey.
 
 ## Sample Configuration
 
+### darwin
```
JanDeDobbeleer: I would keep the JSON sample the same, and only differentiate below under `Properties` eg:

```markdown
### Properties

- playing_icon: `string` - text/icon to show when playing - defaults to `\uE602 `
- paused_icon: `string` - text/icon to show when paused - defaults to `\uF8E3 `
- stopped_icon: `string` - text/icon to show when paused - defaults to `\uF8E3 `
- track_separator: `string` - text/icon to put between the artist and song name - defaults to ` - `

### Windows
- autohotkey_script: `string` - path to the autohotkey script - defaults to `""`

//TODO: add autohotkey explanation and script here
```

segment_spotify others.go
```diff
@@ -0,0 +1,8 @@
+// +build !darwin
```
JanDeDobbeleer: this file needs to be renamed to `segment_spotify_others.go `, notice the underscore.

JanDeDobbeleer (changes requested):
We can also add tests that are specific for Windows.
You can create a file called `segment_spotify_test_windows.go` and look at how to validate the logic in the `segment_spotify_test.go` file (which needs to be renamed to `segment_spotify_test_darwin.go` due to the changes we did for MacOS.

docs/docs/segment-spotify.ahk
lnu: My bad. I did it thiniking the link could be downloaded directly from the doc but it's already the case if I put it directly in the page.

segment_spotify others.go
lnu: Sorry... I was half coding half watching football :sweat_smile:

lnu:
I added some unit tests. I don't know if it's what you were expecting so tell me if it suits you.

JanDeDobbeleer:
I'll go through it later today. Thanks for all the efforts, much appreciated 🙏🏻

docs/docs/segment-spotify.md
```diff
@@ -30,4 +30,21 @@ Be aware this can make the prompt a tad bit slower as it needs to get a response
...
 - playing_icon: `string` - text/icon to show when playing - defaults to `\uE602 `
 - paused_icon: `string` - text/icon to show when paused - defaults to `\uF8E3 `
+- paused_icon: `string` - text/icon to show when paused - defaults to `\uF8E3  `
 - track_separator: `string` - text/icon to put between the artist and song name - defaults to ` - `
+
+### Windows
+
+- autohotkey_script: `string` - path to the autohotkey script - defaults to `""`
+
+  The script content:
```
JanDeDobbeleer, line 45: Maybe we need to extend this explanation a bit, and add a link to Autohotkey.

segment_spotify_test.go
```diff
@@ -6,6 +8,38 @@ import (
...
+	env.On("string", nil).Return(expected)
+	s := &spotify{
+		artist: "Candlemass",
+		track:  "Spellbreaker",
+		status: "playing",
+	}
+	assert.Equal(t, expected, s.string())
+}
+
+func TestSpotifyPausedSong(t *testing.T) {
```
JanDeDobbeleer: I'm going to draft a few tests to give you an idea how to leverage `MockedEnvironment`, give me a day :-)

segment_spotify_windows_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestSpotifyWindowsEnabledWithAutoHotkeyAndSpotifyPlaying(t *testing.T) {
+	env := new(MockedEnvironment)
+	props := &properties{}
+	env.On("hasCommand", "AutoHotkey").Return(true)
+	env.On("runCommand", "AutoHotkey", []string{""}).Return("Candlemass - Spellbreaker", nil)
+	s := &spotify{
+		env:   env,
+		props: props,
+	}
+	expected := &spotify{
```
JanDeDobbeleer: I expect the properties to be set when running `s.enabled()`, not up front (that might even give a positive test when in fact it might not work correctly).

segment_spotify_test.go
lnu: There's no purpose for the `MockedEnvironment` to be here in that case :confused:

JanDeDobbeleer: Yes, there is. The second comment clarifies the purpose. You want to validate the logic, meaning control whatever `MockedEnvironment` outputs and see if everything works as designed.

segment_spotify_windows_test.go
lnu: I checked the other unit tests and tried to do the same.

segment_spotify_test.go
lnu: In this case, I check the return of the `string` function which has no dependencies on other metod calls. What would be the purpose of the ´MockeEnvironment´ in that case?

lnu:
I think we can get rid of the AutoHotkey dependency. I did a poc only using win32 api call and it works.  
What do you think?

docs/docs/segment-spotify.md
```diff
@@ -30,4 +30,26 @@ Be aware this can make the prompt a tad bit slower as it needs to get a response
 
 - playing_icon: `string` - text/icon to show when playing - defaults to `\uE602 `
 - paused_icon: `string` - text/icon to show when paused - defaults to `\uF8E3 `
+- paused_icon: `string` - text/icon to show when paused - defaults to `\uF8E3  `
 - track_separator: `string` - text/icon to put between the artist and song name - defaults to ` - `
+
+### Windows
+
+:::info AutoHotkey
+Please note [AutoHotkey](https://www.autohotkey.com/) must be installed and set in your `PATH`.
```
JanDeDobbeleer, line 39: Please add all links below like in the other files.

segment_spotify_darwin_test.go
```diff
@@ -0,0 +1,13 @@
...
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSpotifyEnabledDarwin(t *testing.T) {
+	assert.True(t, true)
```
JanDeDobbeleer, line 12: I'll extend this one ;-)

segment_spotify_test.go
JanDeDobbeleer: The string function serves little purpose to be tested other than the formatting indeed. But then the test name is incorrect ;-)

JanDeDobbeleer (changes requested):

lnu:
> I think we can get rid of the AutoHotkey dependency. I did a poc only using win32 api call and it works.
> What do you think?

I have a fully working [version](https://github.com/lnu/oh-my-posh3/blob/feature/spotify_segment_for_windows_nodependency/segment_spotify_windows.go) but now I have to refactor a little bit and update the test. I guess the api functions should be put in `Environment`? 

The calls can be seen here https://github.com/lnu/oh-my-posh3/blob/feature/spotify_segment_for_windows_nodependency/segment_spotify_windows.go

JanDeDobbeleer:
Did you push it? I'll have to check, to see what the actuals calls are. We can either add them to Environment, or add them to the Windows only Spotify code and also inject them as an interface to allow testing. Env is probably more suited as it's already there.

lnu:
> Did you push it? I'll have to check, to see what the actuals calls are. We can either add them to Environment, or add them to the Windows only Spotify code and also inject them as an interface to allow testing. Env is probably more suited as it's already there.

I was thinking the same. I moved it to `Environment` but it seems too akward and specific. I think creating an extra interface for unit test and putting it in the spotify windows only code is better.  

The calls can be seen here :https://github.com/lnu/oh-my-posh3/blob/feature/spotify_segment_for_windows_nodependency/segment_spotify_windows.go

JanDeDobbeleer:
> I think creating an extra interface for unit test and putting it in the spotify windows only code is better

Yes, specifically for Windows we can create a new interface separately to interact with Win32, it won't be limited to this case for sure. I would extract it from the spotify code so others can also easily find this when needed.

lnu:
I did some changes and moved everything to Environment(https://github.com/lnu/oh-my-posh3/tree/feature/spotify_segment_for_windows_nodependency). My knowledge in go is a bit short for the creation of the new interface.

lnu:
And for linux it can be done using the dbus interface.
https://github.com/dawidd6/go-spotify-dbus

JanDeDobbeleer:
> My knowledge in go is a bit short for the creation of the new interface.

Rather straightforward actually. You define the interface and any struct implementing all of its function automatically gets that interface compatibility. In the sample below, `Win32` is automatically also a valid `Win32Info` interface as it implements `getWindowTitle`.

```golang
type Win32Info interface {
  getWindowTitle(imageName string) (string, error)
}

type Win32 struct {}

func(w *Win32) getWindowTitle(imageName string) (string, error) {
  //whatever logic you had
}
```

JanDeDobbeleer:
@lnu shall I close this one?

PR #130, opened 2020-11-06
fix: set $Global:GitStatus when Get-GitStatus exists

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/packages/powershell/oh-my-posh/oh-my-posh.psm1
+++ b/packages/powershell/oh-my-posh/oh-my-posh.psm1
@@ -26,6 +26,12 @@ if (!$IsWindows) {
 
 function Set-PoshContext {}
 
+function Set-GitStatus {
+    if (Get-Command -Name "Get-GitStatus" -ErrorAction SilentlyContinue) {
+        $Global:GitStatus = Get-GitStatus
+    }
+}
+
 function Set-PoshPrompt {
     param(
         [Parameter(Mandatory = $false)]
@@ -70,6 +76,7 @@ function Set-PoshPrompt {
         $standardOut
         $global:LASTEXITCODE = $realLASTEXITCODE
         Remove-Variable realLASTEXITCODE -Confirm:$false
+        Set-GitStatus
     }
     Set-Item -Path Function:prompt -Value $Prompt -Force
 }


packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -70,6 +76,9 @@ function Set-PoshPrompt {
         $standardOut
         $global:LASTEXITCODE = $realLASTEXITCODE
         Remove-Variable realLASTEXITCODE -Confirm:$false
+        Start-Job -ScriptBlock {
```
lnu: I tried but it does not work with Start-Job. It works if the call is done directly like this:
``` 
    Remove-Variable realLASTEXITCODE -Confirm:$false
    Set-GitStatus
```
and performance wise, it's slower than version 2 with the same config. But the performance issue is related to the git segment because if I remove the Set-GitStatus, the v3 version is still slower than the v2.

JanDeDobbeleer: I assume this is on Windows? I noticed slower behavior on Windows with the git segement vs MacOS and Linux. Not sure why.

lnu: yes on windows. In wsl2 it's really fast, no latency.

lnu: and in wsl2 I don't have any completion if I hit tab after `git add`(even with ohmyzsh)

JanDeDobbeleer: > yes on windows. In wsl2 it's really fast, no latency.

Yup. So speeding this up is a Windows only thing. I could optionally add the `get-gitstatus` output as input and parse that for a faster experience. Could be interesting to look at.

lnu: It would be nice to have method call timings for debugging purpose when invoking segments. Could be turned on with a parameter.

JanDeDobbeleer: The truth is, this isn't an issue anywhere else but Windows. I'm more in favor of calling `Get-GitStatus` and passing the output to oh-my-posh to avoid unneeded calls.

PR #134, opened 2020-11-09
feat: Spotify segment for windows(no dependency)

lnu:
### Prerequisites

### Description

Windows support added for spotify segment. AutoHotkey is used to get the informations. AutoHoykey script added to the project. Should be deployed in the same folder as the oh-my-posh executable.

Preview:
ohmyposh_spotify_windows

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -25,6 +25,7 @@ indent_size = 2
 ; Markdown - match markdownlint settings
 [*.{md,markdown}]
 indent_size = 2
+trim_trailing_whitespace = false
 
 ; PowerShell - match defaults for New-ModuleManifest and PSScriptAnalyzer Invoke-Formatter
 [*.{ps1,psd1,psm1}]
--- a/docs/docs/segment-spotify.md
+++ b/docs/docs/segment-spotify.md
@@ -6,8 +6,8 @@ sidebar_label: Spotify
 
 ## What
 
-Show the currently playing song in the Spotify MacOS client. Only available on MacOS for obvious reasons.
-Be aware this can make the prompt a tad bit slower as it needs to get a response from the Spotify player using Applescript.
+Show the currently playing song in the Spotify MacOS/Windows client.
+Be aware this can make the prompt a tad bit slower as it needs to get a response from the Spotify player.
 
 ## Sample Configuration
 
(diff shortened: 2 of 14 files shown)


lnu:
The win32 interface has been created but I don't know exactly where to inject it. Right now I added a function in environment.go with a windows implementation and a not implemented error for others.

win32.go
```diff
@@ -0,0 +1,121 @@
...
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"syscall"
+	"unsafe"
+
+	"github.com/mitchellh/go-ps"
```
JanDeDobbeleer: We removed this dependency in favor of [gopsutil](https://github.com/shirou/gopsutil). Can you double check if we can also use `gopsutil` here?

win32.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+type win32Info interface {
```
JanDeDobbeleer: As this is now part of `environment`, the interface here is redundant. I can live with it being removed as mocking is possible via `MockedEnviroment` anyhow.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
Almost there now! 👏🏻

win32.go
lnu: I checked but there's a difference in the function that returns the processes.  
`go-ps` executes only one call to CreateToolhelp32Snapshot and then will loop through the snapshot with Process32First and Process32Next, which is very fast.
`gopsutil` executes a call to get the process pids, but when we acces the process name, it executes a call to  CreateToolhelp32Snapshot  for each process, which is very slow.
I'm still investigating there's a way to have the same behaviour with gopsutil.

JanDeDobbeleer: That would be a valid reason to use `go-ps` in that case.

lnu: dependency removed and only needed code added

JanDeDobbeleer (changes requested):
One last thing and we can merge this, can you squash all of these in a single commit? No need to see separate history for these, we can treat this as 1.

lnu:
Squash and rebase. I’ll do it tonight.

PR #160, opened 2020-11-15
feat: new ShowDebug parameter

lnu:
display segments and total timing

### Prerequisites

### Description

New `ShowDebug` parameter  used to add debug information on command line.  
Right now it displays only each segments execution time and the total execution time.  
image

image  

timings

rendering the same git folder:  
wsl2: 5ms
windows: 381ms

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -67,6 +67,9 @@ func (e *engine) renderText(text string) {
 	prefix := e.activeSegment.getValue(Prefix, " ")
 	postfix := e.activeSegment.getValue(Postfix, " ")
 	e.renderer.write(e.activeSegment.Background, e.activeSegment.Foreground, fmt.Sprintf("%s%s%s", prefix, text, postfix))
+	if *e.env.getArgs().Debug {
+		e.renderer.write(e.activeSegment.Background, e.activeSegment.Foreground, fmt.Sprintf("(%s:%s)", e.activeSegment.Type, e.activeSegment.timing))
+	}
 }
 
 func (e *engine) renderSegmentText(text string) {
@@ -107,13 +110,11 @@ func (e *engine) setStringValues(segments []*Segment) {
 	wg.Add(len(segments))
 	defer wg.Wait()
 	cwd := e.env.getcwd()
+	debug := *e.env.getArgs().Debug
 	for _, segment := range segments {
 		go func(s *Segment) {
 			defer wg.Done()
-			err := s.mapSegmentWithWriter(e.env)
-			if err == nil && !s.hasValue(IgnoreFolders, cwd) && s.enabled() {
-				s.stringValue = s.string()
-			}
+			s.setStringValue(e.env, cwd, debug)
 		}(segment)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@ type args struct {
 	Config      *string
 	Shell       *string
 	PWD         *string
+	Debug       *bool
 }
 
 func main() {
@@ -42,6 +43,10 @@ func main() {
 			"pwd",
 			"",
 			"the path you are working in"),
+		Debug: flag.Bool(
+			"debug",
+			false,
+			"Print debug information"),
 	}
 	flag.Parse()
 	env := &environment{
--- a/packages/powershell/oh-my-posh/oh-my-posh.psm1
+++ b/packages/powershell/oh-my-posh/oh-my-posh.psm1
@@ -5,6 +5,7 @@
 
 $global:PoshSettings = New-Object -TypeName PSObject -Property @{
     Theme = "$PSScriptRoot\themes\jandedobbeleer.json";
+    ShowDebug = $false
 }
 
 function Get-PoshCommand {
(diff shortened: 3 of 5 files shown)


lnu:
Do you have an idea why there's such a difference between go on windows and linux in terms of performance?

JanDeDobbeleer:
> Do you have an idea why there's such a difference between go on windows and linux in terms of performance?

It's not Go that's for sure. Calling external commands is more expensive on Windows due to reasons unknown :-)

JanDeDobbeleer:
I'm going through this tomorrow, need a fresh pair of eyes ;-)

lnu:
> > Do you have an idea why there's such a difference between go on windows and linux in terms of performance?
> 
> It's not Go that's for sure. Calling external commands is more expensive on Windows due to reasons unknown :-)

https://github.com/golang/go/issues/31895
https://github.com/golang/go/issues/25965#issuecomment-482037476

JanDeDobbeleer:
@lnu ah crap. It's really there. Why though, I can't even bypass this without forking go entirely.

JanDeDobbeleer:
@lnu can you squash all commits into 1 already? I'm reviewing this tomorrow morning.

lnu:
Can be tested easily in windows with:`Set-PoshPrompt -Theme jandedobbeleer -ShowDebug $true`
For mac or linux, by updating the shell config file.

JanDeDobbeleer:
> I can't even bypass this without forking go entirely

I've been busy figuring out ways to bypass the wait() and timeout it includes. Almost there. Shaves off some time but some calls are broken (spotify for example on MacOS). Without that one it's blazing fast. I'll properly review this one as I can use it to get a diff view of both versions. Feeling faster can be deceiving, numbers never lie.

engine.go
```diff
@@ -160,3 +165,22 @@ func (e *engine) reset() {
 	e.previousActiveSegment = nil
 	e.activeBlock = nil
 }
+
+func trackTiming(showDebug bool, s *Segment) {
```
JanDeDobbeleer: I would rename this and make it part of `Segment`:

```golang
func (s *Segment) setStringValue(env environmentInfo, cwd string, ignoreFolders []string, debug bool) {
    err := s.mapSegmentWithWriter(env)
    if err != nil || s.hasValue(IgnoreFolders, cwd) {
	return
    }
    if debug {
        start := time.Now()
	defer (func() {
		s.active = true
		elapsed := time.Since(start)
		s.timing = elapsed
	})()
    }
    if s.enabled() {
	s.stringValue = s.string()
    }
}
```

We can then rewrite our for loop above to:

```golang
for _, segment := range segments {
	go func(s *Segment) {
		defer wg.Done()
                s.setStringValue(e.env, cwd, IgnoreFolders, showDebug)
	}(segment)
}
```

main.go
```diff
@@ -14,6 +14,7 @@ type args struct {
 	Config      *string
 	Shell       *string
 	PWD         *string
+	ShowDebug   *bool
```
JanDeDobbeleer: Would rename to `Debug`

packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -36,9 +37,14 @@ function Set-PoshPrompt {
     param(
         [Parameter(Mandatory = $false)]
         [string]
-        $Theme
+        $Theme,
+        [Parameter(Mandatory = $false)]
+        [bool]
+        $ShowDebug = $false
```
JanDeDobbeleer, line 43: Same as above: `Debug`

JanDeDobbeleer (changes requested):

main.go
lnu: The thing is Debug is a reserved keyword in powershell I think, so the parameter in powershell and in oh-my-posh will have a different name.

packages/powershell/oh-my-posh/oh-my-posh.psm1
lnu: same as above too :)

main.go
JanDeDobbeleer: That's fine for me, I don't mind the name offset, it makes more sense to name this Debug and extend more afterwards.

segment.go
```diff
@@ -149,3 +153,26 @@ func (segment *Segment) mapSegmentWithWriter(env environmentInfo) error {
...
+	}
+	// add timing only in debug
+	if debug {
+		start := time.Now()
+		defer (func() {
+			// force segment rendering to display the time it took
+			// to check if the segment is enabled or not
+			// depending on the segement, calling enabled()
+			// can be time consuming
+			segment.active = true
```
JanDeDobbeleer, line 170: Can you elaborate why you'd want to do this? As it won't necessarily run the logic (which is what you want to time).

lnu: because for some segments, calling `enabled()` is time consuming even if it returns false(ie. git segment).
This way we have all configured segments from the json and their timings. Otherwise we can miss some informations. 
Example on a non git folder:
image

JanDeDobbeleer:
If you squash these commits into 1, I'll merge this one! 👍🏻

PR #171, opened 2020-11-19
feat: customizable short path

lnu:
### Prerequisites

### Description

Customizable paths. Used like this:
```
{
      "type": "path",
      "style": "powerline",
      "powerline_symbol": "\uE0B0",
      "foreground": "#ffffff",
      "background": "#ff479c",
      "properties": {
        "prefix": " \uE5FF ",
        "style": "agnoster_full",
        "mappedlocations": [
          ["D:\\dev", "\ue799"],
          ["D:\\tools", "\ue20f"],
          ["D:\\sql", "\ue706"],
          ["C:\\temp", "\uf1f8"]
        ]
     }
},
```

### DIFF
--- a/properties.go
+++ b/properties.go
@@ -85,6 +85,20 @@ func (p *properties) getBool(property Property, defaultValue bool) bool {
 	return boolValue
 }
 
+func (p *properties) getKeyValueMap(property Property, defaultValue map[string]string) map[string]string {
+	if p == nil || p.values == nil {
+		return defaultValue
+	}
+	val, found := p.values[property]
+	if !found {
+		return defaultValue
+	}
+
+	keyValues := parseKeyValueArray(val)
+
+	return keyValues
+}
+
 func parseStringArray(value interface{}) []string {
 	expectedValue, ok := value.([]interface{})
 	if !ok {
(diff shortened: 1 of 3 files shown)


segment_path.go
```diff
@@ -64,11 +64,27 @@ func (pt *path) getShortPath() string {
...
+
 	mappedLocations := map[string]string{
 		"HKCU:":          pt.props.getString(WindowsRegistryIcon, "\uE0B1"),
 		"HKLM:":          pt.props.getString(WindowsRegistryIcon, "\uE0B1"),
 		pt.env.homeDir(): pt.props.getString(HomeIcon, "~"),
 	}
+
+	// merge custom locations with mapped locations
+	// mapped locations can override predefined locations
+	if pt.props.values["mappedlocations"] != nil {
```
JanDeDobbeleer: This needs to be a function in `properties.go` just like [here](https://github.com/JanDeDobbeleer/oh-my-posh3/blob/main/properties.go#L88) for example. Then we can use `append` to merge the maps (I believe).

JanDeDobbeleer (changes requested):

properties.go
```diff
@@ -96,3 +110,20 @@ func parseStringArray(value interface{}) []string {
 	}
 	return list
 }
+
+func parseKeyValueArray(value interface{}) map[string]string {
+	locations, ok := value.([]interface{})
+	if !ok {
+		return map[string]string{}
+	}
+	mappedLocations := make(map[string]string)
```
JanDeDobbeleer: this variable needs a name change as it's not necessary related to mapped locations ;-)
`keyValueArray` might work.

JanDeDobbeleer (changes requested):

properties.go
lnu: my bad

Issue #179, opened 2020-11-21
[Feature request] Set background & foreground inline 


fedeiglesias:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

This is not an issue, its a "nice to have" feature.
As we read in the [color documentation](https://ohmyposh.dev/docs/configure#colors) we can set a foreground color in the segment for any specific text.
It will be nice to be able to set also the background, i think something like this

<#ff0000,#00ff00>Hello budy!</>

Where the first or the second can be the foreground and background.

It will be usefull in escenarios like that, in where we want to put a "label" for an specific path (but not for the rest)

![image](https://user-images.githubusercontent.com/21363793/99865061-8cff3380-2ba7-11eb-84f4-b2356f97b517.png)

Thank you for this  awesome tool !!!

JanDeDobbeleer:
Let's see what can be done. I think it's possible, and the suggestion could even work in terms of interface.

shedric1:
I was looking for a way to do this, and the syntax suggestion inspired me to get it set up. Hope that's all right! I'm no expert with Go, so feel free to suggest changes/totally alter it as needed.

Existing themes with in-line colors shouldn't need to change anything, the suggested syntax only adds new options.

And now I can connect `plain` style segments to `powerline` segments without a blank space between!
![image of inline background color changes](https://i.imgur.com/wLDPV34.png)

JanDeDobbeleer:
Should be live! (or ASAP 😅)

fedeiglesias:
@shedric1 thank you very much mate! I think i need to learn some GoLang 😄
@JanDeDobbeleer  I'm in love with oh-my-posh!!! in my new job we use Win env for work and I won't miss bash at all!
Here is the feature working like a charm 👌

![ohmyposh](https://user-images.githubusercontent.com/21363793/99890762-fee88300-2c62-11eb-92b2-f8125e26d20b.gif)

PR #180, opened 2020-11-21
feat: add syntax for inline background color selection

shedric1:
### Prerequisites

### Description

New feature as discussed in #179 

Adds additional syntax to inline color selection so that both background and foreground colors can be customized in supported strings. Should be a non-breaking change. I have noticed no parsing errors or problems with existing themes. Previous in-line color syntax is only supplemented, not changed. As far as I know, no themes should need to be altered to keep working with the new feature.

Previous syntax for foreground color selection (unchanged):
`"prefix": "normal color string <#CB4B16>changed color string</>"`

New syntax for selecting both foreground and background:
`"prefix": "normal color string <#CB4B16,#000000>changed color string</>"`

Alternative syntax to change only background color, leaving foreground unchanged:
`"prefix": "normal color string <,#000000>changed color string</>"`
 
As shown, just separate the two colors by a comma to specify both. Or omit the first color and leave the comma to set only the background color

### DIFF
--- a/renderer_test.go
+++ b/renderer_test.go
@@ -41,6 +41,27 @@ func TestWriteColorOverride(t *testing.T) {
 	assert.NotContains(t, renderer.string(), "<#ff5733>")
 }
 
+func TestWriteColorOverrideBackground(t *testing.T) {
+	renderer := &Renderer{
+		Buffer: new(bytes.Buffer),
+	}
+	text := "This is white, <,#000000>this is black</>, white again"
+	renderer.init("pwsh")
+	renderer.write("#193549", "#ff5733", text)
+	assert.NotContains(t, renderer.string(), "000000")
+}
+
+func TestWriteColorOverrideBoth(t *testing.T) {
+	renderer := &Renderer{
+		Buffer: new(bytes.Buffer),
+	}
+	text := "This is white, <#000000,#ffffff>this is black</>, white again"
+	renderer.init("pwsh")
+	renderer.write("#193549", "#ff5733", text)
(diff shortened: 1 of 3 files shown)


shedric1:
Reopening pull request after addressing issues with golangci code standards

renderer_test.go
```diff
@@ -41,6 +41,27 @@ func TestWriteColorOverride(t *testing.T) {
...
+	renderer := &Renderer{
+		Buffer: new(bytes.Buffer),
+	}
+	text := "This is white, <,#000000>this is black</>, white again"
+	renderer.init("pwsh")
+	renderer.write("#193549", "#ff5733", text)
+	assert.NotContains(t, renderer.string(), "000000")
+}
+
+func TestWriteColorOverrideBoth(t *testing.T) {
```
JanDeDobbeleer, line 66: can you add a test with 16 colors eg: `"This is white, <black,white>this is black</>, white again"`

JanDeDobbeleer (changes requested):

renderer_test.go
shedric1: I added some additional unit tests to ensure the 16 color format is also properly handled, let me know if this is what you're looking for!

JanDeDobbeleer: I'm going to do a quick checkout of your branch as I feel the regex could be updated. But I'm not sure so I'll do a quick debug ;-)

JanDeDobbeleer: I'll refactor the regex later. It's out of scope for this one. Well done!

PR #200, opened 2020-11-29
feat: support Powershell 5

PIYUSH194:
### Prerequisites
- [ ] Fix iswindows  variable found in windows powershell

### Description

If iswindows variable not available in windows powershell it will through chmod command not found.
not applicable on powershell 6 and greater


JanDeDobbeleer:
Wouldn't it semantically make more sense to check for PSVersion > 5?

JanDeDobbeleer:
Given that support for 6 isn't available, I'd opt for lowering the version number to 5 if this is all that's needed to support 5. I suppose you're using this code in PS5?

PIYUSH194:
It works fine in PS5.. why no support for PS5 ?

JanDeDobbeleer:
That was my question indeed. Allow me to extend this properly tomorrow.

packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -20,7 +20,7 @@ function Get-PoshCommand {
 }
 
 # Set the right binary to executable before doing anything else
-if (!$IsWindows) {
+if (!$IsWindows  -and (Get-Item "Variable:*IsWindows*")) {
```
JanDeDobbeleer: Can we change this to: `if ($PSVersionTable.PSEdition -eq "Core" -and !$IsWindows)`
Which means we can change `oh-my-posh.psm1` to include 5 as well within the scope of this PR and use `feat: support Powershell 5` as a commit message.

JanDeDobbeleer (changes requested):

packages/powershell/oh-my-posh/oh-my-posh.psm1
PIYUSH194: How can i add only last commit to this pull request? Or you can close this request and add it manually

JanDeDobbeleer: If you need to know how, learning to rebase is key here -> https://www.algolia.com/blog/master-git-rebase/
I can always pick this change, you stay the author in that case, let me know what would work for you.

PIYUSH194: Just pick the change that will be good.

Issue #201, opened 2020-11-29
Powerlevel10k Classic renders gaps between blocks in PowerShell


SirJson:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

This behavior does not happen with every font, only on startup and only with the Powerlevel10k classic theme. I've made a screenshot of this issue. The problematic font is Hasklug NF.

![grafik](https://user-images.githubusercontent.com/1042414/100552287-7f703a80-3286-11eb-917b-aa80e09665db.png)

The weird part is, it seems to be inconsistently broken, just two hours before I've seen gaps in basically every block. Pressing enter resolves the issue or executing the `clear` command.

I've had a look at the theme file and the code but couldn't see anything obvious. The source seems to be the `postfix` space char used in the theme. Removing that property or replacing it with for example a 'x' fills the gab.  

After some debugging I've found that when the `postfix` spaces are replaced with \u2588 (the Unicode char for "Full Block") the issue is fixed. I tried to use the Unicode char for space to see if that solves the issue but it didn't make any difference. I will submit a PR with the fixed theme, as far as I can tell it doesn't break anything but I could be wrong.

### Environment

- Oh my Posh version: 3.46.1
- Theme: powerlevel10k_classic
- Operating System: Windows 10 Version 20H2
- Shell: PowerShell 7.1.0
- Terminal: Windows Terminal 1.4.3243.0 Preview

### Steps to Reproduce

1. Add `Set-PoshPrompt -Theme powerlevel10k_classic` in your `Microsoft.PowerShell_profile.ps1` file.
2. Open the Windows Terminal settings and set  `Hasklug NF` as default Font (I've used size 12 but other size values did the same)
3. Modify the PowerShell command line to not print the logo `"commandline": "pwsh.exe -nol"`

**Expected behavior:** That the first line of my shell prompt is rendered correctly without gaps between segments

**Actual behavior:** The first line will get rendered with gaps in between segments, sometimes everywhere, sometimes just on the right side.

SirJson:
It looks like the bug is still around in the latest version:

![grafik](https://user-images.githubusercontent.com/1042414/101982865-28e80080-3c77-11eb-9b90-c613db6c534b.png)

[...]

PR #202, opened 2020-11-29
WIP: fix(themes): powerlevel10k_classic renders without gaps on startup

SirJson:
fix: powerlevel10k_classic renders without gaps on startup 

Related to #201 

### Prerequisites

### Description

My solution is pretty much replacing `postfix` spaces with \u2588 and the correct color.
The session segment could not be fixed that way here I had to remove the `postfix` color.
Reviewing my changes made me realize it could be related to inline coloring code for the prompt since the session segment added a gap every time I restored the original color on this segment. 

_Off-Topic_: I see you require a test for the bug fix and documentation to be updated.

I would need guidance here though because I am not a Go programmer and I've never written an integration test in any language where PowerShell was a component. The documentation part is a bit unclear because it could mean that you require an entry in a CHANGELOG file which I understand or in the actual documentation which I wouldn't understand since the PR is fixing unexpected behavior.

### DIFF
--- a/themes/powerlevel10k_classic.omp.json
+++ b/themes/powerlevel10k_classic.omp.json
@@ -26,7 +26,7 @@
           "background": "#546E7A",
           "properties": {
             "style": "short",
-            "postfix": " "
+            "postfix": "<#546E7A>\u2588</>"
           }
         },
         {
@@ -62,7 +62,7 @@
           "background": "#546E7A",
           "leading_diamond": "\uE0B2",
           "properties": {
-            "postfix": " <#26C6DA>\uE0B3</> "
+            "postfix": " \uE0B3"
           }
         },
         {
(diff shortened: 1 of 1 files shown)


themes/powerlevel10k_classic.omp.json
```diff
@@ -26,7 +26,7 @@
           "background": "#546E7A",
           "properties": {
             "style": "short",
-            "postfix": " "
+            "postfix": "<#546E7A>\u2588</>"
```
JanDeDobbeleer, line 29: What if we swapped the default value of `prefix` to this character? This isn't the only theme having this issue on startup on Powershell, and I like this crazy out-of-the-box idea.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@SirJson I've made a new PR #209 which tackles this all across the board. Thank you so much for this idea!

PR #225, opened 2020-12-06
feat: execution time segment

tradiff:
### Prerequisites

### Description

Add a segment to display the execution time of the most recently run command.  The user can specify a minimum threshold required in order to enable the segment.

image

### Note:
Because Powershell stores this information in memory, and only for the current session, the execution time could not be determined by the Go code. For this reason, I had to modify the Powershell code which bootstraps oh-my-posh3, and pass this data in as a CLI parameter.

This segment currently only works in Powershell.  If this is an issue, please let me know I can try to learn about the other shells, and how to implement this from those shells.  If this limitation is acceptable, the documentation should probably mention that this segment is Powershell only. I could not find any prior precedent for that (but I did find segments that only work on Windows). Please let me know how you'd like me to proceed here.

### Roadmap:
The next step for this segment is probably to allow custom formatting of the time string. I decided to wait and do this in a separate PR, because Go doesn't have a built in way to format Duration with a template string, the way it does for Time. And the default string representation for Duration is good enough. It is not in my preferred format though, so that PR is definitely coming 😉

### DIFF
--- a/docs/docs/contributing-segment.md
+++ b/docs/docs/contributing-segment.md
@@ -76,7 +76,7 @@ go build -o $GOPATH/bin/oh-my-posh
 
 ## Add the documentation
 
-Create a new `markdown` file underneath the [`docs/docs`][docs] folder called `new-segment.md`.
+Create a new `markdown` file underneath the [`docs/docs`][docs] folder called `segment-new.md`.
 Use the following template as a guide.
 
 ````markdown
@@ -113,6 +113,43 @@ Display something new.
 ## Map the new documentation in the sidebar
 
 Open [`sidebars.js`][sidebars] and add your document id (`new`) to the items of the Segments category.
+
+## Add the JSON schema
+
+Edit the `themes/schema.json` file to add your segment.
+
+At `$.definitions.segment.properties.type.enum`, add your `SegmentType` to the array:
+
+```json
+new,
+```
+
+At `$.definitions.segment.allOf`, add your segment details:
+```json
+{
+  "if": {
+    "properties": {
+      "type": { "const": "new" }
+    }
+  },
+  "then": {
+    "title": "Display something new",
+    "description": "https://ohmyposh.dev/docs/new",
+    "properties": {
+      "properties": {
+        "properties": {
+          "nwprop": {
+            "type": "string",
+            "title": "New Prop",
+            "description": "the new text to show",
+            "default": "\uEFF1"
+          }
+        }
+      }
+    }
+  }
+}
+```
 
 ## Create a pull request
 
--- a/docs/docs/segment-executiontime.md
+++ b/docs/docs/segment-executiontime.md
@@ -1,0 +1,33 @@
+---
+id: executiontime
+title: Execution Time
+sidebar_label: Execution Time
+---
+
+## What
+
+Displays the execution time of the previously executed command.
+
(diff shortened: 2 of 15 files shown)


docs/docs/segment-executiontime.md
```diff
@@ -0,0 +1,29 @@
+---
+id: executiontime
+title: Execution Time
+sidebar_label: Execution Time
+---
+
+## What
+
+Displays the execution time of the previously executed command.
```
JanDeDobbeleer, line 9: Interesting to add here is how to use it. Given you either need to use the PowerShell module (or set it up manually by using the snippet), it might be handy to know this is needed as a prerequisite. There's information on how to do this in [zsh](https://gist.github.com/knadh/123bca5cfdae8645db750bfb49cb44b0) and bash (see the accepted answer on [this post](https://stackoverflow.com/questions/2704635/is-there-a-way-to-find-the-running-time-of-the-last-executed-command-in-the-shel)) as well.

JanDeDobbeleer (changes requested):
A bit that's missing, but that's on me as I still have to ad this to the docs, is the extension of `schema.json` in `/themes`.

docs/docs/segment-executiontime.md
tradiff: * I have updated schema.json
* I have added schema.json information to the contribution docs
* I have updated the zsh instructions
  * I took the liberty to rename your functions and namespace everything with "omp". Let me know if you dislike that change.

Later today, I'll look into bash and also make the segment-execution.md changes. I also want to change this documentation to use a different icon, since the icon I selected is one that is bugged on powerline segments.

docs/docs/contributing-segment.md
```diff
@@ -114,6 +114,43 @@ Display something new.
 
 Open [`sidebars.js`][sidebars] and add your document id (`new`) to the items of the Segments category.
 
+## Add the JSON schema
+
+Edit the `themes/schema.json` file to add your segment.
+
+At `$.definitions.segment.properties.type.enum`, add your `SegmentType` to the array:
+
+```javascript
```
JanDeDobbeleer: I think it's possible to use `JSON`

docs/docs/segment-executiontime.md
JanDeDobbeleer: Looks good to me, love the addition of the `omp` renames, makes sense (I initially copied those from elsewhere).

JanDeDobbeleer (changes requested):

docs/docs/contributing-segment.md
tradiff: done

docs/docs/segment-executiontime.md
tradiff: @JanDeDobbeleer I had trouble getting the execution time with bash. I did find some scripts that give you the zshell-style hooks, but they are quite long, and would likely complicate the bash installation quite a bit. [Here is an example](https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh).

I'm thinking we include a general description stating this segment works with the Powershell module, or with an execution-time parameter passed in. And provide the installation instructions for Powershell and Zshell. I have updated the docs in this manner. Please have a look and see what you think.

JanDeDobbeleer (approved):

PR #230, opened 2020-12-10
feat: execution time formats

tradiff:
### Prerequisites

### Description

Allow the user to select a display format for the execution time.

### DIFF
--- a/segment_executiontime.go
+++ b/segment_executiontime.go
@@ -1,7 +1,11 @@
 package main
 
 import (
-	"time"
+	"fmt"
+	"strconv"
+
+	lang "golang.org/x/text/language"
+	"golang.org/x/text/message"
 )
 
 type executiontime struct {
@@ -10,9 +14,26 @@ type executiontime struct {
 	output string
 }
 
+type DurationStyle string
+
 const (
 	// ThresholdProperty represents minimum duration (milliseconds) required to enable this segment
 	ThresholdProperty Property = "threshold"
+
+	Austin    DurationStyle = "austin"
+	Roundrock DurationStyle = "roundrock"
+	Dallas    DurationStyle = "dallas"
+	Galveston DurationStyle = "galveston"
+	Houston   DurationStyle = "houston"
+	Amarillo  DurationStyle = "amarillo"
+
+	second           = 1000
+	minute           = 60000
+	hour             = 3600000
+	day              = 86400000
+	secondsPerMinute = 60
+	minutesPerHour   = 60
+	hoursPerDay      = 24
 )
 
 func (t *executiontime) enabled() bool {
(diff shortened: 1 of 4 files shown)


segment_executiontime.go
```diff
@@ -35,3 +41,115 @@ func (t *executiontime) init(props *properties, env environmentInfo) {
 	t.props = props
 	t.env = env
 }
+
+func (t *executiontime) formatDuration(ms int64, format string) string {
+	switch format {
+	case "austin":
```
JanDeDobbeleer: In other segments (path for example), these are created as a separate type (maybe `DurationFormat`).

JanDeDobbeleer (changes requested):

segment_executiontime.go
tradiff: Just making sure I'm on the same page. I'm looking at `segment_path.go`. You're asking for constants, right?
Also, unrelated, I see `path` calls a similar attribute `style` instead of `format`. I'm thinking `executiontime` should also go with the name `style`, for consistency.

JanDeDobbeleer: About `style`, YES! That's a very good reference to use. When you have properties which can be used cross segment, add them to `properties.go` just like `DisplayVersion`.

tradiff: @JanDeDobbeleer  all done!

segment_executiontime.go
```diff
@@ -35,3 +48,115 @@ func (t *executiontime) init(props *properties, env environmentInfo) {
 	t.props = props
 	t.env = env
 }
+
+func (t *executiontime) formatDuration(ms int64, style DurationStyle) string {
+	switch style {
+	case AustinStyle:
```
JanDeDobbeleer: I would remove the `Style` postfix, the type implies it's a style.

segment_executiontime.go
```diff
@@ -35,3 +48,115 @@ func (t *executiontime) init(props *properties, env environmentInfo) {
...
+
+func (t *executiontime) formatDurationAustin(ms int64) string {
+	if ms < 1000 {
+		return fmt.Sprintf("%dms", ms%1000)
+	}
+
+	seconds := float64(ms%60000) / 1000
+	result := strconv.FormatFloat(seconds, 'f', -1, 64) + "s"
+
+	if ms >= 60000 {
```
JanDeDobbeleer: Looking at this, the `6000`, `3600000` etc numbers are recurring and can be considered as magic variables (you need to do the math). Can we replace these with a meaningful constant?

```go
const (
    second = 1000
    minute = 60000
    hour = 3600000
    ....
)
```

JanDeDobbeleer (changes requested):
Last bit, looks great!

segment_executiontime.go
tradiff: done

segment_executiontime.go
tradiff: done

JanDeDobbeleer (approved):

PR #242, opened 2020-12-15
feat: powershell rprompt support

tradiff:
### Prerequisites

### Description

Process `rprompt` blocks for Powershell.  For Powershell, align `rprompt` blocks to the right on the same line as the cursor.

** Note:
There is one side effect which some people may not like. If the user's input goes over into the region of the right aligned prompt text, it just clobbers the right aligned prompt text.  I'm not aware of any way around this, and I'm personally fine with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,6 @@ dist
 .history
 
 # End of https://www.toptal.com/developers/gitignore/api/node,go,visualstudiocode
+
+# linux binary
+/oh-my-posh3
--- a/ansi_renderer.go
+++ b/ansi_renderer.go
@@ -30,12 +30,14 @@ func lenWithoutANSI(text, shell string) int {
 }
 
 type formats struct {
-	linechange string
-	left       string
-	right      string
-	title      string
-	creset     string
-	clearOEL   string
+	linechange            string
+	left                  string
+	right                 string
+	title                 string
+	creset                string
+	clearOEL              string
+	saveCursorPosition    string
+	restoreCursorPosition string
 }
 
 // AnsiRenderer exposes functionality using ANSI
@@ -46,8 +48,10 @@ type AnsiRenderer struct {
 }
 
 const (
-	zsh  = "zsh"
-	bash = "bash"
+	zsh         = "zsh"
+	bash        = "bash"
+	pwsh        = "pwsh"
+	powershell5 = "powershell"
 )
 
 func (r *AnsiRenderer) init(shell string) {
@@ -61,6 +65,8 @@ func (r *AnsiRenderer) init(shell string) {
 		r.formats.title = "%%{\033]0;%s\007%%}"
 		r.formats.creset = "%{\x1b[0m%}"
 		r.formats.clearOEL = "%{\x1b[K%}"
+		r.formats.saveCursorPosition = "%{\x1b7%}"
+		r.formats.restoreCursorPosition = "%{\x1b8%}"
 	case bash:
 		r.formats.linechange = "\\[\x1b[%d%s\\]"
 		r.formats.left = "\\[\x1b[%dC\\]"
@@ -68,6 +74,8 @@ func (r *AnsiRenderer) init(shell string) {
 		r.formats.title = "\\[\033]0;%s\007\\]"
 		r.formats.creset = "\\[\x1b[0m\\]"
 		r.formats.clearOEL = "\\[\x1b[K\\]"
+		r.formats.saveCursorPosition = "\\[\x1b7\\]"
+		r.formats.restoreCursorPosition = "\\[\x1b8\\]"
 	default:
 		r.formats.linechange = "\x1b[%d%s"
 		r.formats.left = "\x1b[%dC"
@@ -75,6 +83,8 @@ func (r *AnsiRenderer) init(shell string) {
 		r.formats.title = "\033]0;%s\007"
 		r.formats.creset = "\x1b[0m"
 		r.formats.clearOEL = "\x1b[K"
+		r.formats.saveCursorPosition = "\x1b7"
+		r.formats.restoreCursorPosition = "\x1b8"
 	}
 }
 
@@ -115,4 +125,12 @@ func (r *AnsiRenderer) clearEOL() {
 
 func (r *AnsiRenderer) string() string {
 	return r.buffer.String()
+}
+
+func (r *AnsiRenderer) saveCursorPosition() {
+	r.buffer.WriteString(r.formats.saveCursorPosition)
+}
+
+func (r *AnsiRenderer) restoreCursorPosition() {
+	r.buffer.WriteString(r.formats.restoreCursorPosition)
 }
--- a/engine.go
+++ b/engine.go
@@ -168,10 +168,18 @@ func (e *engine) render() {
 func (e *engine) write() {
 	if *e.env.getArgs().Eval {
 		fmt.Printf("PS1=\"%s\"", e.renderer.string())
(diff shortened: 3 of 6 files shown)


renderer.go
```diff
@@ -235,3 +235,13 @@ func (r *Renderer) print(text string) {
 func (r *Renderer) clearEOL() {
 	fmt.Print(r.formats.clearOEL)
 }
+
+func (r *Renderer) saveCursorPosition() {
+	// todo: move to r.formats and handle zsh, bash, pwsh
+	fmt.Printf("\x1b7")
```
JanDeDobbeleer: I tried this before and somehow this didn't work in some shells. Would be cool to see this actually working as initially this is what I intended to create using the right alignment.

JanDeDobbeleer:
> I should move forward with it?

Yes, very much yes :-)

renderer.go
tradiff: Yeah I see what you mean. Things are acting pretty strange for me in bash and zsh. :-/

JanDeDobbeleer: Yup, that's what I had as well.

JanDeDobbeleer: @TravisTX you do need to escape them properly. It might as well have been because of that (I only learned the escaping after trying this).

tradiff: I'm pretty sure I'm escaping correctly. I followed your pattern. Just to be sure on that, I've pushed my changes adjusted for bash and zsh. I have not had much luck getting those shells to cooperate.

Bash: I get some strange behavior when the user's input attempts to word-wrap. Here's a gif of what I'm seeing in bash: https://i.imgur.com/3WmaFxq.gif

Zsh: I also get the strange word-wrap behavior, but additionally the right-aligned segment doesn't display for me.

Have you seen that word-wrap behavior before? Any ideas on that?

JanDeDobbeleer: I tried this as well and I can't get it to work either on zsh/bash. Just like I experienced before. There's however something I'm going to try for zsh specifically.

JanDeDobbeleer: @TravisTX I made #247 which works on ZSH. Maybe we can do something similar for Powershell specifically. As in, the normal behaviour is having blocks which can be moved around. But you have the ability to have 1 block with will be displayed right (as the last one) for certain shells where this actually works.

tradiff: I see that's been merged. I'll change this PR to support `rprompt` in powershell. Once rprompt is handled, do we still need the alignment property? We may end up with some weird conflicts if the app supports both.

JanDeDobbeleer: I've been wondering the same. I thinking they can coexist as the purpose of rprompt is to be rendered AFTER the cursor whereas the other one isn't. And some shells do not support this properly (or we need to figure out how).

engine.go
```diff
@@ -168,11 +171,19 @@ func (e *engine) render() {
 func (e *engine) write() {
 	if *e.env.getArgs().Eval {
 		fmt.Printf("PS1=\"%s\"", e.renderer.string())
-		if e.rprompt != "" && e.env.getShellName() == zsh {
+		if e.env.getShellName() == zsh {
```
tradiff, line 171: I fixed a bug here with zsh. If the rprompt block had output once, and then later didn't have output, the zsh RPOMPT would never get cleared. An easy way to repro this is to create a block with a single text element, get a prompt, and then remove this block from your template and save it. Another way would be using the executiontime segment, which can sometimes emit output and sometimes not, depending on the threshold.

renderer.go
tradiff: Ready for review 😊

ansi_renderer.go
```diff
@@ -48,6 +50,7 @@ type AnsiRenderer struct {
 const (
 	zsh  = "zsh"
 	bash = "bash"
+	pwsh = "pwsh"
```
JanDeDobbeleer: Do you have an idea what this value is in PS5?

engine.go
```diff
@@ -127,14 +127,14 @@ func (e *engine) setStringValues(segments []*Segment) {
 
 func (e *engine) render() {
 	for _, block := range e.settings.Blocks {
+		if block.VerticalOffset != 0 {
```
JanDeDobbeleer: Any participants reason this was moved?

engine.go
```diff
@@ -147,6 +147,9 @@ func (e *engine) render() {
 		case RPrompt:
 			e.rprompt = e.renderBlockSegments(block)
 		}
+		if block.VerticalOffset != 0 {
```
JanDeDobbeleer: Isn't this one doing it twice?

engine.go
JanDeDobbeleer: Got it, didn't think of that!

ansi_renderer.go
tradiff: It looks like it has a shell value of `powershell`. I didn't realize PS5 was one of the supported shells. I'll test things out there and see if I can get this PR to support PS5.

engine.go
tradiff: I'll revert my `vertical_offset` changes in this PR.

But to answer your question, this was actually undoing the vertical_offset, notice the `-` in the `e.renderer.changeLine(-block.VerticalOffset)`. I noticed `vertical_offset` had some buggy feeling behavior depending on how and where you use it, and this was an attempt to make it feel a little more robust. I think that probably deserves it's own PR, as it's a pretty nuanced issue, and not really related to rprompt.  Also I don't think my changes fully fixed it.

engine.go
tradiff: this has been reverted

engine.go
tradiff: this has been reverted

ansi_renderer.go
tradiff: I have added Powershell 5 support. The logic worked fine 🎉

.gitignore
```diff
@@ -145,3 +145,6 @@ dist
 .history
 
 # End of https://www.toptal.com/developers/gitignore/api/node,go,visualstudiocode
+
+# linux binary
+/oh-my-posh3
```
JanDeDobbeleer, line 150: maybe add the exe as well, that way it's ignore everywhere

engine.go
```diff
@@ -168,11 +168,19 @@ func (e *engine) render() {
...
 	if *e.env.getArgs().Eval {
 		fmt.Printf("PS1=\"%s\"", e.renderer.string())
-		if e.rprompt != "" && e.env.getShellName() == zsh {
+		if e.env.getShellName() == zsh {
 			fmt.Printf("\nRPROMPT=\"%s\"", e.rprompt)
 		}
 		return
 	}
+
+	if e.rprompt != "" && (e.env.getShellName() == pwsh || e.env.getShellName() == powershell5) {
```
JanDeDobbeleer, line 177: It's perfectly fine like this for now, I'm going to refactor this function a bit afterwards.

JanDeDobbeleer:
@TravisTX did you validate on Windows/MacOS?

.gitignore
tradiff: `*.exe` already gets ignored right?  Line 7

tradiff:
> @TravisTX did you validate on Windows/MacOS?

Windows only. Unfortunately, I do not have a MacOS computer.

JanDeDobbeleer:
I'll validate on MacOS then 👍🏻

JanDeDobbeleer:
image

.gitignore
JanDeDobbeleer: Yup, you're right!

PR #246, opened 2020-12-18
[WIP] feat: Added display_mode Property to Language Segment

moritz-meier:
### Prerequisites

### Description
It would be nice if language segments (for example python) could always be displayed
This PR adds an 'display_mode' property to the language segment which determines when the segment is displayed.

Following values are allowed for the `display_mode` property:
* `always`: The segment is always displayed
* `when_in_context`: The segment is only displayed if the current folder contains files with certain extensions
* `never`: The segment is never displayed

Default value is: `when_in_context`

Example:
```json
{
    "type": "python",
    "style": "plain",
    "foreground": "magenta",
    "properties": {
        "display_mode": "when_in_context",
	"display_version": true
    }
},
```

### DIFF
--- a/segment_language.go
+++ b/segment_language.go
@@ -10,6 +10,17 @@ type language struct {
 	version      string
 }
 
+const (
+	// DisplayModeProperty sets the display mode (always, when_in_context, never)
+	DisplayModeProperty Property = "display_mode"
+	// DisplayModeAlways displays the segement always
+	DisplayModeAlways string = "always"
+	// DisplayModeContext displays the segment when the current folder contains certain extensions
+	DisplayModeContext string = "context"
+	// DisplayModeNever hides the segment
+	DisplayModeNever string = "never"
+)
+
 func (l *language) string() string {
 	if l.props.getBool(DisplayVersion, true) {
 		return l.version
@@ -18,6 +29,23 @@ func (l *language) string() string {
 }
 
 func (l *language) enabled() bool {
+	displayMode := l.props.getString(DisplayModeProperty, DisplayModeContext)
+	displayVersion := l.props.getBool(DisplayVersion, true)
+	hasVersion := l.getVersion()
+
+	switch displayMode {
+	case DisplayModeAlways:
+		return (hasVersion || !displayVersion)
+	case DisplayModeNever:
+		return false
+	case DisplayModeContext:
+		fallthrough
+	default:
+		return l.isInContext() && (hasVersion || !displayVersion)
+	}
+}
+
+func (l *language) isInContext() bool {
 	for i, extension := range l.extensions {
 		if l.env.hasFiles(extension) {
 			break
@@ -26,6 +54,11 @@ func (l *language) enabled() bool {
 			return false
 		}
 	}
+
+	return true
+}
+
+func (l *language) getVersion() bool {
 	var executable string
 	for i, command := range l.commands {
 		if l.env.hasCommand(command) {
@@ -39,5 +72,6 @@ func (l *language) enabled() bool {
 	versionInfo, _ := l.env.runCommand(executable, l.versionParam)
 	values := findNamedRegexMatch(l.versionRegex, versionInfo)
 	l.version = values["version"]
+
 	return true
 }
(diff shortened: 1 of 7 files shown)


segment_language.go
```diff
@@ -18,6 +23,21 @@ func (l *language) string() string {
 }
 
 func (l *language) enabled() bool {
+
+	displayMode := l.props.getString(DisplayMode, "when_in_context")
+	displayVersion := l.props.getBool(DisplayVersion, true)
+	hasVersion := l.getVersion()
+
+	if displayMode == "always" {
```
JanDeDobbeleer: These would need to be a specific type too: `type DisplayMode string`. I would also opt to change `when_in_context` to `context`.

segment_language.go
```diff
@@ -10,9 +10,21 @@ type language struct {
...
 
+// DisplayMode represents a value for the property display_mode (always, context or never)
+type DisplayMode string
+
 const (
-	// DisplayMode sets the display mode (always, when_in_context, never)
-	DisplayMode Property = "display_mode"
+	// DisplayModeProperty sets the display mode (always, when_in_context, never)
+	DisplayModeProperty Property = "display_mode"
+
```
JanDeDobbeleer: Please omit the spaces between the constants 🙏🏻

segment_language.go
```diff
@@ -23,14 +35,13 @@ func (l *language) string() string {
...
 
 func (l *language) enabled() bool {
-
-	displayMode := l.props.getString(DisplayMode, "when_in_context")
+	displayMode := l.props.getString(DisplayModeProperty, string(DisplayModeContext))
 	displayVersion := l.props.getBool(DisplayVersion, true)
 	hasVersion := l.getVersion()
 
-	if displayMode == "always" {
+	if displayMode == string(DisplayModeAlways) {
```
JanDeDobbeleer: you shouldn't need to unbox it, same as above.

segment_language.go
```diff
@@ -10,6 +10,20 @@ type language struct {
...
 }
 
+// DisplayMode represents a value for the property display_mode (always, context or never)
+type DisplayMode string
+
+const (
+	// DisplayModeProperty sets the display mode (always, when_in_context, never)
+	DisplayModeProperty Property = "display_mode"
+	// DisplayModeAlways displays the segement always
+	DisplayModeAlways DisplayMode = "always"
```
JanDeDobbeleer: I was incorrect in my previous example. Just like in `segment_path.go` it's better to use strings:

```golang
// DisplayModeAlways displays the segement always
DisplayModeAlways string = "always"
// DisplayModeContext displays the segment when the current folder contains certain extensions
DisplayModeContext string = "context"
// DisplayModeNever hides the segment
DisplayModeNever string = "never"
```

segment_language.go
```diff
@@ -18,6 +32,20 @@ func (l *language) string() string {
 }
 
 func (l *language) enabled() bool {
+	displayMode := DisplayMode(l.props.getString(DisplayModeProperty, string(DisplayModeContext)))
+	displayVersion := l.props.getBool(DisplayVersion, true)
+	hasVersion := l.getVersion()
+
+	if displayMode == DisplayModeAlways {
```
JanDeDobbeleer: I'm a strong proponent of [object calisthenics](https://williamdurand.fr/2013/06/03/object-calisthenics/), and try to avoid `else` most I can. It's better to use a `switch` statement in this case. Taking my previous comment in consideration:

```golang
func (l *language) enabled() bool {
	displayMode := l.props.getString(DisplayModeProperty, DisplayModeContext)
	displayVersion := l.props.getBool(DisplayVersion, true)
	hasVersion := l.getVersion()
	switch displayMode {
	case DisplayModeAlways:
		return (hasVersion || !displayVersion)
	case DisplayModeNever:
		return false
	case DisplayModeContext:
		fallthrough
	default:
		return l.isInContext() && (hasVersion || !displayVersion)
	}
}
```

JanDeDobbeleer (changes requested):
Can you address the latest comments and rebase (rather than merge master in your branch) so we can do a fast forward only merge? Also keep the conventional commits in mind ;-)

JanDeDobbeleer:
@moritz-meier perfect, can you rebase so there's only 1 commit? An interactive rebase with fixups should do the job. There's no need for a historical context to contain PR comment rework 🙏🏻

JanDeDobbeleer:
Added this to #255 to avoid having to ask to once again rebase, thanks for all the work @moritz-meier 🙏🏻

PR #247, opened 2020-12-18
feat: zsh rprompt compatibility

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]


tradiff:
Interesting idea using RPROMPT on supported shells. What changes need to happen in .zshrc in order to test this PR out in zshell?

JanDeDobbeleer:
> What changes need to happen in .zshrc

@TravisTX  I adjusted the docs here as well, it's in the installation guide. I would see if we can reuse `rprompt` for PowerShell as well but then by storing the pointer, writing the block on the far right and restoring the pointer (like what you did). We can add this in the `write()` function in `engine.go`, but without `eval` as that won't work in PowerShell (and isn't needed). What do you think?

Issue #236, opened 2020-12-14
Falls back to standard prompt after failed command


firlus:
### Prerequisites

- [X] I have read and understand the `CONTRIBUTING` guide
- [X] I looked for duplicate issues before submitting this one

### Description

When I type in a command that e.g. does not exists, the prompt shows only `PS>` after that until I open a new console window.

### Environment

- Oh my Posh version: 3
- Theme: Agnoster
- Operating System: Windows 10
- Shell: Powershell
- Terminal: Windows Terminal

### Steps to Reproduce

This is my $PROFILE:
```
[ScriptBlock]$Prompt = {
    $lastCommandSuccess = $?
    $realLASTEXITCODE = $global:LASTEXITCODE
    $errorCode = 0
    if ($lastCommandSuccess -eq $false) {
        #native app exit code
        if ($realLASTEXITCODE -is [int] -and $realLASTEXITCODE -gt 0) {
            $errorCode = $realLASTEXITCODE
        }
        else {
            $errorCode = 1
        }
    }

    $executionTime = -1
    $history = Get-History -ErrorAction Ignore -Count 1
    if ($null -ne $history) {
        $executionTime = $history.Duration.TotalMilliseconds
    }

    $startInfo = New-Object System.Diagnostics.ProcessStartInfo
    $startInfo.FileName = "C:\tools\oh-my-posh.exe"
    $cleanPWD = $PWD.ProviderPath.TrimEnd("\")
    $startInfo.Arguments = "-config=""$env:USERPROFILE\.poshthemes\agnoster.omp.json"" -error=$errorCode -pwd=""$cleanPWD"" -execution-time=$executionTime"
    $startInfo.Environment["TERM"] = "xterm-256color"
    $startInfo.CreateNoWindow = $true
    $startInfo.StandardOutputEncoding = [System.Text.Encoding]::UTF8
    $startInfo.RedirectStandardOutput = $true
    $startInfo.UseShellExecute = $false
    if ($PWD.Provider.Name -eq 'FileSystem') {
      $startInfo.WorkingDirectory = $PWD.ProviderPath
    }
    $process = New-Object System.Diagnostics.Process
    $process.StartInfo = $startInfo
    $process.Start() | Out-Null
    $standardOut = $process.StandardOutput.ReadToEnd()
    $process.WaitForExit()
    $standardOut
    $global:LASTEXITCODE = $realLASTEXITCODE
    #remove temp variables
[...]

PR #248, opened 2020-12-19
feat: execution time for powershell 5

tradiff:
### Prerequisites

### Description

Powershell 5's history object does not contain the `Duration` property, but it does have `StartExecutionTime` and `EndExecutionTime` which can be used to calculate the duration.

The calculation didn't seem to impact performance, so I changed the code to only do the calculation (even for powershell 7). That decision was made in order to keep the powershell code as short as possible, since this is code we're asking people to paste into their profile. If you don't agree with this decision, let me know and we can check to see if `Duration` exists, and if not, fall back to doing the calculation.

BTW:  I believe this is what was happening in #236 


JanDeDobbeleer (approved):
Just a quick rebase and we can merge this one!

PR #251, opened 2020-12-21
feat: allow comments in json

tradiff:
### Prerequisites

### Description

As a user who is playing with my theme files quite a bit, I would like the ability to use comments in the json configuration. This would be useful for explanations of configuration as well as for temporarily commenting out configuration.  

Comments aren't permitted in JSON, but other products have adopted JSONC, a superset of JSON which does allow comments. Products like VS Code and Windows Terminal for example.  JSONC is a superset of JSON, so existing JSON configurations are still perfectly valid.

This PR uses [muhammadmuzzammil1998/jsonc](https://github.com/muhammadmuzzammil1998/jsonc) to convert the JSONC to JSON.  Go's standard `encoding/json` is used after that, just like before.

The `--print-config` option still returns regular JSON.


JanDeDobbeleer:
Awesome!

PR #257, opened 2020-12-23
chore: add execution condition for powershell

Vixb1122:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -223,7 +223,9 @@ oh-my-posh --print-shell
 Edit `$PROFILE` in your preferred PowerShell version and add the following line.
 
 ```powershell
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@Vixb1122 can you elaborate why this change is necessary?

Vixb1122:
You don't have to delete the contents of $profile after uninstalling oh-my-posh. @JanDeDobbeleer

JanDeDobbeleer:
I gave this some thought over the night and decided not to accept this. The reasoning is that when you install a module for example, and add that to you `$PROFILE`, you're also not going to check if that module is still on your machine before running `Import-Module whatever`. If you removed oh-my-posh (deliberately or not), one would expect the user to also remove this statement, or to see it fail so he/she is reminded this still needs to go way.

PR #263, opened 2020-12-24
feat: use mapped_locations in folder style

### LINKED ISSUE #261: When I am in the home directory, I want my prompt to show "~"
When I am in the home directory, I want my prompt to show "~"

But when I am not, I don't want it to show the full path. Just the current directory name.

/cc @JanDeDobbeleer

_Originally posted by @itsme-alan in https://github.com/JanDeDobbeleer/oh-my-posh3/discussions/258_

tradiff:
### Prerequisites

### Description

Allow the `mapped_locations` property to work when using the `folder` style.

closes #261 

### DIFF
--- a/docs/docs/segment-path.md
+++ b/docs/docs/segment-path.md
@@ -34,7 +34,7 @@ Display the current path.
 - windows_registry_icon: `string` - the icon to display when in the Windows registry - defaults to `\uE0B1`
 - style: `enum` - how to display the current path
 - mapped_locations: `map[string]string` - custom glyph/text for specific paths(only when `style` is set to `agnoster`,
-`agnoster_full` or `short`)
+`agnoster_full`, `agnoster_short`, `short`, or `folder`)
 
 ## Style
 
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/distatus/battery v0.10.1-0.20200722221337-7e1bf2bbb15c
 	github.com/go-ole/go-ole v1.2.4 // indirect
 	github.com/gookit/color v1.3.1
+	github.com/kevinburke/go-bindata v3.22.0+incompatible // indirect
 	github.com/mattn/go-isatty v0.0.12 // indirect
 	github.com/sergi/go-diff v1.1.0 // indirect
 	github.com/shirou/gopsutil v2.20.9+incompatible
--- a/segment_path.go
+++ b/segment_path.go
@@ -55,7 +55,7 @@ func (pt *path) string() string {
 	case Full:
 		return pt.env.getcwd()
 	case Folder:
-		return base(pt.env.getcwd(), pt.env)
+		return pt.getFolderPath()
 	default:
 		return fmt.Sprintf("Path style: %s is not available", style)
 	}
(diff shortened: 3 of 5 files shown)


docs/docs/segment-path.md
```diff
@@ -34,7 +34,7 @@ Display the current path.
 - windows_registry_icon: `string` - the icon to display when in the Windows registry - defaults to `\uE0B1`
 - style: `enum` - how to display the current path
 - mapped_locations: `map[string]string` - custom glyph/text for specific paths(only when `style` is set to `agnoster`,
-`agnoster_full` or `short`)
+`agnoster_full`, `agnoster_short`, `short`, or `folder`)
```
tradiff, line 37: `agnoster_short` already supported `mapped_locations`, it just wasn't mentioned in the docs.

segment_path.go
```diff
@@ -135,12 +135,17 @@ func (pt *path) getAgnosterShortPath() string {
 	root := pt.rootLocation()
 	base := base(pt.env.getcwd(), pt.env)
 	pathDepth := pt.pathDepth(pt.getShortPath())
-	if pathDepth == 1 {
+	if pathDepth <= 1 {
```
tradiff: IMO, this was a bug with `agnoster_short`. It didn't match the documentation, which reads "When **more than 1 level deep**, it renders one folder_icon followed by the name of the current folder separated by the folder_separator_icon.". 

The previous code was displaying the folder_icon when the depth was 0. On windows, if the cwd was `C:\`, the path segment would display `C: > .. > \` instead of the expected `C: > \`

The test `TestGetAgnosterShortPathWindowsZeroLevels` was created for this bug.

tradiff: This is failing tests on ubuntu, because [filepath.VolumeName](https://golang.org/pkg/path/filepath/#VolumeName) behaves differently on windows vs ubuntu. I'm going to revert this part of the PR, and address `agnoster_short` in another PR. I also believe it's acting unexpectedly on linux when you're at `/`.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/distatus/battery v0.10.1-0.20200722221337-7e1bf2bbb15c
 	github.com/go-ole/go-ole v1.2.4 // indirect
 	github.com/gookit/color v1.3.1
+	github.com/kevinburke/go-bindata v3.22.0+incompatible // indirect
```
JanDeDobbeleer, line 13: I'm wondering, isn't go mod complaining about this as it isn't referenced anywhere?

JanDeDobbeleer: @TravisTX maybe we can split the commits here and add this addition, and the use of `go generate` in the contributing guide. We can add a line in `main.go` to tell `go generate` to use `go-bindata`:

```golang
//go:generate go-bindata -o init.go init/
```

This needs to be on the first line, as indicated in the [documentation](https://blog.golang.org/generate).

JanDeDobbeleer: @TravisTX you know what, I'll facilitate this change. We need to make sure all the build scripts are aligned (same for homebrew).

tradiff: I'm new to Go, so I understood about half of what you just said 😄.  So, it's probably good that you're handling that. I look forward to seeing your change, so I can learn from it!

PR #267, opened 2020-12-24
feat: mapped_locations_enabled

tradiff:
### Prerequisites

### Description
Discussed in #258 

This PR separates the concept of path `style` from whether the `mapped_locations` are enabled. This separation of concerns allows us to remove some of the permeations from `style`, and generally leads to a set of configuration options that are easier to reason about.

There is a new property called `mapped_locations_enabled` which defaults to true, and can be used with any style.

At this point, the styles `short` and `full` are duplicates of each other, because the only difference between the two was whether they support `mapped_locations`. I realize `short` was used in lots of default themes, however I chose to keep the name `full` and hide the name `short`. There is nothing particularly short about this style anymore. The only "shortness" was related to `mapped_locations`, which has been decoupled from style. Compared to the other styles, it's actually one of the longer options 😛 

Note: a future PR will take this concept a step further and make `folder_separator_icon` apply to all styles. At that point, I believe the style `agnoster_full` can be retired, as it will also be a duplicate of `full`.

### DIFF
--- a/segment_path_test.go
+++ b/segment_path_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"math/rand"
 	"testing"
 
 	"github.com/distatus/battery"
@@ -129,7 +128,6 @@ func (env *MockedEnvironment) doGet(url string) ([]byte, error) {
 }
 
 const (
-	homeGates       = "/home/gates"
 	homeBill        = "/home/bill"
 	homeJan         = "/usr/home/jan"
 	homeBillWindows = "C:\\Users\\Bill"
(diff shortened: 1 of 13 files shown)


segment_path_test.go
```diff
@@ -624,3 +571,43 @@ func TestWritePathInfoUnixOutsideHomeOneLevels(t *testing.T) {
 	got := testWritePathInfo(home, "/mnt/folder/location", "/")
 	assert.Equal(t, want, got)
 }
+
+func TestGetPwd(t *testing.T) {
+	cases := []struct {
```
JanDeDobbeleer, line 576: I like this a lot

JanDeDobbeleer (approved):

PR #271, opened 2020-12-26
feat: allow custom folder separator for all styles

tradiff:
### Prerequisites

- [n/a] Docs have been added / updated (for bug fixes / features)

### Description
Discussed in #258 

Allow `folder_separator_icon` to be used in all themes, because why not?  If you want bicycle-delimited paths, you can now have them, even with the `full` style 🚴.

Originally I thought this was going to make `agnoster_full` and `full` be duplicates of each other, but that's not actually true, `agnoster_full` has a slight difference in that it removes the leading separator. Which makes a lot of sense on *nix file systems.

### DIFF
--- a/src/segment_path.go
+++ b/src/segment_path.go
@@ -85,12 +85,10 @@ func (pt *path) getAgnosterPath() string {
 
 func (pt *path) getAgnosterFullPath() string {
 	pwd := pt.getPwd()
-	pathSeparator := pt.env.getPathSeperator()
-	folderSeparator := pt.props.getString(FolderSeparatorIcon, pathSeparator)
-	if string(pwd[0]) == pathSeparator {
+	if string(pwd[0]) == pt.env.getPathSeperator() {
 		pwd = pwd[1:]
 	}
-	return strings.ReplaceAll(pwd, pathSeparator, folderSeparator)
+	return pt.replaceFolderSeparators(pwd)
 }
 
 func (pt *path) getAgnosterShortPath() string {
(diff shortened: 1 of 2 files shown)


src/segment_path.go
```diff
@@ -180,13 +190,12 @@ func (pt *path) rootLocation() string {
...
+	depth := 0
 	for _, part := range splitted {
 		if part != "" {
-			validParts = append(validParts, part)
+			depth++
 		}
 	}
-	depth := len(validParts)
 	return depth - 1
 }
```
tradiff, line 200: Small performance improvement here.  I wrote a quick benchmark to prove it:

```
goos: windows
goarch: amd64
pkg: github.com/jandedobbeleer/oh-my-posh3
BenchmarkPathDepthOld-12           38396             31473 ns/op
BenchmarkPathDepthNew-12           40593             29583 ns/op
PASS
ok      github.com/jandedobbeleer/oh-my-posh3   3.968s


goos: linux
goarch: amd64
pkg: github.com/jandedobbeleer/oh-my-posh3
BenchmarkPathDepthOld-12           41817             27922 ns/op
BenchmarkPathDepthNew-12           45936             26325 ns/op
PASS
ok      github.com/jandedobbeleer/oh-my-posh3   3.021s
```

Benchmark code if you're interested:
```go
func BenchmarkPathDepthOld(b *testing.B) {
	for n := 0; n < b.N; n++ {
		pwd := "/usr"
		for i := 0; i < 99; i++ {
			pwd += levelDir
		}
		env := new(MockedEnvironment)
		env.On("getPathSeperator", nil).Return("/")
		path := &path{
			env: env,
		}
		path.pathDepthOld(pwd)
	}
}

func BenchmarkPathDepthNew(b *testing.B) {
	for n := 0; n < b.N; n++ {
		pwd := "/usr"
		for i := 0; i < 99; i++ {
			pwd += levelDir
		}
		env := new(MockedEnvironment)
		env.On("getPathSeperator", nil).Return("/")
		path := &path{
			env: env,
		}
		path.pathDepth(pwd)
	}
}
```

PR #280, opened 2020-12-27
fix: *env.args.Shell not set in getShellName(cache never used)

lnu:
*env.args.Shell was never set in getShellName.

### Prerequisites

### Description

*env.args.Shell not set in getShellName(cache never used)

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -233,8 +233,9 @@ func (env *environment) getShellName() string {
 	if err != nil {
 		return unknown
 	}
-	shell := strings.Replace(name, ".exe", "", 1)
-	return strings.Trim(shell, " ")
+	// Cache the shell value to speed things up.
+	*env.args.Shell = strings.Trim(strings.Replace(name, ".exe", "", 1), " ")
+	return *env.args.Shell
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
My bad, in my use-case it was set in `main.go` when you provide the `--shell` flag. Obviously that's not the case everywhere.

src/environment.go
```diff
@@ -233,8 +233,8 @@ func (env *environment) getShellName() string {
 	if err != nil {
 		return unknown
 	}
-	shell := strings.Replace(name, ".exe", "", 1)
-	return strings.Trim(shell, " ")
+	*env.args.Shell = strings.Replace(name, ".exe", "", 1)
```
JanDeDobbeleer: I would set the name AFTER trimming it, no?
Meaning we do this:

```golang
shell := strings.Replace(name, ".exe", "", 1)
*env.args.Shell = strings.Trim(shell, " ")
return *env.args.Shell
```

lnu: I thought the same but was not sure about what was the replace doing.

PR #282, opened 2020-12-27
feat: timings

lnu:
### Prerequisites

### Description

Similar as Starship timings command:

103178055-ebdb5b00-487f-11eb-92de-fe01e30052a1


JanDeDobbeleer:
Isn't this a lot like `--debug`? I like this a lot more though.

lnu:
Yes, indeed. I also like it more than the debug. I think we could get rid of the debug stuff and keep only this one. It makes more sense and is easier to understand.

tradiff:
What about changing `--debug` to use this output instead of adding a new parameter? Just a thought. I agree, this output is nice.

JanDeDobbeleer:
I'm with @TravisTX on this one.

lnu:
Let's go for `debug` then.

JanDeDobbeleer:
@lnu please remove the vscode `launch.json`. There's no need for IDE specific files inside a repository, I thought I ignored this, will double check.

lnu:
> @lnu please remove the vscode `launch.json`. There's no need for IDE specific files inside a repository, I thought I ignored this, will double check.

I don't really agree on this point: : the `launch.json` file is really useful, otherwise I have to keep it and try not to delete it when switching branch/resetting branch/... . It also gives a default working environment without having to create the file manually.
But, it's your repo so I can of course remove it if you prefer.
And in the `.gitignore` it's explicitly set:
```
### VisualStudioCode ###
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
*.code-workspace
```

Which editor do you use?

JanDeDobbeleer:
@lnu alright, I gave this some thoughts and here's my final stance. I will allow it to be added. But not for the sake of your convenience as that can be solved by ignoring the file, but for the sake of anyone new trying to give a go at this repository. With that in mind it will be really useful to have an experience that works out of the box (also for our sake, given I usually have to assist) that works on vscode. I was about to extend the getting started guide so that it can be made even more accessible towards anyone looking for a first time contribution, maybe not experienced in go.

lnu:
> @lnu alright, I gave this some thoughts and here's my final stance. I will allow it to be added. But not for the sake of your convenience as that can be solved by ignoring the file, but for the sake of anyone new trying to give a go at this repository. With that in mind it will be really useful to have an experience that works out of the box (also for our sake, given I usually have to assist) that works on vscode. I was about to extend the getting started guide so that it can be made even more accessible towards anyone looking for a first time contribution, maybe not experienced in go.

I find it really useful when you can just clone the repo, open vs code and get started without having to mess around with config files, especially when you're not experienced with the language.  
If you want I can also add a couple of needed extensions in vscode(go, test explorer, ...). The bare minimum to be comfortable.

JanDeDobbeleer:
@lnu that would indeed be the way to go in that case!

lnu:
image

.vscode/launch.json
```diff
@@ -0,0 +1,43 @@
...
+        "-config=${workspaceRoot}/themes/jandedobbeleer.omp.json"
+      ]
+    },
+    {
+      "name": "Print init pwsh",
+      "type": "go",
+      "request": "launch",
+      "mode": "debug",
+      "program": "${workspaceRoot}/src",
+      "args": ["--print-init","-shell=pwsh","-config=${workspaceRoot}/themes/jandedobbeleer.omp.json"]
```
JanDeDobbeleer: can you make the args consistent: `--shell`, same for the commands above. 🙏🏻

packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -29,6 +28,7 @@ function Set-PoshContext {}
 
 function Set-GitStatus {
     if (Get-Command -Name "Get-GitStatus" -ErrorAction SilentlyContinue) {
+        [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification='Variable used later(not in this scope)')]
```
JanDeDobbeleer, line 31: I've been looking for this one!

src/main.go
```diff
@@ -127,8 +127,38 @@ func main() {
 		fmt.Println(Version)
 		return
 	}
+
+	if *args.Debug {
```
JanDeDobbeleer: is there a reason this is duplicated here rather than to use the flag to return the debug information after everything has been created? There's a lot of duplicate code here right now.

.vscode/launch.json
lnu: Which syntax do you prefer?
`-config=xxx` vs `-config xxx`  vs `--config=xxx` vs `--config xxx` 
The second one is not really pretty in launch.json:
```
 {
      "name": "Print init pwsh",
      "type": "go",
      "request": "launch",
      "mode": "debug",
      "program": "${workspaceRoot}/src",
      "args": [
        "-print-init",
        "-shell=pwsh",
        "-config=${workspaceRoot}/themes/jandedobbeleer.omp.json"
      ]
}
```
vs
```
{
      "name": "Print init pwsh",
      "type": "go",
      "request": "launch",
      "mode": "debug",
      "program": "${workspaceRoot}/src",
      "args": [
        "-print-init",
        "-shell",
        "pwsh",
        "-config",
        "${workspaceRoot}/themes/jandedobbeleer.omp.json"
      ]
}
```

JanDeDobbeleer: `--config=config`

src/main.go
lnu: Not really. At first, It was more different. But since your refactoring, it's the same. So yes, unless we need a specific renderer(which does not exist yet) we can merge them.

JanDeDobbeleer: I don't see a specific renderer coming anytime soon 😅

JanDeDobbeleer:
@lnu is this to be considered WIP?

lnu:
> @lnu is this to be considered WIP?

I added a small section to the doc on how to get code and install recommended extensions.

docs/docs/contributing-started.mdx
```diff
@@ -65,8 +65,8 @@ Make sure to remove this before submitting a PR as there's [a check][pr-go-mod]
 
 ## Get the source code
 
-The source is hosted on [Github][omp]. When you want to contribute, create a [fork][gh-fork] so you can make changes in your repository
-and create pull request in the official Oh my Posh repository.
+The source is hosted on [Github][omp]. When you want to contribute, create a [fork][gh-fork] so you can make changes in
```
lnu, line 68: > 120 chars line (markdownlint warning)

lnu:
> @lnu is this to be considered WIP?

We can consider this as version 1 :)

JanDeDobbeleer:
I'll go through it asap.

docs/docs/contributing-started.mdx
JanDeDobbeleer: 👍🏻

src/engine.go
```diff
@@ -159,6 +167,71 @@ func (e *engine) render() {
...
+				segmentTiming.stringValue = e.color.string()
+				e.color.buffer.Reset()
+			}
+
+			segmentTimings = append(segmentTimings, segmentTiming)
+		}
+	}
+
+	// 7 => (false)
+	nameMaxLength += 7
```
JanDeDobbeleer, line 221: what's the meaning of this?

src/main.go
```diff
@@ -149,7 +151,12 @@ func main() {
 		renderer:     renderer,
 		consoleTitle: title,
 	}
-	engine.render()
+
+	if *args.Debug {
+		engine.debug()
+	} else {
```
JanDeDobbeleer: change this for:

```golang
if *args.Debug {
    engine.debug()
    return
}
engine.render()
```

I have a strong opinion against using `if else` when an early return is more elegant.

src/engine.go
lnu: To align the columns, I get the length of the longest segment names + 7 char, corresponding to (false).  
(true) or (false) is added after the name of the segment to show if it's enabled or not.  
You can see the time took by a segment even if it's not enabled(like the languages segment for example)

lnu:
I just fixed the powerline end, before it was not there.

JanDeDobbeleer (approved):

PR #289, opened 2020-12-30
fix: $global:LASTEXITCODE

### LINKED ISSUE #288: $LASTEXITCODE does not preserve its value in PowerShell
### Prerequisites

- [ ] I looked for duplicate issues before submitting this one

### Description

When previous command exits with non-zero return code, the prompt function restores it too early so later commands will change it. The restoring of variable value should be the last command in prompt function. 

### Environment

- Oh my Posh version: 3.59.0
- Theme: doesn't make difference
- Operating System: Windows 10
- Shell: PowerShell (5.1 and 7.1)
- Terminal: Windows Terminal

### Steps to Reproduce

1. [First Step]
   ```
   ping.exe doesnotexist
   $LASTEXITCODE
   ```

**Expected behavior:** [What you expected to happen]
1

**Actual behavior:** [What actually happened]
0

tradiff:
### Prerequisites

- [ ] n/a Tests for the changes have been added (for bug fixes / features)
- [ ] n/a Docs have been added / updated (for bug fixes / features)

### Description

In the powershell prompt function, `Set-GitStatus` was clobbering the `$global:LASTEXITCODE`.  I moved `Set-GitStatus` and that seems to have resolved the issue.

closes #288 

### DIFF
--- a/packages/powershell/oh-my-posh/oh-my-posh.psm1
+++ b/packages/powershell/oh-my-posh/oh-my-posh.psm1
@@ -92,11 +92,11 @@ function Set-PoshPrompt {
         $standardOut = $process.StandardOutput.ReadToEnd()
         $process.WaitForExit()
         $standardOut
+        Set-GitStatus
         $global:LASTEXITCODE = $realLASTEXITCODE
         #remove temp variables
         Remove-Variable realLASTEXITCODE -Confirm:$false
         Remove-Variable lastCommandSuccess -Confirm:$false
-        Set-GitStatus
     }
     Set-Item -Path Function:prompt -Value $Prompt -Force
 }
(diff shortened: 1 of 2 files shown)


packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -92,11 +92,11 @@ function Set-PoshPrompt {
         $standardOut = $process.StandardOutput.ReadToEnd()
         $process.WaitForExit()
         $standardOut
+        Set-GitStatus
```
tradiff, line 95: @JanDeDobbeleer I'm not actually sure what `Set-GitStatus` does, so I'm not sure what to test to see if this broke anything. I find it odd that `Set-GitStatus` was/is called AFTER executing the ohmyposh executable. Can you shed some insight on this?

JanDeDobbeleer: Yes. It makes sure to populate the `posh-git` status when that module is present. It does that by itself by setting the prompt but as we override that, that context (and the accompanying autocomplete) is broken.

PR #297, opened 2020-12-31
feat: include  schema declaration in themes

tradiff:
### Prerequisites

- [ ] n/a Tests for the changes have been added (for bug fixes / features)

### Description

[The docs](https://ohmyposh.dev/docs/installation/#json-schema)  instruct the user how to add the schema to VS Code for `*.omp.json` files, however I feel that's not very discoverable and we can take this a step further. By including the `$schema` key in the default theme files, it improves the out of the box experience, so that users don't need to mess with their IDE settings.  I assume most users who are creating their own theme are probably starting with one of the default themes, so they will likely end up with this key in their custom theme if it's present in all default themes.

BTW there is another option, which is to submit our schema to [SchemaStore](https://github.com/SchemaStore/schemastore), which I believe VS Code and other IDEs periodically pull from. However, this carries additional responsibilities of maintaining the schema in that project, and I'm not sure how often the IDEs will refresh from this source.

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -329,21 +329,6 @@ and store it somehwere. Once adjusted to your liking, [change the prompt setting
 oh-my-posh --config ~/.mytheme.omp.json
 ```
 
-#### JSON Schema
-
-There's an easy to use [JSON schema][schema] available to validate your theme and have auto completion when editing.
-When using [Visual Studio Code][vscode], you can extend your settings file (F1 -> Preferences: Open Settings (JSON))
-by adding the following lines:
-
-```json
-"json.schemas": [
-  {
-    "fileMatch": ["*.omp.json"],
-    "url": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh3/main/themes/schema.json"
-  }
-]
-```
-
 🎉🎉🎉
 
 [wt]: https://github.com/microsoft/terminal
@@ -358,5 +343,3 @@ by adding the following lines:
 [brew]: /docs/homebrew
 [prompt]: /docs/installation#4-replace-your-existing-prompt
 [configuration]: /docs/configure
-[schema]: https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh3/main/themes/schema.json
-[vscode]: https://code.visualstudio.com/
--- a/themes/agnoster.omp.json
+++ b/themes/agnoster.omp.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh3/main/themes/schema.json",
   "blocks": [
     {
       "type": "prompt",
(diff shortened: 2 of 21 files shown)


tradiff:
If you want to move forward with this PR, let me know if you think we should also edit the docs to remove the instructions regarding the schema.  The more I think about it, they might be just unnecessary clutter if we add `$schema` to every default theme.

JanDeDobbeleer:
@TravisTX can we reference the local scheme? That way it stays in sync when doing changes.

tradiff:
@JanDeDobbeleer We could change it reference the local file like this:
```json
"$schema": "./schema.json",
```

That would make things nicer for someone who is contributing to this repo and making changes to the default themes.

But I think there is a much larger use case, for the end users.  Downloading themes (and the schema.json file) is a separate installation step. If a user has not re-downloaded their themes in a long time, they could have an older version of schema.json in their `~/.poshthemes` folder. If we used the local schema.json, they may not have accurate hints in their IDE. I feel this is likely a much larger use case, and the one we should be thinking about. Using a canonical URL helps with this use case.

Maybe in git, we could use `./schema.json`, but in the build steps for bundling the theme file, we could change to the canonical URL?

Or if that's too much complexity, we could use the URL in git, and if someone is changing `schema.json` they could temporarily swap the $schema to point to their local file to test it.  That's what I've been doing all along :)

```json
"$schema": "file:///C:/Projects/Personal/oh-my-posh3/themes/schema.json",
```

What are your thoughts?

JanDeDobbeleer:
I agree with all points made. In that case, the easiest is to use `"$schema": "./schema.json",` in the codebase, but swap out `./schema.json` with for example `https://github.com/JanDeDobbeleer/oh-my-posh3/blob/v3.66.8/themes/schema.json` (latest version at the time of writing) during build (`themes`) so that the exported schema always corresponds with the version of omp you installed.

tradiff:
@JanDeDobbeleer It's your project so we can do that if you want, but let me try one more time to explain the scenario I see.

Unless I'm misunderstanding something, using `../blob/v3.66.8..` would result in the same problem I described above. That is, the new user installs OMP, and downloads the themes, copies one of the themes to begin their journey of a custom theme. then a couple months later they have updated the OMP binary.  Their custom theme is now giving outdated information in the editor because it's still hard-coded to `../blob/v3.66.8..`.

Ideally it would be nice if this was always using the schema that matches their installed version of OMP. But that's not really possible, since this is just a text file the user has created. So that's why I think pointing to the schema at  `../main/..` is the best option. Then it matches the docs which are also always current.

JanDeDobbeleer:
You're right, I read though that too fast. In that case, we can merge this, right? For local development purposes I can point my theme to the local scheme idd.

tradiff:
Yep that works!

JanDeDobbeleer:
@TravisTX did you want to adjust the docs too? We can indeed remove that part.

JanDeDobbeleer:
@TravisTX if you find the time to rebase, I can still merge this 😉

PR #307, opened 2021-01-02
feat: hyperlink support in engine + hyperlink for path segment

lnu:
### Prerequisites

### Description

Hyperlink in git and path segment. Supported in windows terminal 1.5(in preview now and available this month for the stable one)

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -291,6 +291,19 @@ Oh my Posh mainly supports three different color types being
 
   `darkGray` `lightRed` `lightGreen` `lightYellow` `lightBlue` `lightMagenta` `lightCyan` `lightWhite`
 
+### Hyperlinks
+
+The engine has the ability to render hyperlinks. Your terminal has to support it and the option
+has to be enabled at the segment level. Hyperlink generation is disabled by default.
+
+#### Supported segments
+
+- [Path](segment-path.md)
+
+#### Supported terminals
+
+- [Terminal list](thttps://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
+
 ## Full Sample
 
 ```json
@@ -341,7 +354,8 @@ Oh my Posh mainly supports three different color types being
             "style": "folder",
             "ignore_folders": [
               "/super/secret/project"
-            ]
+            ],
+            "enable_hyperlink": false
           }
         },
         {
--- a/docs/docs/segment-path.md
+++ b/docs/docs/segment-path.md
@@ -37,6 +37,7 @@ Display the current path.
 is set to `true`)
 - mapped_locations_enabled: `boolean` - replace known locations in the path with the replacements before applying the
 style. defaults to `true`
+- enable_hyperlink: `boolean` - displays an hyperlink for the path - defaults to `false`
 
 ## Style
 
--- a/src/ansi_formats.go
+++ b/src/ansi_formats.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"strings"
 
 	"golang.org/x/text/unicode/norm"
@@ -21,6 +22,7 @@ type ansiFormats struct {
 	colorTransparent      string
 	escapeLeft            string
 	escapeRight           string
+	hyperlink             string
 }
 
 func (a *ansiFormats) init(shell string) {
@@ -40,6 +42,7 @@ func (a *ansiFormats) init(shell string) {
 		a.colorTransparent = "%%{\x1b[%s;49m\x1b[7m%%}%s%%{\x1b[m\x1b[0m%%}"
 		a.escapeLeft = "%{"
 		a.escapeRight = "%}"
+		a.hyperlink = "%%{\x1b]8;;%s\x1b\\%%}%s%%{\x1b]8;;\x1b\\%%}"
 	case bash:
 		a.linechange = "\\[\x1b[%d%s\\]"
 		a.left = "\\[\x1b[%dC\\]"
@@ -54,6 +57,7 @@ func (a *ansiFormats) init(shell string) {
 		a.colorTransparent = "\\[\x1b[%s;49m\x1b[7m\\]%s\\[\x1b[m\x1b[0m\\]"
 		a.escapeLeft = "\\["
 		a.escapeRight = "\\]"
+		a.hyperlink = "\\[\x1b]8;;%s\x1b\\\\\\]%s\\[\x1b]8;;\x1b\\\\\\]"
 	default:
 		a.linechange = "\x1b[%d%s"
 		a.left = "\x1b[%dC"
@@ -68,6 +72,7 @@ func (a *ansiFormats) init(shell string) {
 		a.colorTransparent = "\x1b[%s;49m\x1b[7m%s\x1b[m\x1b[0m"
 		a.escapeLeft = ""
 		a.escapeRight = ""
+		a.hyperlink = "\x1b]8;;%s\x1b\\%s\x1b]8;;\x1b\\"
 	}
 }
 
@@ -84,4 +89,16 @@ func (a *ansiFormats) lenWithoutANSI(text string) int {
 		count++
 	}
 	return count
+}
+
+func (a *ansiFormats) generateHyperlink(text string) string {
+	// hyperlink matching
+	results := findNamedRegexMatch("(?P<all>(?:\\[(?P<name>.+)\\])(?:\\((?P<url>.*)\\)))", text)
+	if len(results) != 3 {
+		return text
+	}
+	// build hyperlink ansi
+	hyperlink := fmt.Sprintf(a.hyperlink, results["url"], results["name"])
+	// replace original text by the new one
+	return strings.Replace(text, results["all"], hyperlink, 1)
 }
--- a/src/engine.go
+++ b/src/engine.go
@@ -82,6 +82,9 @@ func (e *engine) renderText(text string) {
 	if e.activeSegment.Background != "" {
 		defaultValue = fmt.Sprintf("<%s>\u2588</>", e.activeSegment.Background)
 	}
+
+	text = e.color.formats.generateHyperlink(text)
+
 	prefix := e.activeSegment.getValue(Prefix, defaultValue)
 	postfix := e.activeSegment.getValue(Postfix, defaultValue)
 	e.color.write(e.activeSegment.Background, e.activeSegment.Foreground, fmt.Sprintf("%s%s%s", prefix, text, postfix))
@@ -109,10 +112,9 @@ func (e *engine) renderBlockSegments(block *Block) string {
 		}
 		e.activeSegment = segment
 		e.endPowerline()
-		text := segment.stringValue
 		e.activeSegment.Background = segment.props.background
 		e.activeSegment.Foreground = segment.props.foreground
-		e.renderSegmentText(text)
+		e.renderSegmentText(segment.stringValue)
 	}
 	if e.previousActiveSegment != nil && e.previousActiveSegment.Style == Powerline {
 		e.writePowerLineSeparator(Transparent, e.previousActiveSegment.Background, true)
@@ -166,7 +168,7 @@ func (e *engine) render() {
 	e.write()
 }
 
-// debug will lool through your config file and output the timings for each segments
+// debug will loop through your config file and output the timings for each segments
 func (e *engine) debug() {
 	var segmentTimings []SegmentTiming
 	largestSegmentNameLength := 0
--- a/src/settings.go
+++ b/src/settings.go
@@ -34,6 +34,8 @@ const (
 	Left BlockAlignment = "left"
 	// Right aligns right
 	Right BlockAlignment = "right"
+	// EnableHyperlink enable hyperlink
+	EnableHyperlink Property = "enable_hyperlink"
 )
 
 // Block defines a part of the prompt with optional segments
--- a/src/ansi_formats_test.go
+++ b/src/ansi_formats_test.go
@@ -23,3 +23,57 @@ func TestLenWithoutAnsi(t *testing.T) {
 		assert.Equal(t, 5, strippedLength)
 	}
 }
+
+func TestGenerateHyperlinkNoUrl(t *testing.T) {
+	cases := []struct {
+		Text      string
+		ShellName string
+		Expected  string
+	}{
+		{Text: "sample text with no url", ShellName: zsh, Expected: "sample text with no url"},
+		{Text: "sample text with no url", ShellName: pwsh, Expected: "sample text with no url"},
+		{Text: "sample text with no url", ShellName: bash, Expected: "sample text with no url"},
+	}
+	for _, tc := range cases {
+		a := ansiFormats{}
+		a.init(tc.ShellName)
+		hyperlinkText := a.generateHyperlink(tc.Text)
+		assert.Equal(t, tc.Expected, hyperlinkText)
+	}
+}
+
+func TestGenerateHyperlinkWithUrl(t *testing.T) {
+	cases := []struct {
+		Text      string
+		ShellName string
+		Expected  string
+	}{
+		{Text: "[google](http://www.google.be)", ShellName: zsh, Expected: "%{\x1b]8;;http://www.google.be\x1b\\%}google%{\x1b]8;;\x1b\\%}"},
+		{Text: "[google](http://www.google.be)", ShellName: pwsh, Expected: "\x1b]8;;http://www.google.be\x1b\\google\x1b]8;;\x1b\\"},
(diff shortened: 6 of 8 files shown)


lnu:
Work in progress. First draft to see how it works. I have to check how to integrate this cleanly.

JanDeDobbeleer:
The goal is that you can click the branch/tag/commit and link to the repo?

lnu:
> The goal is that you can click the branch/tag/commit and link to the repo?

Yes it's a first draft to see how it looks like.  
Right now the ansi sequence is embedded in the segment for testing purpose. I was thinking of using the markdown syntax for url to decouple the ansi rendering from the segment. I have a more or less working regex but it's not easy to cover all scenarios.

src/segment_git.go
```diff
@@ -131,12 +135,23 @@ func (g *git) string() string {
...
+	g.repo.url = g.getGitCommandOutput("remote", "get-url", upstream)
+
 	// branchName
 	if g.repo.upstream != "" && g.props.getBool(DisplayUpstreamIcon, false) {
-		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol())
+		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol(g.repo.url))
+	}
+	if g.repo.upstream != "" && enableHyperlink {
+		url := strings.Replace(g.repo.url, ".git", "", -1)
+		fmt.Fprintf(buffer, "]8;;%s\\%s]8;;\\", fmt.Sprintf("%s/tree/%s", url, g.repo.HEAD), g.repo.PRETTYHEAD)
```
tradiff: The `/tree/` url seems github specific. At my work, we use Azure Devops and the url is different there. Your code results in a 404 page for my Azure Devops projects.

Example:
git remote get-url origin:
`https://AcmeSoftware@dev.azure.com/AcmeSoftware/OhMyPosh/_git/oh-my-posh3`
branch:
`feat/hyperlink`
url:
`https://AcmeSoftware@dev.azure.com/AcmeSoftware/OhMyPosh/_git/oh-my-posh3?version=GBfeat%2Fmy-feature`

 The url construction code would be something like this (untested):
```go
fmt.Sprintf("%s?version=GB%s", url, g.repo.HEAD)
```

Maybe we could do a check to see if the url contains `github.com` or `dev.azure.com` and apply the proper url building? And for other domains we don't know about, just omit the hyperlink feature. Or make the hyperlink point to the git repo url directly (no branch)

tradiff:
>  Supported in windows terminal 1.5(in preview now and available this month for the stable one)

BTW I'm using windows terminal 1.4.3243.0 and the clickable link works for me ¯\\\_(ツ)_/¯

src/segment_git.go
lnu: I was expecting some issues like that. Yeah of course let's keep a list of main providers and point to the repo when unknown.
Thanks for the feedback!

lnu: I did a first draft of multi providers support, if you have the opportunity to test it.

src/segment_git.go
```diff
@@ -131,12 +146,38 @@ func (g *git) string() string {
...
+		var urlTemplate string
+		// check for a match in the providers map
+		for i, val := range gitProvidersMap {
+			if strings.Contains(baseURL.Host, i) {
+				urlTemplate = val
+				break
+			}
+		}
+		if urlTemplate != "" {
+			// if provider found, open the branch
```
tradiff:
```suggestion
			// if provider found, open the branch

			// normalize azure devops url to change from
			// https://{account}@dev.azure.com/{account}/{project}/_git/{repo}
			// to
			// https://dev.azure.com/{account}/{project}/_git/{repo}
			repoURL = replaceAllString("//[^@]*@", repoURL, "//")						
```

This suggestion makes the code work for me, for azure devops repositories.  I suspect you'll also have problems with people who have cloned using ssh instead of https. I found a couple other repos which are doing something similar to what you're doing, which may be useful to review.

[dotnet/docfx](https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/common/common.js) is probably the easiest to follow
[eamodio/vscode-gitlens](https://github.com/eamodio/vscode-gitlens/tree/main/src/git/remotes) has this functionality too

Hope this helps!

lnu: Thanks. I thought the host property was stripping the account and returning the hostname only. I’ll check this.

lnu:
I still have an issue for path not rendered correctly when mapped_locations is used and a unicode character is set.
Example:
```
{
          "type": "path",
          "style": "powerline",
          "powerline_symbol": "\uE0B0",
          "foreground": "#ffffff",
          "background": "#ff479c",
          "properties": {
            "prefix": " \uE5FF ",
            "style": "full",
            "mapped_locations": [
              ["D:\\dev", "\uf962 "]
            ],
            "mapped_locations_enabled": true,
            "enable_hyperlink": true
          }
}
```
rendering:
image

With something else:
image


I think it can be related to that windows terminal issue: https://github.com/microsoft/terminal/issues/3546

lnu:
and support for opening uri is coming:https://github.com/microsoft/terminal/pull/7526
we're going too fast 🤣 

to track hyperlink progress: https://github.com/microsoft/terminal/issues/5001

lnu:
And  to test hyperlink in the terminal:
`printf '\e]8;;https://github.com/lnu/oh-my-posh3.git\e\\main\e]8;;\e\\'`

JanDeDobbeleer:
@lnu I was looking for that unicode bug the other day. Sometimes I think this whole repo is "too soon" for MS as we're constantly battling with incomplete features whereas users expect all of this to "just work". Or, we're right on time and will provide appropriate focus 😅

lnu:
tested in gnome terminal and everything works(uri + http), and no glitch:
image

src/segment_git.go
```diff
@@ -3,19 +3,32 @@ package main
...
 	"strings"
 )
 
+var (
+	// Map for git service providers
+	gitProvidersMap map[string]string = map[string]string{
+		"github.com":       "%s/tree/%s",
+		"azure.com":        "%s?version=GB%s",
+		"visualstudio.com": "%s?version=GB%s",
+		"gitlab.com":       "%s/-/tree/%s",
```
JanDeDobbeleer: we should also add bitbucket, and that said, add support for these remotes too while we're at it. Maybe that got me thinking, this can be part of that configuration. The remote url match is hardcoded but can be rewritten and easily extended this way. Maybe that's out of scope for this change, but still a good idea.

src/segment_git.go
```diff
@@ -131,12 +146,44 @@ func (g *git) string() string {
...
-		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol())
+		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol(g.repo.url))
 	}
-	fmt.Fprintf(buffer, "%s", g.repo.HEAD)
+	// default name
+	repoHEAD := g.repo.PRETTYHEAD
+	if g.repo.upstream != "" && enableHyperlink {
+		// parse url to and strips extra info
+		baseURL, err := url.Parse(g.repo.url)
+		if err == nil {
```
JanDeDobbeleer: Do an early return here rather than embedding it in a whole block.

src/segment_git.go
```diff
@@ -131,12 +146,44 @@ func (g *git) string() string {
...
 	// branchName
 	if g.repo.upstream != "" && g.props.getBool(DisplayUpstreamIcon, false) {
-		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol())
+		fmt.Fprintf(buffer, "%s", g.getUpstreamSymbol(g.repo.url))
 	}
-	fmt.Fprintf(buffer, "%s", g.repo.HEAD)
+	// default name
+	repoHEAD := g.repo.PRETTYHEAD
+	if g.repo.upstream != "" && enableHyperlink {
+		// parse url to and strips extra info
```
JanDeDobbeleer: this whole logic should be a separate function

src/segment_git.go
```diff
@@ -131,12 +146,44 @@ func (g *git) string() string {
...
+					break
+				}
+			}
+			// construct url(remove .git from url(github))
+			// removing extra data could be done using a regex stored in the map
+			repoURL := fmt.Sprintf("%s://%s%s", baseURL.Scheme, baseURL.Host, strings.ReplaceAll(baseURL.Path, ".git", ""))
+			if urlTemplate != "" {
+				// if provider found, open the branch
+				repoHEAD = fmt.Sprintf("]8;;%s\\%s]8;;\\", fmt.Sprintf(urlTemplate, repoURL, g.repo.HEAD), g.repo.PRETTYHEAD)
+			} else {
```
JanDeDobbeleer: once this logic is in a separate function, the else can be removed when you return a string and write this in the function above

src/segment_git.go
```diff
@@ -131,12 +146,44 @@ func (g *git) string() string {
...
+					urlTemplate = val
+					break
+				}
+			}
+			// construct url(remove .git from url(github))
+			// removing extra data could be done using a regex stored in the map
+			repoURL := fmt.Sprintf("%s://%s%s", baseURL.Scheme, baseURL.Host, strings.ReplaceAll(baseURL.Path, ".git", ""))
+			if urlTemplate != "" {
+				// if provider found, open the branch
+				repoHEAD = fmt.Sprintf("]8;;%s\\%s]8;;\\", fmt.Sprintf(urlTemplate, repoURL, g.repo.HEAD), g.repo.PRETTYHEAD)
```
JanDeDobbeleer: the formatting needs to become part of `ansi_formats.go`. Wondering if we want to expose this just like we do for color rendering overrides?

lnu: yes of course => https://github.com/JanDeDobbeleer/oh-my-posh3/pull/307#issuecomment-753464224

src/segment_git.go
lnu: I think we should handle hyperlink at a global level(https://github.com/JanDeDobbeleer/oh-my-posh3/pull/307#issuecomment-753464224).  
Outputting something like a template and detect it in the ansi renderer.

JanDeDobbeleer: Yes!

lnu: Now that i’m back to work I have less time but I’ll try do finish it by the end of the week.
Url template
Global hyperlink
Extra: support ssh cloned repo

lnu:
I did some refactoring regarding the hyperlink support. I think we should split the functionality in two:
- this pull request with:
  - hyperlink support by the engine using url markdown syntax
  - Hyperlink support for path segment
  - Initial hyperlink support for git segment, limited to http cloned repo
- a second pull with 
  - A refactoring of the git url parsing (like it's done in git lens for ex.)
  - Support for ssl cloned repo
  - Support for custom config

Otherwise, it will take too long to have a first working version. What do you think @JanDeDobbeleer ?

JanDeDobbeleer:
@lnu I would then maybe, for the sake of simplicity, only limit to the engine and path segment for 1. Leave git on a separate track.

src/ansi_formats.go
```diff
@@ -69,6 +70,7 @@ func (a *ansiFormats) init(shell string) {
 		a.escapeLeft = ""
 		a.escapeRight = ""
 	}
+	a.hyperlink = "\x1b]8;;%s\x1b\\%s\x1b]8;;\x1b\\"
```
JanDeDobbeleer: this needs to be escaped differently for zsh and bash, otherwise the line of the prompt won't be correct anymore and rendering is ruined. You can validate this by doing a tab completion using `cd` for example and notice the prompt just a bit off after pressing `TAB` the first time.

src/engine.go
```diff
@@ -82,6 +83,16 @@ func (e *engine) renderText(text string) {
 	if e.activeSegment.Background != "" {
 		defaultValue = fmt.Sprintf("<%s>\u2588</>", e.activeSegment.Background)
 	}
+
+	// hyperlink matching
+	results := findNamedRegexMatch("(?P<all>(?:\\[(?P<name>.+)\\])(?:\\((?P<url>.*)\\)))", text)
```
JanDeDobbeleer: I would make this a function in `ansi_formats.go` so it can also be tested properly.

src/engine.go
```diff
@@ -166,7 +176,7 @@ func (e *engine) render() {
 	e.write()
 }
 
-// debug will lool through your config file and output the timings for each segments
+// debug will loop through your config file and output the timings for each segments
```
JanDeDobbeleer, line 171: that made me laugh actually :laughing

src/engine.go
```diff
@@ -217,6 +227,7 @@ func (e *engine) debug() {
 		segmentName := fmt.Sprintf("%s(%t)", segment.name, segment.enabled)
 		e.renderer.print(fmt.Sprintf("%-*s - %3d ms - %s\n", largestSegmentNameLength, segmentName, duration, segment.stringValue))
 	}
+
```
JanDeDobbeleer: unneeded change unless our listing expects this

src/settings.go
```diff
@@ -14,6 +14,7 @@ type Settings struct {
 	ConsoleTitle         bool              `json:"console_title"`
 	ConsoleTitleStyle    ConsoleTitleStyle `json:"console_title_style"`
 	ConsoleTitleTemplate string            `json:"console_title_template"`
+	EnableHyperlink      bool              `json:"enable_hyperlink"`
```
JanDeDobbeleer: Is there a side-effect of having this enabled but not supported in your terminal?

lnu: yes I think your prompt won't be happy if the terminal you use don't support hyperlink. I read various threads and it seems there's no way to know if the terminal supports it or not. So, by default it's disabled.

src/segment_git.go
lnu: resolved by moving git segment to another pull request

lnu:
Still have to change the call in engine.go to generateHyperlink. Calling through color is a bit strange.

src/settings.go
lnu: If you check [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), a majority of terminal already support it.

JanDeDobbeleer:
> Calling through color is a bit strange

I agree, but `ansi_formats.go` makes sense.

lnu:
> > Calling through color is a bit strange
> 
> I agree, but `ansi_formats.go` makes sense.

I think that in case of Color the reponsibilities are mixed. The ansi renderer should take care of converting everything to ansi text but here Color, which outputs ansi, is called directly from the engine. Should not AnsiColor be called only from the ansi renderer?

JanDeDobbeleer (changes requested):
What I'm missing is an addition to the docs about this like colors. Because I can create a text segment/prefix/postfix, that contains a link which will be rendered when hyperlinks are enabled.

lnu:
> What I'm missing is an addition to the docs about this like colors. Because I can create a text segment/prefix/postfix, that contains a link which will be rendered when hyperlinks are enabled.

Yes but be careful that right now the engine replaces only hyperlink in segment text, not prefix or postfix.

JanDeDobbeleer:
> Yes but be careful that right now the engine replaces only hyperlink in segment text, not prefix or postfix.

True. Which is fine I guess, this needs to be more contextual anyways. It's not because we can that we should.

docs/docs/segment-path.md
```diff
@@ -37,6 +37,7 @@ Display the current path.
 is set to `true`)
 - mapped_locations_enabled: `boolean` - replace known locations in the path with the replacements before applying the
 style. defaults to `true`
+- enable_hyperlink: `boolean` - displays an hyperlink for the path - defaults to `false`
```
JanDeDobbeleer, line 40: Can't believe I missed this, do we want to keep it at this level or higher up at `Settings`?

lnu: Now, the engine detects url from any segments. Hyperlink can be enabled/disabled for each segment and it takes care of rendering the text or the url markdown.  
On the other side, it could be enabled at global level and implying that all segments supporting url will render them.
Both have their pros and cons.  
It’s a matter of taste :).

JanDeDobbeleer: I believe both can be an option, but let's keep it like this for now. Amazing work.

docs/docs/configuration.md
```diff
@@ -291,6 +291,19 @@ Oh my Posh mainly supports three different color types being
...
+The engine has the ability to render hyperlinks. Your terminal has to support it and the option
+has to be enabled at the segment level. Hyperlink generation is disabled by default.
+
+#### Supported segments
+
+- [Path](segment-path.md)
+
+#### Supported terminals
+
+- [Terminal list](thttps://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
```
imba-tjd, line 305: `t`

JanDeDobbeleer: fixed, thx.

PR #320, opened 2021-01-05
git improvements

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/src/ansi_color.go
+++ b/src/ansi_color.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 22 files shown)


src/environment.go
```diff
@@ -152,6 +152,9 @@ func (env *environment) getPlatform() string {
 }
 
 func (env *environment) runCommand(command string, args ...string) (string, error) {
+	if cmd, ok := env.commands[command]; ok {
```
JanDeDobbeleer, line 163: @lnu this might interest you, I removed need to put the operator in control when it comes to command caching. Something I only thought of afterwards.

src/ansi_color.go
```diff
@@ -46,7 +45,7 @@ func getColorFromName(colorName string, isBackground bool) (string, error) {
 
 // AnsiColor writes colorized strings
 type AnsiColor struct {
-	buffer  *bytes.Buffer
+	builder strings.Builder
```
lnu, line 48: Noticeable?

JanDeDobbeleer: I was reading through string concatenation performance benchmarks and this one came out on top. So it made sense to swap it. They made it compatible with every other way to ensure an easy transition.

PR #334, opened 2021-01-10
fix: lock around shared map write

### LINKED ISSUE #333: Concurrent read/write map issue
### Prerequisites

### Description

Since one of the refactoring(https://github.com/JanDeDobbeleer/oh-my-posh3/commit/c2bc901a414ac640e88a06ee2ddc0af7cc97a341) I'm getting randomly a concurrent read/write map error, which causes omp to crash with a fatal panic.

Map supports concurrent read but not writes. If a map is used concurrently, we should lock write access(like https://github.com/lnu/oh-my-posh3/blob/16c92aff2d38d25ac3c0cf79ef02b1e74e5c9092/src/regex.go)
I found it while testing languages. For this I put several files from different languages in the same folder.

### Environment

- Oh my Posh version: latest
- Theme: default
- Operating System: all
- Shell: all
- Terminal: all

### Steps to Reproduce

**Expected behavior:** [What you expected to happen]
No crash

**Actual behavior:** [What actually happened]
Random crash

lnu:
### Prerequisites

### Description

Fix https://github.com/JanDeDobbeleer/oh-my-posh3/issues/333 

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/distatus/battery"
@@ -65,10 +66,14 @@ type environmentInfo interface {
 }
 
 type environment struct {
-	args     *args
-	cwd      string
-	commands map[string]string
+	args *args
+	cwd  string
 }
+
+var (
+	commands map[string]string = make(map[string]string)
+	lock                       = sync.Mutex{}
+)
 
 func (env *environment) getenv(key string) string {
 	return os.Getenv(key)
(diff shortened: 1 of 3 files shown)


src/environment.go
```diff
@@ -65,11 +66,15 @@ type environmentInfo interface {
...
 type environment struct {
-	args     *args
-	cwd      string
-	commands map[string]string
+	args *args
+	cwd  string
 }
 
+var (
+	commands map[string]string = make(map[string]string)
```
JanDeDobbeleer, line 74: is there a specific reason we need to move it outside the `environment` scope?

lnu: I tried by just adding the lock but the issue was still there. Honestly, I have to investigate. I don't know if it was because the variable was initialized elsewhere and the lock here. I'll dig to see the reason behind.

PR #342, opened 2021-01-14
feat: envvars in console title

tradiff:
### Prerequisites

### Description

Add the ability to include environment variables in the console title.

Closes #341 

### DIFF
--- a/src/console_title.go
+++ b/src/console_title.go
@@ -28,14 +28,6 @@ const (
 	incorrectTitleTemplate = "unable to create title based on template"
 )
 
-// TitleTemplateContext defines what can be aded to the title when using the template
-type TitleTemplateContext struct {
-	Root   bool
-	Path   string
-	Folder string
-	Shell  string
-}
-
 func (t *consoleTitle) getConsoleTitle() string {
 	var title string
 	switch t.settings.ConsoleTitleStyle {
@@ -52,12 +44,21 @@ func (t *consoleTitle) getConsoleTitle() string {
 }
 
 func (t *consoleTitle) getTemplateText() string {
-	context := &TitleTemplateContext{
-		Root:   t.env.isRunningAsRoot(),
-		Path:   t.getPwd(),
-		Folder: base(t.getPwd(), t.env),
-		Shell:  t.env.getShellName(),
+	context := make(map[string]interface{})
+
+	context["Root"] = t.env.isRunningAsRoot()
+	context["Path"] = t.getPwd()
+	context["Folder"] = base(t.getPwd(), t.env)
+	context["Shell"] = t.env.getShellName()
+
+	// load environment variables into the map
+	envVars := map[string]string{}
+	matches := findAllNamedRegexMatch(`\.Env\.(?P<ENV>[^ \.}]*)`, t.settings.ConsoleTitleTemplate)
+	for _, match := range matches {
+		envVars[match["ENV"]] = t.env.getenv(match["ENV"])
 	}
+	context["Env"] = envVars
+
 	tmpl, err := template.New("title").Parse(t.settings.ConsoleTitleTemplate)
 	if err != nil {
 		return invalidTitleTemplate
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -64,20 +64,23 @@ to `{{ .Shell }} in {{ .Folder }}`
 
 - `folder`: show the current folder name
 - `path`: show the current path
-- `template`: show the current path
+- `template`: show a custom template
 
 ### Console Title Template
 
 You can create a more custom console title with the use of `"console_title_style" = "template"`.
 When this is set, a `console_title_template` is also expected, otherwise the title will remain empty.
-Under the hood this uses go's text/template feature and offers a few standard properties to work with.
+Under the hood this uses go's [text/template][go-text-template] feature and offers a few standard
+properties to work with.
 
 - `.Root`: `boolean` - is the current user root/admin or not
 - `.Path`: `string` - the current working directory
 - `.Folder`: `string` - the current working folder
 - `.Shell`: `string` - the current shell name
+- `.Env.VarName`: `string` - Any environment variable where `VarName` is the environment variable name
 
 A `boolean` can be used for conditional display purposes, a `string` can be displayed.
+
 The following examples illustrate possible contents for `console_title_template`, provided
 the current working directory is `/usr/home/omp` and the shell is `zsh`.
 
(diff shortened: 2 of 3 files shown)


tradiff:
I can think of 3 ways we can solve this, each one has it's own pros and cons:
1. Use Golang's `ExpandEnv` function. 
    - Pros: simple
    - Cons: template syntax is different from the other properties: `$USERDOMAIN`
2. Add a `map[string]string` to the `TitleTemplateContext` struct. 
    - Pros: template syntax would use go's template syntax
    - Cons: the template syntax is not very intuitive: `{{index .EnvVars \"USERDOMAIN\"}}`
3. Add the specific properties that were requested, `.Host` and `.Domain`.
    - Pros: most intuitive
    - Cons: not as flexible

I went with option 1, but I can change it if there is a strong preference for another option.

lnu:
`ExpandEnv` is the most straightforward solution. 👍

peter-dolkens:
Would it make more sense to do a 2-phase approach:

- Apply existing templates
- Replace any remaining `{{.VARS}}` with `$VARS`
- Run through ExpandEnv

This would have the advantage or preserving a single templating format

Other than that - this PR looks like it would do exactly what I was after.

tradiff:
@peter-dolkens with that idea, we would lose some of the existing template behaviors, such as the "if" actions.  You can see it in the [docs here](https://ohmyposh.dev/docs/configure#console-title-template)

peter-dolkens:
I'd imagine those would be evaluated in step 1, no?

I'm not familiar with go or its templating language, so I'm not sure if it will choke on unresolved tokens - I was more imagining it as a post-processing step on any unresolved tokens.

tradiff:
@peter-dolkens Oh, I understand now.
So your idea is a template could be `"{{.USERDOMAIN}} :: {{.Folder}}{{if .Root}} :: root{{end}} :: {{.Shell}}"`
1. The go template processing would change this to `{{.USERDOMAIN}} :: /usr/home/omp :: root :: zsh`
2. We replace any remaining `[[.*}}` with `$*` resulting in `$USERDOMAIN :: /usr/home/omp :: root :: zsh`
3. Use `ExpandEnv` resulting in `MyCompany :: /usr/home/omp :: root :: zsh`

I don't think this would work, as go will return an error at step 1 when it encounters unexpected expression because USERDOMAIN is not a property of the object.

We could probably accomplish the same input you described in another way. The code could could get all environment variables, loop through them updating the template to replace `{{.var_name}}` with the value of the `var_name`.  After that has occurred, we could use go's template for the rest (step 1 above).

@JanDeDobbeleer Do you have any thoughts?

lnu:
I quickly tested this one and it outputs the right value without changing the template syntax:
```
func (t *consoleTitle) getTemplateText() string {

	context := make(map[string]interface{})

	context["Root"] = t.env.isRunningAsRoot()
	context["Path"] = t.getPwd()
	context["Folder"] = base(t.getPwd(), t.env)
	context["Shell"] = t.env.getShellName()

	// load environment variables into the map
	envVars := os.Environ()
	for _, envVar := range envVars {
		splittedVar := strings.Split(envVar, "=")
		context[splittedVar[0]] = splittedVar[1]
	}

	tmpl, err := template.New("title").Parse(t.settings.ConsoleTitleTemplate)
	if err != nil {
		return invalidTitleTemplate
	}
	buffer := new(bytes.Buffer)
	defer buffer.Reset()
	err = tmpl.Execute(buffer, context)
	if err != nil {
		return incorrectTitleTemplate
	}
	return buffer.String()
}
```

Instead of a struct, a map is filled with custom properties and environment variables(yes all, maybe we filter it?). Then the template is applied and expected values are in the map.

tradiff:
Very nice, @lnu I like that option, still 100% leveraging go's templates.  I'll update the PR.

lnu:
@TravisTX  Maybe having all env variables is a bit overkill but it won't hurt the perf. I'm adding the console timing to `--debug` to monitor the time it takes to generate.

image

https://github.com/JanDeDobbeleer/oh-my-posh3/pull/345

lnu:
One more thing, we can have a collision between the built-in name(ex. Path) and the environment variables. Juste have to decide if we rename or if we override.

src/console_title.go
```diff
@@ -52,12 +45,23 @@ func (t *consoleTitle) getConsoleTitle() string {
...
 
 func (t *consoleTitle) getTemplateText() string {
-	context := &TitleTemplateContext{
-		Root:   t.env.isRunningAsRoot(),
-		Path:   t.getPwd(),
-		Folder: base(t.getPwd(), t.env),
-		Shell:  t.env.getShellName(),
+	context := make(map[string]interface{})
+
+	context["Root"] = t.env.isRunningAsRoot()
```
lnu, line 49: Check if custom variables should override the environment ones? or if a prefix should be added to environment variables(env_path vs path)?

tradiff: I went with ignoring conflicting environment variables (line 60)

lnu: sorry, overlooked the continue, my bad.

src/console_title.go
```diff
@@ -52,12 +45,23 @@ func (t *consoleTitle) getConsoleTitle() string {
...
+	context["Path"] = t.getPwd()
+	context["Folder"] = base(t.getPwd(), t.env)
+	context["Shell"] = t.env.getShellName()
+
+	// load environment variables into the map
+	envVars := os.Environ()
```
JanDeDobbeleer: If we do this, it needs to be part of environment. But hold your horses as I want to test a different approach first.

JanDeDobbeleer:
A high level comment as I looked at this and came with the following suggestion:

```golang
envVars := map[string]string{}
matches := findAllNamedRegexMatch(`\.Env\.(?P<ENV>[^ \.}]*)`, t.settings.ConsoleTitleTemplate)
for _, match := range matches {
	envVars[match["ENV"]] = t.env.getenv(match["ENV"])
}
context["Env"] = envVars
```

This way we can use templates like `{{.Folder}}{{if .Root}} :: Admin{{end}} :: {{.Shell}} :: {{.Env.USER}}` where `.Env.USER` is expanded to whatever the variable holds and avoid collisions and expanding the entire environment into memory.

lnu:
> A high level comment as I looked at this and came with the following suggestion:
> 
> ```go
> envVars := map[string]string{}
> matches := findAllNamedRegexMatch(`\.Env\.(?P<ENV>[^ \.}]*)`, t.settings.ConsoleTitleTemplate)
> for _, match := range matches {
> 	envVars[match["ENV"]] = t.env.getenv(match["ENV"])
> }
> context["Env"] = envVars
> ```
> 
> This way we can use templates like `{{.Folder}}{{if .Root}} :: Admin{{end}} :: {{.Shell}} :: {{.Env.USER}}` where `.Env.USER` is expanded to whatever the variable holds and avoid collisions and expanding the entire environment into memory.

Why not. We had this discussion of loading all variables or not. I don’t think there will be a difference in performance, but it’s cleaner to load only the needed variables.

JanDeDobbeleer:
> Why not.

Not feeling the motivation here @lnu 😄

tradiff:
@JanDeDobbeleer / @lnu done.  Thanks for the assistance. I think we landed at a much better solution 💯

JanDeDobbeleer (approved):

JanDeDobbeleer:
@TravisTX @lnu I wouldn't have reached this point without you, so 👊🏻

peter-dolkens:
Wow - I walk away for a day and magic happens - thanks guys, looks like a great outcome!

PR #344, opened 2021-01-14
fix: return the right error type from runCommand

lnu:
The error returned by runCommand must be cast to exec.ExitError

### Prerequisites

### Description

When calling runCommand, the error returned is of type exitError instead of commandError.

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -170,7 +170,12 @@ func (env *environment) runCommand(command string, args ...string) (string, erro
 	}
 	out, err := exec.Command(command, args...).Output()
 	if err != nil {
-		return "", err
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			return "", &commandError{
+				err:      exitErr.Error(),
+				exitCode: exitErr.ExitCode(),
+			}
+		}
 	}
 	return strings.TrimSpace(string(out)), nil
 }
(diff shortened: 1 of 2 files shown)


lnu:
Still have to fix the test, takes more time than the fix itself ...

lnu:
Tested on windows and linux(for the dotnet 145 exit code).

JanDeDobbeleer:
Why not remove the `commandError` and switch to `exitError` in the `dotnet` segment?

lnu:
> Why not remove the `commandError` and switch to `exitError` in the `dotnet` segment?

I wanted to do that a first, but then I thought it was maybe better to not « leak » wrapped command exception from helper class.
Moreover it’s easier to mock commandError than exitError. I tried but did not find any easy way to do it.

src/environment.go
```diff
@@ -170,7 +170,12 @@ func (env *environment) runCommand(command string, args ...string) (string, erro
 	}
 	out, err := exec.Command(command, args...).Output()
 	if err != nil {
```
JanDeDobbeleer, line 172: There's a nicer way to do this:

```golang
if exitErr, ok := err.(*exec.ExitError); ok {
    return "", &commandError{
 	err:      exitErr.Error(),
 	exitCode: exitErr.ExitCode(),
    }
}
```

JanDeDobbeleer:
> it’s easier to mock commandError than exitError

You are right. Just noticed that as well.

src/environment.go
lnu: I saw this syntax but I had no idea how to use it. Thanks!

lnu:
I've also added a safety check when casting the error type in language segment(belt and braces).

PR #348, opened 2021-01-16
feat: release notes hyperlink for all languages

lnu:
If version enabled and hyperlink enabled(global level), an hyperlink to the release notes of the current version is generated
The regex and the url template can be modified in the json template.

### Prerequisites

### Description

All languages segment will generate an hyperlink pointing to the release notes the current version(if version and hyperlink enabled). 

temp

### DIFF
--- a/src/segment_language.go
+++ b/src/segment_language.go
@@ -1,8 +1,36 @@
 package main
 
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
 type loadContext func()
 
 type inContext func() bool
+
+type version struct {
+	full        string
+	major       string
+	minor       string
+	patch       string
+	regex       string
+	urlTemplate string
+}
+
+func (v *version) parse(versionInfo string) error {
+	values := findNamedRegexMatch(v.regex, versionInfo)
+	if len(values) == 0 {
+		return errors.New("cannot parse version string")
+	}
+
+	v.full = values["version"]
+	v.major = values["major"]
+	v.minor = values["minor"]
+	v.patch = values["patch"]
+	return nil
+}
 
 type language struct {
 	props        *properties
@@ -11,8 +39,7 @@ type language struct {
 	commands     []string
 	executable   string
 	versionParam string
-	versionRegex string
-	version      string
+	version      *version
 	exitCode     int
 	loadContext  loadContext
 	inContext    inContext
@@ -42,8 +69,21 @@ func (l *language) string() string {
 	if !l.hasCommand() {
 		return l.props.getString(MissingCommandTextProperty, MissingCommandText)
 	}
-	l.setVersion()
-	return l.version
+
+	err := l.setVersion()
+	if err != nil {
+		return ""
+	}
+
+	// build release notes hyperlink
+	if l.props.getBool(EnableHyperlink, false) && l.version.urlTemplate != "" {
+		version, err := TruncatingSprintf(l.version.urlTemplate, l.version.full, l.version.major, l.version.minor, l.version.patch)
+		if err != nil {
+			return l.version.full
+		}
+		return version
+	}
+	return l.version.full
 }
 
 func (l *language) enabled() bool {
@@ -77,16 +117,18 @@ func (l *language) hasLanguageFiles() bool {
 	return true
 }
 
-// getVersion returns the version and exit code returned by the executable
-func (l *language) setVersion() {
+// setVersion parses the version string returned by the command
+func (l *language) setVersion() error {
 	versionInfo, err := l.env.runCommand(l.executable, l.versionParam)
 	if exitErr, ok := err.(*commandError); ok {
 		l.exitCode = exitErr.exitCode
-		return
+		return errors.New("error executing command")
 	}
-	values := findNamedRegexMatch(l.versionRegex, versionInfo)
-	l.exitCode = 0
-	l.version = values["version"]
+	err = l.version.parse(versionInfo)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 // hasCommand checks if one of the commands exists and sets it as executable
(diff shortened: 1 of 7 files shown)


lnu:
Don't know if it's very useful but it's a good sample for the hyperlink and I had fun doing it :)

src/segment_language.go
```diff
@@ -43,6 +57,20 @@ func (l *language) string() string {
...
 	}
 	l.setVersion()
+
+	// if regex found no version
+	if l.version == "" {
+		return ""
+	}
+
+	// build release notes hyperlink
+	if l.props.getBool(EnableHyperlink, false) {
```
JanDeDobbeleer: We might want to validate the presence of a `urlTemplate` too

src/segment_language.go
```diff
@@ -84,9 +112,12 @@ func (l *language) setVersion() {
 		l.exitCode = exitErr.exitCode
 		return
 	}
+
 	values := findNamedRegexMatch(l.versionRegex, versionInfo)
-	l.exitCode = 0
 	l.version = values["version"]
+	l.major = values["major"]
```
JanDeDobbeleer: is there ever a changelog URL where the need to split this is present?

JanDeDobbeleer (changes requested):

src/segment_language.go
lnu: Yes, in the end if it's blank it will return the version(error will be returned from TruncatingSprintf). But an extra check before calling it would be better.

src/segment_language.go
lnu: the url to the release notes are quite different for each languages. Some like golang takes only major.minor, others take all and dotnet is also different. So it's more flexible to follow semver and extract all the data we need.  
go=> https://golang.org/doc/go1.15
.net=>https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0.2/5.0.2.md
python=>https://www.python.org/downloads/release/python-385/
node=>https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#10.19.0

src/segment_language.go
```diff
@@ -33,6 +43,10 @@ const (
 	MissingCommandTextProperty Property = "missing_command_text"
 	// MissingCommandText displays empty string by default
 	MissingCommandText string = ""
+	// VersionRegex sets the regex used to detect the language version
+	VersionRegex string = "version_regex"
```
JanDeDobbeleer: I would not expose this. Most users won't even know what to do if this fails and we're faster anyways to create a quick fix.

lnu: Makes sense. Was added for more flexibility but in the end, probably 90% of users just want it to work.

lnu: version_regex and url_template removed from config. As you said, it's better to keep it internal and update it if an issue arise.

src/segment_language.go
```diff
@@ -43,6 +53,24 @@ func (l *language) string() string {
...
+
+	// if regex found no version
+	if l.version == "" {
+		return ""
+	}
+
+	// build release notes hyperlink
+	if l.props.getBool(EnableHyperlink, false) {
+		// if url template is blank in the confi file
+		if l.urlTemplate == "" {
```
JanDeDobbeleer: this check can be in the initial if statement as the return below does the same.

src/segment_language.go
```diff
@@ -13,9 +19,13 @@ type language struct {
 	versionParam string
 	versionRegex string
 	version      string
+	major        string
```
JanDeDobbeleer: I would opt for an additional struct here called `version`

```golang
type version struct {
    major string
    minor string
    patch string
}

func (v *version) string() string {
    return fmt.Sprintf("%s.%s.%s", v.major, v.minor, v.patch)
}
```

JanDeDobbeleer: That one can even have an init function and hold the regex for example.

src/segment_language.go
```diff
@@ -1,18 +1,40 @@
...
 type loadContext func()
 
 type inContext func() bool
 
+type version struct {
+	full         string
+	major        string
+	minor        string
+	patch        string
+	versionRegex string
```
JanDeDobbeleer: rename to `regex` as `version.versionRegex` is a mouthful :-)

src/segment_language.go
```diff
@@ -1,18 +1,40 @@
...
+	urlTemplate  string
+}
+
+func (v *version) string() string {
+	version, err := TruncatingSprintf(v.urlTemplate, v.full, v.major, v.minor, v.patch)
```
JanDeDobbeleer: The url creation should not be in here, all it should do is return the version, everything else should be handled in `language.go`.

src/segment_language.go
```diff
@@ -43,7 +65,17 @@ func (l *language) string() string {
 		return l.props.getString(MissingCommandTextProperty, MissingCommandText)
 	}
 	l.setVersion()
-	return l.version
+
+	// if regex found no version
+	if l.version.full == "" {
```
JanDeDobbeleer: this is an internal property only, removing the hyperlink creation in `version.string()` will allow you to not do this check at all. On top of that, it's not even useful as you return the value you validate it contains meaning the last return does exactly the same.

src/segment_language.go
```diff
@@ -84,9 +116,14 @@ func (l *language) setVersion() {
 		l.exitCode = exitErr.exitCode
 		return
 	}
-	values := findNamedRegexMatch(l.versionRegex, versionInfo)
-	l.exitCode = 0
-	l.version = values["version"]
+
+	values := findNamedRegexMatch(l.version.versionRegex, versionInfo)
```
JanDeDobbeleer: this should be a function of the `version` struct

JanDeDobbeleer (changes requested):

src/segment_language.go
lnu: I don't get those points related to the generation of the hyperlink. In the end, to build the hyperlink all properties from version are needed. If it's not done from a function on `version`, then I need to get the data from those properties inside `language.string()`?

JanDeDobbeleer (approved):

PR #349, opened 2021-01-16
feat: cache compiled regex

lnu:
### Prerequisites

### Description

Compiled regex caching

### DIFF
--- a/src/init/omp.ps1
+++ b/src/init/omp.ps1
@@ -45,7 +45,7 @@ function global:Set-PoshGitStatus {
     $config = $global:PoshSettings.Theme
     $cleanPWD = $PWD.ProviderPath.TrimEnd("\")
     $cleanPSWD = $PWD.ToString().TrimEnd("\")
-    $standardOut = @(&$omp --error="$errorCode" --pwd="$cleanPWD" --pswd="$cleanPSWD" --execution-time="$executionTime" --config="$config")
+    $standardOut = @(&$omp --error="$errorCode" --pwd="$cleanPWD" --pswd="$cleanPSWD" --execution-time="$executionTime" --config="$config" 2>&1)
     # Restore initial encoding
     [Console]::OutputEncoding = $originalOutputEncoding
     # the ouput can be multiline, joining these ensures proper rendering by adding line breaks with `n
--- a/src/regex.go
+++ b/src/regex.go
@@ -1,9 +1,36 @@
 package main
 
-import "regexp"
+import (
+	"regexp"
+	"sync"
+)
+
+var (
+	regexCache     map[string]*regexp.Regexp = make(map[string]*regexp.Regexp)
+	regexCacheLock                           = sync.Mutex{}
+)
+
+func getCompiledRegex(pattern string) *regexp.Regexp {
+	// try in cache first
+	re := regexCache[pattern]
+	if re != nil {
+		return re
+	}
+
+	// should we panic or return the error?
+	re = regexp.MustCompile(pattern)
+
+	// lock for concurrent access and save the compiled expression in cache
+	regexCacheLock.Lock()
+	regexCache[pattern] = re
+	regexCacheLock.Unlock()
+
+	return re
+}
 
 func findNamedRegexMatch(pattern, text string) map[string]string {
-	re := regexp.MustCompile(pattern)
+	// error ignored because mustCompile will cause a panic
+	re := getCompiledRegex(pattern)
 	match := re.FindStringSubmatch(text)
 	result := make(map[string]string)
 	if len(match) == 0 {
@@ -19,7 +46,7 @@ func findNamedRegexMatch(pattern, text string) map[string]string {
 }
 
 func findAllNamedRegexMatch(pattern, text string) []map[string]string {
-	re := regexp.MustCompile(pattern)
+	re := getCompiledRegex(pattern)
 	match := re.FindAllStringSubmatch(text, -1)
 	var results []map[string]string
 	if len(match) == 0 {
@@ -40,6 +67,11 @@ func findAllNamedRegexMatch(pattern, text string) []map[string]string {
 }
 
 func replaceAllString(pattern, text, replaceText string) string {
-	re := regexp.MustCompile(pattern)
+	re := getCompiledRegex(pattern)
 	return re.ReplaceAllString(text, replaceText)
+}
+
+func matchString(pattern, text string) bool {
+	re := getCompiledRegex(pattern)
+	return re.MatchString(text)
 }
--- a/src/environment_windows_win32.go
+++ b/src/environment_windows_win32.go
@@ -4,7 +4,6 @@ package main
 
 import (
 	"fmt"
-	"regexp"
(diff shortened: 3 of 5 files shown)


src/regex.go
```diff
@@ -19,7 +33,13 @@ func findNamedRegexMatch(pattern, text string) map[string]string {
 }
 
 func findAllNamedRegexMatch(pattern, text string) []map[string]string {
-	re := regexp.MustCompile(pattern)
+	re := regexCache[pattern]
+	if re == nil {
```
JanDeDobbeleer: this is the same functionality twice, can we split this in a separate function?

JanDeDobbeleer (changes requested):

src/regex.go
lnu: Yes, I put the compile and cache in a separate function

src/regex.go
```diff
@@ -1,9 +1,39 @@
...
+
+var (
+	regexCache     map[string]*regexp.Regexp = make(map[string]*regexp.Regexp)
+	regexCacheLock                           = sync.Mutex{}
+)
+
+func getCompiledRegex(pattern string, mustCompile bool) (*regexp.Regexp, error) {
+	// try in cache first
+	re := regexCache[pattern]
+	if re == nil {
```
JanDeDobbeleer: switch this block to `!=` and return early.

src/regex.go
```diff
@@ -1,9 +1,39 @@
...
+		regexCacheLock.Lock()
+		regexCache[pattern] = re
+		regexCacheLock.Unlock()
+	}
+	return re, nil
+}
 
 func findNamedRegexMatch(pattern, text string) map[string]string {
-	re := regexp.MustCompile(pattern)
+	re, _ := getCompiledRegex(pattern, true)
```
JanDeDobbeleer: if we never catch the error, we should ignore it higher up and only return the compiled regex

lnu: > if we never catch the error, we should ignore it higher up and only return the compiled regex

the error is ignored because true will cause a panic by calling mustCompile instead of compile

JanDeDobbeleer: Must compile is only used once, and that function in the end returns a boolean. The boolean is unneeded if we always do a `Compile` and return `nil` when `Compile` returns an error rather than panicking. I tried to figure out of there's a severe performance issue, but couldn't find info on that right away.

lnu: Was thinking about it, wouldn't it be better to panic(MustCompile all the time) if a regex does not compile? I don't know if it's better to fail silently or ensure that all regex can be compiled safely.

JanDeDobbeleer: These are all under our control, right? Meaning if this happens, we have a missing test or unvalidated change? That does provide additional support for my remark on the other PR where we should in that case never export the regex.

lnu: A regex can be set by the user for mapped_locations I think. So, you can make omp panic if you fail in your regex. But, in the end, you want something that fails and tells you you missed something, no?

=>not mapped_locations but ignore_folders

JanDeDobbeleer: Yes, but with some additional as to why it failed. So I would opt to fail OMP, but with proper context rather than a stack trace. Because you know who will need to assist whenever that happens 🙋🏻‍♂️

lnu: Let's try to make an intelligible error when you fail at regex(like I often do ^^)

lnu: Ok I tried something. Looks like this when compiling the regex fail:
image
You know at least the regex failed and why.

JanDeDobbeleer: I like this!!

src/regex.go
```diff
@@ -1,9 +1,45 @@
...
+)
+
+var (
+	regexCache     map[string]*regexp.Regexp = make(map[string]*regexp.Regexp)
+	regexCacheLock                           = sync.Mutex{}
+)
+
+func getCompiledRegex(pattern string) *regexp.Regexp {
+	defer func() {
+		if err := recover(); err != nil {
```
JanDeDobbeleer: early return ;-)

JanDeDobbeleer:
@lnu should I still consider a WIP or are we almost good to go?

lnu:
> @lnu should I still consider a WIP or are we almost good to go?

Are there still some changes to do?

src/regex.go
JanDeDobbeleer: and this one @lnu

JanDeDobbeleer:
> Are there still some changes to do?

@lnu well, the build is failing 😓

lnu:
> > Are there still some changes to do?
> 
> @lnu well, the build is failing 😓

Sorry, let me check this asap

Update: I rebased on your changes but now with the multiline we get the full stacktrace:
image
I'm not sure i'll be able to clear everything.

JanDeDobbeleer:
@lnu that's OK, at least we have a clear message about the issue.

lnu:
> @lnu that's OK, at least we have a clear message about the issue.

I tried one more thing:
image

I recover in SetStringValues and set "error" and active for the failing segment. Better? 
We could also generate a log file containing the stack trace with an hyperlink to the file, but for another pull. What do you think?

JanDeDobbeleer:
> What do you think?

@lnu I think we can ship this one. I like the log idea, can be useful for more things, but indeed, in a later stage.

.vscode/launch.json
```diff
@@ -10,15 +10,15 @@
       "request": "launch",
       "mode": "debug",
       "program": "${workspaceRoot}/src",
-      "args": ["--config=${workspaceRoot}/themes/jandedobbeleer.omp.json"]
+      "args": ["-config=${workspaceRoot}/themes/jandedobbeleer.omp.json"]
```
JanDeDobbeleer: why not `--`? that's the way that's "normal"

src/init/omp.ps1
```diff
@@ -45,7 +45,7 @@ function global:Set-PoshGitStatus {
     $config = $global:PoshSettings.Theme
     $cleanPWD = $PWD.ProviderPath.TrimEnd("\")
     $cleanPSWD = $PWD.ToString().TrimEnd("\")
-    $standardOut = @(&$omp --error="$errorCode" --pwd="$cleanPWD" --pswd="$cleanPSWD" --execution-time="$executionTime" --config="$config")
+    $standardOut = @(&$omp "--config=$config" "--error=$errorCode" "--pwd=$cleanPWD" "--pswd=$cleanPSWD" "--execution-time=$executionTime" 2>&1)
```
JanDeDobbeleer: what's the advantage of `"--config=$config"` over `--config="$config"`?

JanDeDobbeleer:
@lnu 2 comments as those changes seem new and I'm curious.

.vscode/launch.json
lnu: mistake. Should not have been pushed.

src/init/omp.ps1
lnu: rebasing issue.

JanDeDobbeleer (approved):

PR #363, opened 2021-01-21
fix: check permissions before attempting chmod

RobCannon:
### Prerequisites

- [x ] I have read and understand the `CONTRIBUTING` guide
- [ x] The commit message follows the [conventional commits][cc] guidelines

### Description

On a linux system, when you import the powershell module that was installed for AllUsers, the current user will usually not have the permissions to change the executable permission on the executable.  That returns an error every time the module is imported.

This fix will check to see if the executable permission is already set and skip running chmod.  It will also make sure the current user has the correct permissions to change the executable permissions before attempting to run chmod.  If the module is installed by a sudo user, as long as the module is imported by that user, the file permissions will get set correctly.

### DIFF
--- a/packages/powershell/oh-my-posh/oh-my-posh.psm1
+++ b/packages/powershell/oh-my-posh/oh-my-posh.psm1
@@ -17,7 +17,28 @@ function Get-PoshCommand {
 # Set the right binary to executable before doing anything else
 if ($PSVersionTable.PSEdition -eq "Core" -and !$IsWindows) {
     $executable = Get-PoshCommand
-    Invoke-Expression -Command "chmod +x $executable"
+    if (Test-Path $executable) {
+        # Check the permissions on the file
+        $permissions = ((ls -l $executable) -split ' ')[0]
+        if ((id -u) -eq 0) {
+            # Running as root, give global permissions if needed
+            $hasWrite = $permissions[2] -eq 'w'
+            $hasExecutable = $permissions[3] -eq 'x'
+            if ($hasWrite -and -not $hasExecutable) {
+                Invoke-Expression -Command "chmod g+x $executable"
+            }
+        }
+        else {
+            $hasWrite = $permissions[8] -eq 'w'
+            $hasExecutable = $permissions[9] -eq 'x'
+            if ($hasWrite -and -not $hasExecutable) {
+                Invoke-Expression -Command "chmod +x $executable"
+            }
+        }
+    }
+    else {
+      Write-Warning "Executable at $executable was not found"
+    }
 }
 
 function Set-PoshPrompt {
--- a/packages/powershell/oh-my-posh/oh-my-posh.psm1
+++ b/packages/powershell/oh-my-posh/oh-my-posh.psm1
@@ -14,30 +14,37 @@ function Get-PoshCommand {
     return $poshCommand
 }
 
-# Set the right binary to executable before doing anything else
-if ($PSVersionTable.PSEdition -eq "Core" -and !$IsWindows) {
+function SetExecutablePermissions {
+    # Set the right binary to executable before doing anything else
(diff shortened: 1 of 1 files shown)


packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -17,7 +17,28 @@ function Get-PoshCommand {
...
+        else {
+            $hasWrite = $permissions[8] -eq 'w'
+            $hasExecutable = $permissions[9] -eq 'x'
+            if ($hasWrite -and -not $hasExecutable) {
+                Invoke-Expression -Command "chmod +x $executable"
+            }
+        }
+    }
+    else {
+      Write-Warning "Executable at $executable was not found"
```
JanDeDobbeleer: How can this happen?

packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -17,7 +17,28 @@ function Get-PoshCommand {
...
+        $permissions = ((ls -l $executable) -split ' ')[0]
+        if ((id -u) -eq 0) {
+            # Running as root, give global permissions if needed
+            $hasWrite = $permissions[2] -eq 'w'
+            $hasExecutable = $permissions[3] -eq 'x'
+            if ($hasWrite -and -not $hasExecutable) {
+                Invoke-Expression -Command "chmod g+x $executable"
+            }
+        }
+        else {
```
JanDeDobbeleer: rather than `if..else...` return early in the previous statements. Unless there's an actual use-case try to avoid `if else` as much a you can.

JanDeDobbeleer (changes requested):
See my comments in-line 🙏🏻

packages/powershell/oh-my-posh/oh-my-posh.psm1
RobCannon: It shouldn't happen, unless the install is messed up.  But I always check for the presence of a file before I operate on it.  Otherwise the error message that results would be difficult to diagnose.

JanDeDobbeleer: Fair point even though it technically can't happen without a failing installation. If you address the other comment we can merge this 👍🏻

packages/powershell/oh-my-posh/oh-my-posh.psm1
RobCannon: This code is not in a function, so you can't really return.  If it even worked, it would skip all of the other function declarations and the Set-Aliases later in the script.

JanDeDobbeleer: You're right, didn't take that into account, reviews using the mobile app aren't all that... In that case, let's make it an internal function and call that.

RobCannon (commented):
Code has been moved to private function

packages/powershell/oh-my-posh/oh-my-posh.psm1
```diff
@@ -14,10 +14,38 @@ function Get-PoshCommand {
...
+
+    # Check the permissions on the file
+    $permissions = ((ls -l $executable) -split ' ')[0]  # $permissions will be something like '-rw-r--r--'
+    if ((id -u) -eq 0) {
+        # Running as root, give global executable permissions if needed
+        $hasWrite = $permissions[2] -eq 'w'
+        $hasExecutable = $permissions[3] -eq 'x'
+        if ($hasWrite -and -not $hasExecutable) {
+            Invoke-Expression -Command "chmod g+x $executable"
+        }
```
JanDeDobbeleer, line 39: Return here as well, and the else can be removed.

JanDeDobbeleer (changes requested):

PR #383, opened 2021-02-06
feat: add the User and Host properties to the console title template.

beppler:
### Prerequisites

### Description

This pull request adds the User and Host to the console title template so we can make a cross-platform configuration.

### DIFF
--- a/src/console_title.go
+++ b/src/console_title.go
@@ -50,6 +50,11 @@ func (t *consoleTitle) getTemplateText() string {
 	context["Path"] = t.getPwd()
 	context["Folder"] = base(t.getPwd(), t.env)
 	context["Shell"] = t.env.getShellName()
+	context["User"] = t.env.getCurrentUser()
+	context["Host"] = ""
+	if host, err := t.env.getHostName(); err == nil {
+		context["Host"] = host
+	}
 
 	// load environment variables into the map
 	envVars := map[string]string{}
--- a/src/console_title_test.go
+++ b/src/console_title_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -11,6 +12,7 @@ func TestGetConsoleTitle(t *testing.T) {
 		Style         ConsoleTitleStyle
 		Template      string
 		Root          bool
+		User          string
 		Cwd           string
 		PathSeperator string
 		ShellName     string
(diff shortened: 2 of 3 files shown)


src/console_title_test.go
```diff
@@ -49,6 +59,8 @@ func TestGetConsoleTitle(t *testing.T) {
 		env.On("isRunningAsRoot", nil).Return(tc.Root)
 		env.On("getShellName", nil).Return(tc.ShellName)
 		env.On("getenv", "USERDOMAIN").Return("MyCompany")
+		env.On("getCurrentUser", nil).Return("MyUser")
+		env.On("getHostName", nil).Return("MyHost", nil)
```
JanDeDobbeleer, line 64: Can we create a test where we expect a host name but the command errors? Check if the template works.

JanDeDobbeleer (changes requested):

beppler:
I wrote the test and confirmed that template works without the Host value, but it prints "<no value>" instead.
So I changed the code to put an empty string on Host if the getHostName returns an error.

Do you prefer that I use a new commit or amend the previous?

src/console_title_test.go
beppler: I wrote the test and confirmed that template works without the Host value, but it prints "&lt;no value&gt;" instead.
So I changed the code to put an empty string on Host if the getHostName returns an error.

beppler:
I wrote the test and confirmed that template works without the Host value, but it prints "&lt;no value&gt;" instead.
So I changed the code to put an empty string on Host if the getHostName returns an error.

src/console_title.go
```diff
@@ -50,6 +50,12 @@ func (t *consoleTitle) getTemplateText() string {
 	context["Path"] = t.getPwd()
 	context["Folder"] = base(t.getPwd(), t.env)
 	context["Shell"] = t.env.getShellName()
+	context["User"] = t.env.getCurrentUser()
+	if host, err := t.env.getHostName(); err == nil {
+		context["Host"] = host
+	} else {
```
JanDeDobbeleer: I only use if else when there's no other option. Can we initialize it as empty first and override it when there's no error?

JanDeDobbeleer (changes requested):

src/console_title.go
beppler: Done.

JanDeDobbeleer:
Thank you for this 🙏🏻

beppler:
Thank you for your patience and orientation.

PR #448, opened 2021-02-24
feat(session): default user override

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/src/segment_session.go
+++ b/src/segment_session.go
@@ -6,12 +6,14 @@ import (
 )
 
 type session struct {
-	props        *properties
-	env          environmentInfo
-	UserName     string
-	ComputerName string
-	SSHSession   bool
-	Root         bool
+	props           *properties
+	env             environmentInfo
+	UserName        string
+	DefaultUserName string
+	ComputerName    string
+	SSHSession      bool
+	Root            bool
+	templateText    string
 }
 
 const (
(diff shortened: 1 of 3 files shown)


jetersen (approved):
LGTM 🚀 Thanks for adding feat!

src/segment_session.go
```diff
@@ -29,12 +31,26 @@ const (
...
+	s.DefaultUserName = s.getDefaultUser()
+	segmentTemplate := s.props.getString(SegmentTemplate, "")
+	if segmentTemplate != "" {
+		s.Root = s.env.isRunningAsRoot()
+		template := &textTemplate{
+			Template: segmentTemplate,
+			Context:  s,
+		}
+		s.templateText = template.render()
+		return len(s.templateText) > 0
```
jetersen, line 51: 😆 Logic bug, so this broke the current behavior as it returns too early and is always enabled.
Since `getUsername` always returns the username so the `segmentTemplate` is always greater than 0.

I tried fixing the code to have `segmentTemplate` as in have `getUsername` returning `""`\
when `showDefaultUser == false && defaultUserName == username`.

Sadly that breaks a lot of tests 😅

So the enabled would look something like this.
```go
	if segmentTemplate != "" {
		s.Root = s.env.isRunningAsRoot()
		template := &textTemplate{
			Template: segmentTemplate,
			Context:  s,
		}
		s.templateText = template.render()
		return len(s.templateText) > 0
	}
	return false
```

Not sure where the logic for `showDefaultUser` should be located as in you would want to show host or session host depending on other logic. 😓

PR #466, opened 2021-02-28
feat: include folders

### LINKED ISSUE #465: Negative lookahead in "ignore_folders": "invalid Perl syntax" error
### Prerequisites

### Description

When using a regular expression containing a negative lookahead in the `"ignore_folders"` property of a segment, I get the following error:
```
oh-my-posh fatal error rendering kubectl segment:regexp: Compile(`^(?!C:\\dev\\).*$`): error parsing regexp: invalid or unsupported Perl syntax: `(?!`
```

Here is the segment description:
```json
{
	"type": "kubectl",
	"style": "powerline",
	"powerline_symbol": "\uE0B0",
	"foreground": "#000000",
	"background": "#ebcc34",
	"properties": {
		"prefix": " \uFD31 ",
		"template": "{{.Context}}{{if .Namespace}} :: {{.Namespace}}{{end}}",
		"ignore_folders": [
			"(?!C:\\\\dev\\\\).*"
		]
	}
}
```

### Environment

- Oh my Posh version: 3.101.1
- Theme: Custom theme
- Operating System: Windows 10 Pro 20H2 19042.804
- Shell: PowerShell 5.1.19041.610
- Terminal: Windows Terminal 1.5.10411.0

tradiff:
### Prerequisites

### Description

Add new segment property, `include_folders`.  This works in conjunction with `ignore_folders` to determine whether a segment will be rendered or not based on the cwd.

Closes #465

### DIFF
--- a/src/properties.go
+++ b/src/properties.go
@@ -17,7 +17,11 @@ const (
 	Postfix Property = "postfix"
 	// ColorBackground color the background or foreground when a specific color is set
 	ColorBackground Property = "color_background"
-	// IgnoreFolders folders to ignore and not run the segment logic
+	// IncludeFolders folders to be included for the segment logic
+	IncludeFolders Property = "include_folders"
+	// ExcludeFolders folders to be excluded for the segment logic
+	ExcludeFolders Property = "exclude_folders"
+	// IgnoreFolders duplicate of ExcludeFolders
 	IgnoreFolders Property = "ignore_folders"
 	// DisplayVersion show the version number or not
 	DisplayVersion Property = "display_version"
--- a/src/segment.go
+++ b/src/segment.go
@@ -112,15 +112,45 @@ func (segment *Segment) getValue(property Property, defaultValue string) string 
 	return defaultValue
 }
 
-func (segment *Segment) shouldIgnoreFolder(cwd string) bool {
-	if value, ok := segment.Properties[IgnoreFolders]; ok {
-		list := parseStringArray(value)
-		for _, element := range list {
-			pattern := fmt.Sprintf("^%s$", element)
-			matched := matchString(pattern, cwd)
-			return matched
+func (segment *Segment) shouldIncludeFolder(cwd string) bool {
+	cwdIncluded := segment.cwdIncluded(cwd)
+	cwdExcluded := segment.cwdExcluded(cwd)
+	return (cwdIncluded && !cwdExcluded)
+}
+
+func (segment *Segment) cwdIncluded(cwd string) bool {
+	value, ok := segment.Properties[IncludeFolders]
+	if !ok {
+		// IncludeFolders isn't specified, everything is included
+		return true
+	}
+
+	list := parseStringArray(value)
+
+	if len(list) == 0 {
+		// IncludeFolders is an empty array, everything is included
+		return true
+	}
+
+	return segment.cwdMatchesOneOf(cwd, list)
+}
+
+func (segment *Segment) cwdExcluded(cwd string) bool {
+	value, ok := segment.Properties[ExcludeFolders]
+	if !ok {
+		value = segment.Properties[IgnoreFolders]
+	}
+	list := parseStringArray(value)
+	return segment.cwdMatchesOneOf(cwd, list)
+}
+
+func (segment *Segment) cwdMatchesOneOf(cwd string, regexes []string) bool {
+	for _, element := range regexes {
+		pattern := fmt.Sprintf("^%s$", element)
+		matched := matchString(pattern, cwd)
+		if matched {
+			return true
 		}
-		return false
 	}
 	return false
 }
@@ -214,7 +244,7 @@ func (segment *Segment) setStringValue(env environmentInfo, cwd string) {
 		segment.active = true
 	}()
 	err := segment.mapSegmentWithWriter(env)
-	if err != nil || segment.shouldIgnoreFolder(cwd) {
+	if err != nil || !segment.shouldIncludeFolder(cwd) {
 		return
 	}
 	if segment.enabled() {
--- a/themes/schema.json
+++ b/themes/schema.json
@@ -185,14 +185,33 @@
               "description": "https://ohmyposh.dev/docs/configure#postfix",
               "default": " "
             },
-            "ignore_folders": {
+            "include_folders": {
               "type": "array",
-              "title": "Ignore rendering in these folders",
-              "description": "https://ohmyposh.dev/docs/configure#ignore-folders",
+              "title": "If specified, segment will only render in these folders",
+              "description": "https://ohmyposh.dev/docs/configure#include--exclude-folders",
               "default": [],
               "items": {
                 "type": "string"
               }
+            },
+            "exclude_folders": {
+              "type": "array",
+              "title": "Exclude rendering in these folders",
+              "description": "https://ohmyposh.dev/docs/configure#include--exclude-folders",
+              "default": [],
+              "items": {
+                "type": "string"
+              }
+            },
+            "ignore_folders": {
+              "type": "array",
+              "title": "Deprecated - please use `exclude_folders` instead",
+              "description": "https://ohmyposh.dev/docs/configure#include--exclude-folders",
+              "default": [],
+              "items": {
+                "type": "string"
+              },
+              "deprecated": true
             }
           }
         }
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -254,7 +254,8 @@ You can use these on any segment, the engine is responsible for adding them corr
 
 - prefix: `string`
 - postfix: `string`
-- ignore_folders: `[]string`
+- include_folders: `[]string`
+- exclude_folders: `[]string`
 
 ##### Prefix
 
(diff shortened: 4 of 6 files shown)


src/segment.go
```diff
@@ -112,16 +112,37 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
 	return defaultValue
 }
 
-func (segment *Segment) shouldIgnoreFolder(cwd string) bool {
-	if value, ok := segment.Properties[IgnoreFolders]; ok {
+func (segment *Segment) shouldIncludeFolder(cwd string) bool {
```
tradiff, line 115: This function was switched to return the positive case instead of the negative, to make it easier to reason about.

themes/schema.json
```diff
@@ -185,6 +185,15 @@
...
+            "include_folders": {
+              "type": "array",
+              "title": "If specified, segment will only render in these folders",
+              "description": "https://ohmyposh.dev/docs/configure#include-folders",
+              "default": [],
+              "items": {
+                "type": "string"
+              }
+            },
             "ignore_folders": {
```
lnu, line 197: Should we rename ignore_folders to exclud_folders. Seems more logical to me but introduces a kind of breaking change.

lnu, line 206: Should we rename ignore_folders to exclude_folders. Seems more logical to me but introduces a kind of breaking change. Or support both and marks the old one obsolete.

tradiff: That makes sense to me.  I updated the code.

src/properties.go
```diff
@@ -17,7 +17,11 @@ const (
 	Postfix Property = "postfix"
 	// ColorBackground color the background or foreground when a specific color is set
 	ColorBackground Property = "color_background"
-	// IgnoreFolders folders to ignore and not run the segment logic
+	// IncludeFolders folders to be included for the segment logic
+	IncludeFolders Property = "include_folders"
+	// ExcludeFolders folders to be excluded for the segment logic
+	ExcludeFolders Property = "exclude_folders"
+	// IgnoreFolders duplicate of ExcludeFolders
```
JanDeDobbeleer, line 24: that's for backwards compatibility, right? Can we mark this property as deprecated in the schema?

themes/schema.json
```diff
@@ -185,10 +185,28 @@
...
               "type": "array",
-              "title": "Ignore rendering in these folders",
-              "description": "https://ohmyposh.dev/docs/configure#ignore-folders",
+              "title": "Obsolete - please use `exclude_folders` instead",
```
JanDeDobbeleer: You can use the `deprecated ` keyword for this

src/segment.go
```diff
@@ -112,16 +112,42 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
...
-		for _, element := range list {
-			pattern := fmt.Sprintf("^%s$", element)
-			matched := matchString(pattern, cwd)
-			return matched
+		if len(list) > 0 {
+			included = cwdMatchesOneOf(cwd, list)
 		}
+	}
+
+	if !included {
```
JanDeDobbeleer: this is hard for me to reason about:

- if there are no values in both lists, it's true.
- if the folder is included, but not excluded it's true
- if the folder is not included and include folders has elements, it's false?
- if the folder is included and excluded, do we prioritise included?
- if the folder is excluded, it's false
- if the folder is not excluded and exclude folders has elements, it's true?

I would like to see this logic reflected in the tests as well, because now I feel we made some assumptions.

JanDeDobbeleer (changes requested):
We need to restructure the tests here, and ensure it works like we think it does :-)

src/segment.go
tradiff: I believe the test cases cover all of these examples.  Below I've quoted your scenarios and which test case I believe covers it. Let me know if there's anything I'm missing there.

```
if there are no values in both lists, it's true. → "Base Case"
if the folder is included, but not excluded it's true → "Included" and "Include Match / Exclude Mismatch"
if the folder is not included and include folders has elements, it's false? → "Include Mismatch"
if the folder is included and excluded, do we prioritise included? → "Include Match / Exclude Match"
if the folder is excluded, it's false → "Exclude"
if the folder is not excluded and exclude folders has elements, it's true? → "Exclude Mismatch"
```

I'll try to refactor the production code logic to make it more clear 👍.

themes/schema.json
tradiff: I actually tried that, but it had no noticeable effect in my editor (VS code).  I also tried upgrading the schema to `"$schema": "https://json-schema.org/draft/2020-12/schema"`.  I'll give it another shot.  Have we used the `deprecated` keyword previously in OMP?

src/segment.go
```diff
@@ -112,16 +112,42 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
...
 		return false
 	}
+
+	excludeValue, excludeOk := segment.Properties[ExcludeFolders]
```
JanDeDobbeleer: you can reuse value and ok like above: `if value, ok := segment.Properties[ExcludeFolders]; ok {`

src/segment.go
```diff
@@ -112,16 +112,42 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
...
-	if value, ok := segment.Properties[IgnoreFolders]; ok {
+func (segment *Segment) shouldIncludeFolder(cwd string) bool {
+	included := true
+	if value, ok := segment.Properties[IncludeFolders]; ok {
 		list := parseStringArray(value)
-		for _, element := range list {
-			pattern := fmt.Sprintf("^%s$", element)
-			matched := matchString(pattern, cwd)
-			return matched
+		if len(list) > 0 {
```
JanDeDobbeleer: I would check the `len` inside `cwdMatchesOneOf ` to reduce the need for this check downstream as well

src/segment.go
JanDeDobbeleer: If it does, perfect, I gave some pointers to make it a bit more readable which imho should be good to go when the tests pass. Thanks!

themes/schema.json
JanDeDobbeleer: no, not yet. I was looking for this for something else but never saw its effect in vscode (or anywhere else). We can add it and keep the title for context.

src/segment.go
tradiff: Please review the refactored code. I think that what's getting you tripped up is the different assumptions made if the IncludedFolders or ExcludedFolders are unspecified. I tried to think about the user's intention with each of these.

* If IncludedFolders is unspecified, everything is included
* If IncludedFolders is specified, the user wants the segment to be displayed only when CWD matches at least 1 included folder.
* If ExcludedFolders is unspecified, nothing is excluded
* If ExcludedFolders is specified, the user wants the segment to be displayed only when CWD matches 0 excluded folders.
* If both are specified, the user wants the segment to be displayed only when CWD matches at least 1 included folders and matches 0 ExcludedFolders 

Maybe a way to think about this logic is that it's both an Allow List and a Deny List.

I've left the changes unsquashed for now, in case you wanted to compare or anything.  If this looks good to you, I will squash them.

Edit: I've unresolved this thread, to ensure you notice my reply.

src/segment.go
```diff
@@ -112,15 +112,45 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
 	return defaultValue
 }
 
-func (segment *Segment) shouldIgnoreFolder(cwd string) bool {
-	if value, ok := segment.Properties[IgnoreFolders]; ok {
-		list := parseStringArray(value)
-		for _, element := range list {
-			pattern := fmt.Sprintf("^%s$", element)
-			matched := matchString(pattern, cwd)
-			return matched
```
tradiff, line 121: BTW I fixed a bug here.  `ignore_folders` was a list, but we only ever processed the first element of that list.

For example, in the below configuration, the segment would be hidden in C:\Projects, but visible in C:\temp

```json
"ignore_folders": [
  "C:\\\\Projects.*",
  "C:\\\\temp.*"
]
```

src/segment.go
tradiff: I had separated the map lookup from from the inline if statement, because of the ExcludeFolders and deprecated IgnoreFolders which will be used if ExcludeFolders is missing.  This logic is still present after my refactor, so let me know if you see a way to make this more elegant.

src/segment.go
tradiff: I don't think this is relevant anymore after my refactor.

themes/schema.json
tradiff: This is done

src/properties.go
tradiff: This seems to be a duplicate of your comment on schema.json.  Let me know if this is asking for something different.

JanDeDobbeleer: Duplicate 😃

src/segment.go
```diff
@@ -112,15 +112,45 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
...
+		return true
+	}
+
+	return cwdMatchesOneOf(cwd, list)
+}
+
+func (segment *Segment) cwdExcluded(cwd string) bool {
+	value, ok := segment.Properties[ExcludeFolders]
+	if !ok {
+		value = segment.Properties[IgnoreFolders]
```
JanDeDobbeleer, line 141: What happens when this isn't set?

src/segment.go
```diff
@@ -112,15 +112,45 @@ func (segment *Segment) getValue(property Property, defaultValue string) string
...
+func (segment *Segment) cwdIncluded(cwd string) bool {
+	value, ok := segment.Properties[IncludeFolders]
+	if !ok {
+		// IncludeFolders isn't specified, everything is included
+		return true
+	}
+
+	list := parseStringArray(value)
+
+	if len(list) == 0 {
```
JanDeDobbeleer, line 130: This check isn't found anymore in exclude. Can be part of `cwdMatchesOneOf`

JanDeDobbeleer (changes requested):

src/segment.go
tradiff: `list` is an emtpy array.  `cwd` isn't part of `list`, so `cwdMatchesOneOf` returns false.

src/segment.go
tradiff: You've suggested this twice now. Do you mean making `cwdMatchesOneOf` return two values? Or are you suggesting have `cwdMatchesOneOf` return true or false?  The reason this was never part of `cwdMatchesOneOf` is the behavior for an empty array is different for Include vs Exclude.

JanDeDobbeleer: The way I see it `cwdMatchesOneOf` works the same regardless of the context. Same goes for `cwdIncluded` and `cwdExcluded`, they are the same apart from the array they check (if it wasn't for `IgnoreFolders` this could be the same function. So the Len check had me confused, I would've expected it everywhere. Adding that to `cwdMatchesOneOf` would return false in case `cwdIncluded` meaning and empty array would not display the segment. I believe that's actually correct. No array means it returns true,  an empty array implies a mistake.

tradiff: > cwdMatchesOneOf works the same regardless of the context. 

Agreed

> Same goes for cwdIncluded and cwdExcluded, they are the same apart from the array they check 

Sorry, but you're mistaken here.  The difference lies in the code you're saying is not consistent: the implied value for an undefined property. 

cwdIncluded:  If the IncludeFolders property is not defined, everything is included (cwdIncluded returns true). 

cwdExcluded: If ExcludedFolders is not defined, nothing is excluded (cwdExcluded returns false).  I could make this more verbose in cwdExcluded, to explicitly return false, but it would not change the result (and would definitely not make these two functions be identical).

> No array means it returns true, an empty array implies a mistake.

I disagree here too. I think these two cases should be treated the same.  For cwdIncluded these two cases mean everything is included (return true).  For cwdExcluded these two cases means nothing is excluded (return false)

JanDeDobbeleer: Alright agreed, maybe it's more about how I would write this rather than the actual logic. And that's not something to argue over. Great work, this is very understandable and well argumented ;-)

JanDeDobbeleer (approved):
@TravisTX can you rebase? I'll gladly merge!

tradiff:
> @TravisTX can you rebase? I'll gladly merge!

Done.  You made me work for this one 😂

PR #498, opened 2021-03-08
feat(dotnet): Add Fsharp in dotnet segment known extentions

evilz:
Add support of Fsharp
`"*.fs", "*.fsx", "*.fsproj"`

Also add `*.csx`


JanDeDobbeleer:
@evilz perfect, all you need to change is the commit message so it follows the guidelines 😉 In this case: `feat(dotnet): support fsharp`

EDIT: the commit itself, not the PR title

Issue #505, opened 2021-03-10
Change dir with git is slow on linux


evilz:
### Description

Very slow on linux when git exists.
It take many seconds to just  change directory.

### Environment

- Oh my Posh version: 3.108.0
- Theme: jandedobbeleer.omp.json with some change (no on git)
- Operating System: Ubuntu 20.04.2 LTS on Windows 10 x86_64
- Shell:  bash 5.0.17
- Terminal: Windows terminal

### Steps to Reproduce

go to a dir with any git

JanDeDobbeleer:
That's the first time we have this on Linux. Can you set `enable_status` to false to see if that resolves it? Also, make sure you're on the latest version.

evilz:
I dont see `enable_status` in doc ? do you mean `display_status`  ?

JanDeDobbeleer:
@evilz oh damn yes, I'm sorry!

evilz:
Ok,

It's clear 

with `"display_status": true`  it's very slow
with `"display_status": false` it's instantaneous.

JanDeDobbeleer:
So somehow running git status is super slow for that repository. Can you time the output of that command in your shell?

evilz:
![image](https://user-images.githubusercontent.com/2937862/110660675-158bf900-81c4-11eb-95c3-1510bb0c023f.png)

evilz:
![image](https://user-images.githubusercontent.com/2937862/110660782-30f70400-81c4-11eb-9d06-0b699812343b.png)

evilz:
OK, it's a issue on WSL2 already known. https://github.com/microsoft/WSL/issues/4197
But maybe a **timeout** should be add for all segment ?

JanDeDobbeleer:
Good idea, although you'd want to control that though. I'll have a look.

evilz:
Ok I think I got a idea of fix I will do a PR :)

evilz:
Working perfectly now \o/ in v3.108.1

PR #509, opened 2021-03-10
fix: slow git status on wsl2

evilz:
Should fix this issue #505

use windows git.exe for mounted drive

### DIFF
--- a/src/segment_git.go
+++ b/src/segment_git.go
@@ -271,8 +271,11 @@ func (g *git) getStatusColor(defaultValue string) string {
 }
 
 func (g *git) getGitCommandOutput(args ...string) string {
+	inWSLSharedDrive := func(env environmentInfo) bool {
+              return env.isWsl() && strings.HasPrefix(env.getcwd(), "/mnt/")
+        }
 	gitCommand := "git"
-	if g.env.getRuntimeGOOS() == windowsPlatform || (g.env.isWsl() && strings.HasPrefix(g.env.getcwd(), "/mnt/")) {
+	if g.env.getRuntimeGOOS() == windowsPlatform || inWSLSharedDrive(g.env) {
 		gitCommand = "git.exe"
 	}
 	args = append([]string{"--no-optional-locks", "-c", "core.quotepath=false", "-c", "color.status=false"}, args...)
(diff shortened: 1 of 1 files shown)


src/segment_git.go
```diff
@@ -272,7 +272,7 @@ func (g *git) getStatusColor(defaultValue string) string {
 
 func (g *git) getGitCommandOutput(args ...string) string {
 	gitCommand := "git"
-	if g.env.getRuntimeGOOS() == windowsPlatform {
+	if g.env.getRuntimeGOOS() == windowsPlatform || (g.env.isWsl() && strings.HasPrefix(g.env.getcwd(), "/mnt/")) {
```
JanDeDobbeleer: Does it matter where it runs in this case?

evilz:
Yes, it's all about fileshare.
Default linux git command should be use outside Windows mount.

Le mer. 10 mars 2021 à 19:05, Jan De Dobbeleer <notifications@github.com> a
écrit :

src/segment_git.go
JanDeDobbeleer: ok, can we in that case add an inline function for clarity?

```golang
func (g *git) getGitCommandOutput(args ...string) string {
        inWSLSharedDrive := func(env environmentInfo) bool {
              return env.isWsl() && strings.HasPrefix(env.getcwd(), "/mnt/")
        }
	gitCommand := "git"
	if g.env.getRuntimeGOOS() == windowsPlatform || inWSLSharedDrive(g.env) {
		gitCommand = "git.exe"
	}
	args = append([]string{"--no-optional-locks", "-c", "core.quotepath=false", "-c", "color.status=false"}, args...)
	val, _ := g.env.runCommand(gitCommand, args...)
	return val
}
```

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@evilz can you make 1 commit out of this and format the file correctly with gofmt? Thx!

evilz:
I think `environmentInfo` should be more typed.

Like `getcwd` should return something else that a string where we can call inWSLSharedDrive on it.

Also `getRuntimeGOOS()` and `getPlatform()` can return enum or something more typed. 

That only my my two cents :P

PR #513, opened 2021-03-11
feat(themes): added new theme - slimfat - a fork of DamianoPellegrini's slim

irdkwmnsb:
I'd like to contribute to the themes variety of oh my posh.

I really liked @DamianoPellegrini's slim theme (#346) , but the path seemed too short. So this is a fork of their theme with the full path displayed.

Example: 
image

### DIFF
--- a/themes/slimfat.omp.json
+++ b/themes/slimfat.omp.json
@@ -1,0 +1,221 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "final_space": true,
+  "console_title": true,
+  "console_title_style": "template",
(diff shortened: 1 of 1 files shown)


irdkwmnsb:
After ~a little~ much googling and github issues browsing I found out that oh my posh can omit osc 9;9 codes. 

I don't know why it should be enabled manualy in themes, seems like it should be on by default.

JanDeDobbeleer:
@irdkwmnsb not all terminals support it. So it's waste at times 😉

irdkwmnsb:
Should I remove it then? Or leave it as is?

JanDeDobbeleer:
@irdkwmnsb it has the downside of having a bell sound in iTerm for example (it's not a standard, but first and foremost a notification). So removing it would indeed be the best solution for standard themes. Also, can you stick to one commit for this context?

PR #525, opened 2021-03-13
fix(git): StatusColors without DisplayStatus

tradiff:
### Prerequisites

- [ ] n/a Docs have been added / updated (for bug fixes / features)

### Description

Fixes a use-case that was broken in #502 .

Prior to #502, when using the combination of `"display_status": false` and `"status_colors_enabled": true`, 
the branch name would display in the configured color for the  current git status. This PR restores that functionality, and adds a test-case to cover it.

Example segment:

```json
{
  "type": "git",
  "style": "plain",
  "foreground": "#13A10E",
  "properties": {
    "display_status": false,
    "display_stash_count": false,
    "display_upstream_icon": false,
    "status_colors_enabled": true,
    "color_background": false,
    "local_changes_color": "#C19C00",
    "ahead_and_behind_color": "#C50F1F",
    "behind_color": "#C50F1F",
    "ahead_color": "#881798"
  }
},
```

### DIFF
--- a/src/segment_git.go
+++ b/src/segment_git.go
@@ -139,11 +139,18 @@ func (g *git) enabled() bool {
 }
 
 func (g *git) string() string {
+	statusColorsEnabled := g.props.getBool(StatusColorsEnabled, false)
 	displayStatus := g.props.getBool(DisplayStatus, true)
+
+	if displayStatus || statusColorsEnabled {
+		g.setGitStatus()
+	}
+	if statusColorsEnabled {
+		g.SetStatusColor()
+	}
 	if !displayStatus {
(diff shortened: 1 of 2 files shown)


JanDeDobbeleer:
That was actually a purposeful decision. No status also implies we don't fetch it. But I get there's a difference in expectations.

tradiff:
Yeah, but this property does use the status.  This broke my personal theme, where I only used the color to know the status. I don't want the status display string.

JanDeDobbeleer:
I'm reviewing later today, want to see this on my laptop 😅

JanDeDobbeleer (approved):

PR #535, opened 2021-03-16
feat: template for time segment

lnu:
### Prerequisites

### Description

Template added for time segment

### DIFF
--- a/docs/docs/segment-time.md
+++ b/docs/docs/segment-time.md
@@ -26,3 +26,12 @@ Show the current timestamp.
 - time_format: `string` - format to use, follows the [golang standard][format] - defaults to `15:04:05`
 
 [format]: https://gobyexample.com/time-formatting-parsing
+
+- template: `string` - A go [text/template][go-text-template] template extended with [sprig][sprig] utilizing the
+properties below. Only used when a value is set, making the above properties obsolete.
+
+  example: `{{ now | date \"January 02, 2006 15:04:05 PM\" | lower }}`
+
+## Template Properties
+
+- `.CurrentDate`: `time` - The time to display(testing purpose)
--- a/src/segment_time.go
+++ b/src/segment_time.go
@@ -1,12 +1,12 @@
 package main
 
-import (
-	"time"
-)
+import "time"
 
 type tempus struct {
-	props *properties
-	env   environmentInfo
+	props        *properties
+	env          environmentInfo
+	templateText string
+	CurrentDate  time.Time
 }
 
 const (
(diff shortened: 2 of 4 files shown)


src/segment_time.go
```diff
@@ -15,15 +15,36 @@ const (
...
-	return time.Now().Format(timeFormatProperty)
+	return t.getFormattedText()
 }
 
 func (t *tempus) init(props *properties, env environmentInfo) {
 	t.props = props
 	t.env = env
 }
+
+func (t *tempus) getFormattedText() string {
```
tradiff, line 43: hi @lnu , I was just playing around with this and found a way that we could reuse the existing property if we're interested in that.  I'm just throwing the idea out there in case we want to go this direction. Fully understand if it's too magic and we'd rather go with a new property that is explicitly a template + dto.

```go
func (t *tempus) string() string {
	timeFormatProperty := t.props.getString(TimeFormat, "15:04:05")
	timeStr := time.Now().Format(timeFormatProperty)
	template := &textTemplate{
		Template: timeStr,
	}
	return template.render()
}
```

with that code either one of these is valid:
```json
// prints the old behavior:
"time_format": "January 02, 2006 (MST)"

// prints in lower case:
"time_format": "{{\"January 02, 2006 (MST)\" | lower}}"
```

docs/docs/segment-time.md
```diff
@@ -26,3 +26,12 @@ Show the current timestamp.
 - time_format: `string` - format to use, follows the [golang standard][format] - defaults to `15:04:05`
 
 [format]: https://gobyexample.com/time-formatting-parsing
+
+- template: `string` - A go [text/template][go-text-template] template extended with [sprig][sprig] utilizing the
+properties below. Only used when a value is set, making the above properties obsolete.
+
+  example: `{{ now | date \"January 02, 2006 (MST)\" | lower }}`
```
tradiff: I feel like the example in the docs should include time, since time is what people are likely using this segment for.  It will save them some time of having to look at the golang docs if we have something like this:

```md
example: `{{ now | date \"January 02, 2006 (MST) 15:04:05 PM\" | lower }}`
```

src/segment_time.go
lnu: ~~it makes more sense for a time segment not to have a custom property. While doing the change, I thought we could even have a "generic" template segment because you can do all in one template for the time: 
` now |  date \"January 02, 2006 (MST)\" | lower`~~
To be consistent with the other segments, maybe it's better to keep a template and a time_format property?

JanDeDobbeleer: Indeed.

JanDeDobbeleer (approved):

PR #542, opened 2021-03-18
fix: preserve exit code in bash init script

erclu:
### Prerequisites

### Description

I had an issue with the exit code segment, in that it did not capture the previous command's exit code but was always 0.
I managed to narrow it down to the fact that i had another hook in the PROMPT_COMMAND, added by pyenv-virtualenv. Currently, they use this function to setup a hook on PROMPT_COMMAND:

```sh
export PATH="/home/erclu/.pyenv/plugins/pyenv-virtualenv/shims:${PATH}";
export PYENV_VIRTUALENV_INIT=1;
_pyenv_virtualenv_hook() {
  local ret=$?
  if [ -n "$VIRTUAL_ENV" ]; then
    eval "$(pyenv sh-activate --quiet || pyenv sh-deactivate --quiet || true)" || true
  else
    eval "$(pyenv sh-activate --quiet || true)" || true
  fi
  return $ret
};
if ! [[ "$PROMPT_COMMAND" =~ _pyenv_virtualenv_hook ]]; then
  PROMPT_COMMAND="_pyenv_virtualenv_hook;$PROMPT_COMMAND";
fi
```
I applied the same overall logic to your omp.bash script and I tested it locally by sourcing the modified file.

This way, the exit code is captured asap and propagated down the line to other hooks that users may have and may require having the exit code of the previous command instead of the exit code of the oh-my-posh invocation

Other changes include:

- renaming the defined functions to simplify future debugging. **Could this break anything else in oh-my-posh?** Sorry for asking but I don't know Go well enough to determine it by myself.
- quoting some variables and simplifying an expression as suggested by shellcheck.
- idempotency when modifying PROMPT_COMMAND. Without the idempotency check, running e.g. `source ~/.bashrc` added the omp prompt function multiple times, resulting in some stuff not working (e.g. executiontime segments). With this change, invoking `omp --init --shell bash` does not break anything.

The final change fixes a minuscule issue that was bothering me:

- the trap function now checks if the timer file exists before trying to remove it. Exiting a shell by using CTRL+D left an error message, no idea if this happened to others beside me.

P.S.: I discovered this project recently and I absolutely fell in love with it. Thanks for your work!

P.P.S.: I left a line in the commit message body that was too long, I amended it and force-pushed to my branch to fix it so I could pass the check ;)

### DIFF
--- a/src/init/omp.bash
+++ b/src/init/omp.bash
@@ -4,23 +4,27 @@ TIMER_START="/tmp/${USER}.start.$$"
 
 PS0='$(::OMP:: --millis > $TIMER_START)'
 
-function _update_ps1() {
+function _omp_hook() {
+    local ret=$?
+
     omp_elapsed=-1
     if [[ -f $TIMER_START ]]; then
         omp_now=$(::OMP:: --millis)
         omp_start_time=$(cat "$TIMER_START")
-        omp_elapsed=$(($omp_now-$omp_start_time))
-        rm $TIMER_START
+        omp_elapsed=$((omp_now-omp_start_time))
+        rm "$TIMER_START"
     fi
-    PS1="$(::OMP:: --config $POSH_THEME --error $? --execution-time $omp_elapsed --shell bash)"
+    PS1="$(::OMP:: --config $POSH_THEME --shell bash --error $ret --execution-time $omp_elapsed --shell bash)"
+
+    return $ret
 }
 
-if [ "$TERM" != "linux" ] && [ -x "$(command -v ::OMP::)" ]; then
-    PROMPT_COMMAND="_update_ps1; $PROMPT_COMMAND"
+if [ "$TERM" != "linux" ] && [ -x "$(command -v ::OMP::)" ] && ! [[ "$PROMPT_COMMAND" =~ "_omp_hook" ]]; then
+    PROMPT_COMMAND="_omp_hook; $PROMPT_COMMAND"
 fi
 
-function runonexit() {
-  rm $TIMER_START
+function _omp_runonexit() {
+  [[ -f $TIMER_START ]] && rm "$TIMER_START"
 }
 
-trap runonexit EXIT
+trap _omp_runonexit EXIT


src/init/omp.bash
```diff
@@ -4,23 +4,27 @@ TIMER_START="/tmp/${USER}.start.$$"
...
     if [[ -f $TIMER_START ]]; then
         omp_now=$(::OMP:: --millis)
         omp_start_time=$(cat "$TIMER_START")
-        omp_elapsed=$(($omp_now-$omp_start_time))
-        rm $TIMER_START
+        omp_elapsed=$((omp_now-omp_start_time))
+        rm "$TIMER_START"
     fi
-    PS1="$(::OMP:: --config $POSH_THEME --error $? --execution-time $omp_elapsed --shell bash)"
+    PS1="$(::OMP:: --config $POSH_THEME --error $ret --execution-time $omp_elapsed)"
```
JanDeDobbeleer: the `--shell bash` statement would need to remain as it causes issues running this on Apple silicon otherwise.

JanDeDobbeleer:
@erclu thanks a lot for this one! You can restructure the init scripts as you see fit as they don't interfere with oh-my-posh. The only required convention is ``::OMP::` as a placeholder to swap with the exact executable. Other than that they are full free form (apart from PowerShell that exports a few functions towards the module).

src/init/omp.bash
erclu: oh damn I lost that piece while copypasting the command between my shell and the file. I'll fix it asap

JanDeDobbeleer (changes requested):
Can you amend this into the previous commit? Makes no sense to keep it separate, thx!

erclu:
> @erclu thanks a lot for this one! You can restructure the init scripts as you see fit as they don't interfere with oh-my-posh. The only required convention is ``::OMP::` as a placeholder to swap with the exact executable. Other than that they are full free form (apart from PowerShell that exports a few functions towards the module).

Yeah I figured it was something like that, it's a really nice implementation by you to allow this kind of customization. 

I was going slightly insane because with the same exact configuration the exit code segment worked inside PowerShell Core but not bash

erclu:
> Can you amend this into the previous commit? Makes no sense to keep it separate, thx!

sure, gimme a sec

erclu:
damn you're too quick on merging :/ I added the `--shell bash` line twice here:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/8de004ff53e3ea6db6482042c25094622af66022/src/init/omp.bash#L17

PR #543, opened 2021-03-18
fix: remove double invocation of --shell argument

erclu:
Yeah well, I'm really sorry for this mishap :| It's kind of the first time I really use this fork & pull workflow. This PR fixes the error just introduced (by me) in #542


JanDeDobbeleer:
I missed that as well 😱

PR #548, opened 2021-03-19
fix: fish 3.2 empty command duration

lnu:
### Prerequisites

- [X] I have read and understand the `CONTRIBUTING` guide

### Description

[Description of the change]

### DIFF
--- a/src/init/omp.fish
+++ b/src/init/omp.fish
@@ -1,11 +1,22 @@
 function fish_prompt
     set -l omp_duration "$CMD_DURATION$cmd_duration"
-    if test "$omp_lastcommand" = ""
-        set omp_duration 0
+    # check if variable set, < 3.2 case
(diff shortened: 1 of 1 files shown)


src/init/omp.fish
```diff
@@ -1,11 +1,19 @@
 function fish_prompt
     set -l omp_duration "$CMD_DURATION$cmd_duration"
```
JanDeDobbeleer, line 2: That's milliseconds, right? So the chance of being off here and accidentally setting it to 0 by accident is minimal?

lnu: milliseconds right, but the lowest value I got is 4 with an empty command but you never know :)

lnu:
So, I tested this one on wsl with fish 3.1.0 and macOS/iterm with fish 3.2.1 and it seems to work in both.

PR #549, opened 2021-03-20
fix: escape double quotes for zsh prompt

lnu:
### Prerequisites

### Description

Only with zsh. Add the following block to your config(left prompt only, rprompt is ok):
```
{
          "type": "text",
          "style": "powerline",
          "powerline_symbol": "\uE0B0",
          "foreground": "#ffffff",
          "background": "#0077C2",
          "properties": {
            "text": " \"Hello world\" "
          }
}
```
it will break zsh with the following message:
image

It seems all double quotes must be escaped. BUT even when escaped, they are not rendered in the prompt:
image

But at least, it's rendered without any error.

I checked and the text is rendered with double quotes in all shell except zsh.... 

Any suggestions are welcome :).

### DIFF
--- a/src/engine.go
+++ b/src/engine.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
Let's get this fixed first. I did a quick google and couldn't find anything specific on this topic.

PR #550, opened 2021-03-20
feat: multiple config formats

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]


JanDeDobbeleer:
@lnu @TravisTX when you have some time available, I'd love some additional manual validation on this one :-) I tested everything and "works on my machine" 😓

src/settings.go
```diff
@@ -68,18 +74,56 @@ func loadUserConfiguration(env environmentInfo) (*Settings, error) {
...
 	return &settings, nil
 }
 
+func exportSettings(settingsFile, format string) string {
+	if len(format) == 0 {
+		format = config.JSON
+	}
+	err := config.LoadFiles(settingsFile)
+	if err != nil {
+		return "UNABLE TO OPEN CONFIG"
```
tradiff: I'm hitting this if I have invalid json, yaml, or toml.  What about an error more like "INVALID CONFIG" in that situation?  For example, if I use a trailing comma in the last element of a json block, or indent too much in yaml, or add a stray `]` in toml.  I also noticed `err` says exactly what the problem was and what line it's on, which could be helpful to the user.  Maybe just `fmt.Println` it?  Just an idea.

JanDeDobbeleer: I'll add the context indeed.

tradiff:
Everything seems to be working correctly. I can't figure out how to get the $schema to work in yaml using vscode.  My understanding is the json schema should also work with yaml.  I do have the `vscode-yaml` extension, which gives me coloring, but seems to not be giving me the context from the schema.  Have you had any luck with this?

lnu:
Will give a try tomorrow morning. Good work 👍

JanDeDobbeleer:
Vscode autocomplete is on my list. Didn't look at it yet.

lnu:
image

~json can be removed~ => "in the default output format(json)" or something

lnu:
> Everything seems to be working correctly. I can't figure out how to get the $schema to work in yaml using vscode. My understanding is the json schema should also work with yaml. I do have the `vscode-yaml` extension, which gives me coloring, but seems to not be giving me the context from the schema. Have you had any luck with this?

It works here when setting the schema url in user config. But it fails when set at the top of the yaml file(it should work):
`# yaml-language-server: $schema=<urlToTheSchema>`


=> works with the right syntax `# yaml-language-server: $schema=https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json`

We can add this extension `redhat.vscode-yaml` to extensions.json in .vscode(support for yaml autocomplete)

src/settings.go
JanDeDobbeleer: done

lnu: works fine 👍

lnu:
The only downside of yaml and toml is the lack of inline schema definition like the json one. ~Should work with yaml but somewhat fails and~ toml supports only schema uploaded to a central repo.
For the pull, looks good to me.

lnu:
even better toml extension supports toml schema but not inline. https://github.com/tamasfe/taplo/issues/95.

JanDeDobbeleer:
Ok. So, on my to-do list:
- Add inline schemas for supported languages if it's not present
- Create helper functions to avoid adding your current config (like pwsh does)

lnu:
Tested all your changes(export_poshconfig and export-poshtheme) and it works well 👍

JanDeDobbeleer:
@lnu maybe I should also align the name, no? `Export-PoshConfig` makes sense.

PR #587, opened 2021-03-27
feat(az): support az pwsh module

tradiff:
### Prerequisites

### Description

Change the az segment to support the Azure PowerShell Module

Closes #569

### DIFF
--- a/docs/docs/segment-environment.md
+++ b/docs/docs/segment-environment.md
@@ -29,7 +29,7 @@ make sure to do this after importing `go-my-posh` and you're good to go.
 function Set-EnvVar {
     $env:POSH=$(Get-Date)
 }
-New-Alias -Name 'Set-PoshContext' -Value 'Set-EnvVar' -Scope Global
+New-Alias -Name 'Set-PoshContext' -Value 'Set-EnvVar' -Scope Global -Force
 ```
 
 The segment will show when the value of the environment variable isn't empty.
--- a/src/init/omp.ps1
+++ b/src/init/omp.ps1
@@ -16,11 +16,22 @@ if (Test-Path $config) {
 
 function global:Set-PoshContext {}
 
-function global:Set-PoshGitStatus {
+function global:Initialize-ModuleSupport {
     if (Get-Module -Name "posh-git") {
         [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification = 'Variable used later(not in this scope)')]
         $global:GitStatus = Get-GitStatus
         $env:POSH_GIT_STATUS = Write-GitStatus -Status $global:GitStatus
+    }
+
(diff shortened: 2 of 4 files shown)


docs/docs/segment-environment.md
```diff
@@ -29,7 +29,7 @@ make sure to do this after importing `go-my-posh` and you're good to go.
 function Set-EnvVar {
     $env:POSH=$(Get-Date)
 }
-New-Alias -Name 'Set-PoshContext' -Value 'Set-EnvVar' -Scope Global
+New-Alias -Name 'Set-PoshContext' -Value 'Set-EnvVar' -Scope Global -Force
```
tradiff, line 32: `-Force` allows the user to run `.$profile` to reload their profile without an error message.

src/segment_az.go
```diff
@@ -36,24 +36,47 @@ func (a *az) init(props *properties, env environmentInfo) {
 }
 
 func (a *az) enabled() bool {
```
tradiff, line 38: I decided to make the EnvVars have priority over the CLI for two reasons:
1. The CLI is slow and this segment has always been slow on my computer.  So we do the faster operation first, and early exit if it was successful.
2. The EnvVars is something the user is opting in to by adding code to the $profile.  To me, the act of doing that signals that the user wants the EnvVars to be used.

docs/docs/segment-az.md
```diff
@@ -31,3 +31,27 @@ Display the currently active Azure subscription information.
...
+## Az PowerShell Module
+
+By default, the az segment will retrieve subscription information for the [Azure CLI][az-cli], not the [Az PowerShell Module][pwsh-module].
+If you use the Az PowerShell Module, add the code below to your PowerShell profile. This will cause
+the az segment to show information relevant to the Az PowerShell module.
+
+```pwsh
+function Set-EnvVar {
+    try {
+        $subscription = Get-AzContext | Select-Object -ExpandProperty "Subscription" | Select-Object "Name", "Id"
```
JanDeDobbeleer: we could preload this just like we do for `posh-git`. If that cmdlet is available, add those env vars.

tradiff: done!

src/init/omp.ps1
```diff
@@ -23,6 +23,19 @@ function global:Set-PoshGitStatus {
     }
 }
 
+function global:Set-AzSubscription {
```
JanDeDobbeleer: Looking at that we now also have `Set-PoshGitStatus` I would maybe rename that one to `Initialize-ModuleSupport` and add this logic there as well. That way we have 1 spot for those items and expand as we see fit.

tradiff: Sure, I can do that

tradiff: done!

JanDeDobbeleer:
@TravisTX if you rebase LGTM

PR #596, opened 2021-03-29
fix: Suppressing error message from chmod command

RobCannon:
### Prerequisites

- [Y] I have read and understand the `CONTRIBUTING` guide
- [Y] The commit message follows the [conventional commits][cc] guidelines
- [Y] Docs have been added / updated (for bug fixes / features)

### Description

Supressing the error message from the chod command when it is run from non-sudo session when the module was installed for all users.

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -65,6 +65,16 @@ A [Powershell][powershell] module is available for easy installation.
 Install-Module oh-my-posh -Scope CurrentUser -AllowPrerelease
 ```
 
+If you wish to install for all users in a linux environment, you will need to install from an elevate session
+of powershell.  You will also need to import the module one time from the elevated session in order to set the
+executable permissions correctly.
+
+```powershell
+sudo pwsh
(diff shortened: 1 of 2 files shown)


docs/docs/installation.mdx
```diff
@@ -65,6 +65,15 @@ A [Powershell][powershell] module is available for easy installation.
 Install-Module oh-my-posh -Scope CurrentUser -AllowPrerelease
 ```
 
+If you wish to install for all users in a linux environment, you will need to install from an elevate session
+of powershell.  You will also need to import the module one time from the elevated session in order to set the
+executable permissions correctly.
+```powershell
```
JanDeDobbeleer, line 72: this needs a space in between (and I need to add markdown validation)

JanDeDobbeleer (changes requested):
See inline comment ;-)

docs/docs/installation.mdx
RobCannon: I think you mean a blank line between the description and the powershell code snippet?  I added a blank line there.

PR #617, opened 2021-04-07
feat(theme): add stelbent

stefanes:
Add new theme, Stelbent, using only the base powerline characters from the private Unicode space. This theme do not require a Nerd Font.


JanDeDobbeleer:
@stefanes perfect, with the `minimal` convention and everything! To pass commitlint, your second paragraph needs to start with a lower case character. Can you adjust that? I'll gladly merge!

Issue #634, opened 2021-04-12
Location stack in path block


Aankhen:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

It would be very useful to have the path block show the size of the location stack if it’s available in the shell and contains more than one entry. In case it helps, the commands I know of to get the size are:

* `(Get-Location -Stack).Count` in PowerShell
* `dirs -p | wc - l` in Bash & zsh

### Environment

- Oh my Posh version: 3.114.0
- Theme: Custom
- Operating System: Windows 10
- Shell: PowerShell
- Terminal: Windows Terminal

### Steps to Reproduce

1. `pushd C:\dir1`
2. `pushd C:\dir2`
3. `pushd C:\dir3`

**Expected behavior:** The path block shows something like:

![Equivalent to 📂2 📁C:\dir3](https://user-images.githubusercontent.com/978992/114349186-352c8d80-9b85-11eb-988e-d68673e3a132.png)

**Actual behavior:** The path block shows only the folder (different one, sorry):

![~ shown](https://user-images.githubusercontent.com/978992/114349412-8046a080-9b85-11eb-829f-b1e679d0918a.png)

lnu:
something like this?
![image](https://user-images.githubusercontent.com/1829553/114353644-8b8dd200-9b6d-11eb-91fe-f2e6cdfbe5be.png)

But the behaviour is different between powershell and unix shells:

- powershell: works only if using pushd/popd (oob)
- zsh (default ohmyzsh config): each time you change directory, the path is added on the stack (and also with pushd/popd)
- fish: pushd/popd only by default (does not support dirs -p)
- bash: pushd/popd only by default

Aankhen:
Yes, like that! (Just with the count hidden if there’s nothing pushed.) May I ask how you have that? I went through `segment_path.go` twice before creating this issue, but I must have missed something, unless it’s a customization of your own.

Thank you for listing the behaviour of the different shells—I only have experience with Bash myself, and all I could find in my brief searches was what I read about zsh. I guess fish would be one of the shells that couldn’t show the stack. As far as zsh goes, I’d defer to actual users on whether it would be useful to show it. They could always hide it if not.

lnu:
I’ll push the branch later today

Aankhen:
Thanks, guys! I wish every feature request went so smoothly! 😬

[...]

PR #636, opened 2021-04-12
feat: show/hide stack count in path segment

lnu:
### Prerequisites

### Description

#634 
Adds the stack count in path segment when using pushd/popd

~~WIP. needs to make it work in Fish.~~

There is room for improvement(maybe with a template) but it works as requested for now.

### DIFF
--- a/src/init/omp.fish
+++ b/src/init/omp.fish
@@ -1,6 +1,7 @@
 set -g posh_theme ::CONFIG::
 
 function fish_prompt
+    set -l stack_count (count $dirstack)
     set -l omp_duration "$CMD_DURATION$cmd_duration"
     # check if variable set, < 3.2 case
     if set -q omp_lastcommand; and test "$omp_lastcommand" = ""
@@ -14,7 +15,7 @@ function fish_prompt
       set -gx omp_last_status_generation $status_generation
     end
 
-    ::OMP:: --config $posh_theme --error $status --execution-time $omp_duration
+    ::OMP:: --config $posh_theme --error $status --execution-time $omp_duration --stack-count $stack_count
 end
 
 function postexec_omp --on-event fish_postexec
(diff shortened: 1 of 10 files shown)


src/init/omp.fish
```diff
@@ -1,6 +1,7 @@
 set -g posh_theme ::CONFIG::
 
 function fish_prompt
+    set stack_count 0 #(dirs -p | wc -l)
```
JanDeDobbeleer: why not local?

src/main.go
```diff
@@ -136,6 +137,10 @@ func main() {
 			"rprompt-offset",
 			40,
 			"Offset the right prompt with x when using --export-img"),
+		StackCount: flag.Int(
+			"stack-count",
+			0,
+			"The current stack count"),
```
JanDeDobbeleer: I would reword this to `The current location stack count`

src/segment_path.go
```diff
@@ -41,6 +41,8 @@ const (
 	MappedLocations Property = "mapped_locations"
 	// MappedLocationsEnabled enables overriding certain locations with an icon
 	MappedLocationsEnabled Property = "mapped_locations_enabled"
+	// StackCountVisible enables the stack count display
+	StackCountVisible Property = "stack_count_visible"
```
JanDeDobbeleer: In line with other properties I would rename this with `stack_count_enabled`

src/init/omp.fish
lnu: I was quickly testing and I forgot to add `- l`. It should count correctly on all supported shells now.

src/segment_path.go
lnu: Yep, i put enabled at first. Will set it back.

lnu:
still need to add some tests

PR #657, opened 2021-04-17
feat(node): show NPM/Yarn context

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/src/segment_node.go
+++ b/src/segment_node.go
@@ -1,11 +1,24 @@
 package main
 
+import "fmt"
+
 type node struct {
-	language *language
+	language           *language
+	packageManagerIcon string
 }
 
+const (
+	// YarnIcon illustrates Yarn is used
+	YarnIcon Property = "yarn_icon"
+	// NPMIcon illustrates NPM is used
+	NPMIcon Property = "npm_icon"
+	// DisplayPackageManager shows if NPM or Yarn is used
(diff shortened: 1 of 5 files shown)


src/segment_node.go
```diff
@@ -22,13 +35,26 @@ func (n *node) init(props *properties, env environmentInfo) {
...
+func (n *node) loadContext() {
+	if !n.language.props.getBool(DisplayPackageManager, false) {
+		return
+	}
+	if n.language.env.hasFiles("yarn.lock") {
+		n.packageManagerIcon = n.language.props.getString(YarnIcon, " \uF61A")
+	}
+	if n.language.env.hasFiles("package-lock.json") || n.language.env.hasFiles("package.json") {
+		n.packageManagerIcon = n.language.props.getString(NPMIcon, " \uE71E")
+	}
```
xt0rted, line 55: @JanDeDobbeleer should this be an `if/else if`? In my yarn based projects I'm seeing the npm icon and I think it's because they contain a `yarn.lock` and a `package.json` which the second `if` is picking up.

JanDeDobbeleer: You're right. There was a return in here but that got removed somehow 🤦🏻‍♂️

JanDeDobbeleer: Fixed.

PR #672, opened 2021-04-20
Fix some default values in git segment documentation

benallred:
I noticed some differences between the docs and actual behavior, so double-checked the implementations and updated the docs.


JanDeDobbeleer:
@benallred can you adjust the commit message to be in line with our conventional commit guidelines?

PR #674, opened 2021-04-21
fix: disable battery segment when no batteries

lnu:
### Prerequisites

I tested the latest version on a desktop without batteries(opensuse tumbleweed, zsh latest, omp latest) and it doesn't return any error but an empty batteries list. I did a quick fix for it but I'm not sure it covers all scenarios.

### DIFF
--- a/src/segment_battery.go
+++ b/src/segment_battery.go
@@ -34,7 +34,13 @@ const (
 
 func (b *batt) enabled() bool {
 	batteries, err := b.env.getBatteryInfo()
+
 	if !b.enabledWhileError(err) {
+		return false
+	}
+
+	// case on computer without batteries(no error, empty array)
+	if err == nil && len(batteries) == 0 {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


src/segment_battery.go
```diff
@@ -34,7 +34,7 @@ const (
 
 func (b *batt) enabled() bool {
 	batteries, err := b.env.getBatteryInfo()
-	if !b.enabledWhileError(err) {
+	if !b.enabledWhileError(err) || len(batteries) == 0 {
```
JanDeDobbeleer: I did mis this when refactoring. Should be good (although the tests probably don't cover the use-case)

JanDeDobbeleer: @lnu can you add a tests that ensures this is validated as well?

JanDeDobbeleer (changes requested):

JanDeDobbeleer (approved):

JanDeDobbeleer:
@lnu can you rebase?

lnu:
Will be my first job tomorrow 🙂

Issue #683, opened 2021-04-23
.Env template property for all color templates or .OS template property


gitolicious:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

I would like to colorize my OS segment depending on which OS I am on (e.g. Windows: blue, Ubuntu: orange, Debian: red, Suse: green). This could be implemented
* in a generic way which would help every segment: by being able to access `.Env` in the `background_templates` and find fitting environment variables for each OS, or
* very specific: by offering a `.OS` template property in the OS segment.

### Environment

- Oh my Posh version: _3.142.0_
- Theme: _custom_
- Operating System: _Win10 / Ubuntu 20.04 on WSL_
- Shell: _PowerShell Core / zsh_
- Terminal: _Windows Terminal_

### Steps to Reproduce

1. Configure `background_templates` for OS segment

```jsonc
{
  "blocks": [
    {
      "type": "prompt",
      "alignment": "left",
      "segments": [
        {
          "type": "os",
          "style": "plain",
          "foreground": "#ffffff",
          "background_templates": [
            "{{if contains \"Windows_NT\" .Env.OS}}#0077c2{{end}}",
            "{{if contains \"Ubuntu\" .Env.WSL_DISTRO_NAME}}#E95420{{end}}",
            "#333333" // fallback gray seen below
          ]
        }
      ]
    }
  ]
}
```

**Expected behavior:**
![image](https://user-images.githubusercontent.com/26963495/115924215-b803f100-a47f-11eb-8d46-986f7b37917d.png)

**Actual behavior:**
![image](https://user-images.githubusercontent.com/26963495/115924233-bdf9d200-a47f-11eb-8b6c-d7c0ee441994.png)

PR #684, opened 2021-04-23
feat: add OS template property

gitolicious:
### Prerequisites

### Description

Your code looked really clean and easy to follow so I gave #683 a shot. The following example results in a nice [Debian rubine red](https://wiki.debian.org/DebianLogo#The_red_color) OS logo:

```jsonc
        {
          "type": "os",
          "style": "plain",
          "foreground": "#ffffff",
          "background_templates": [
            "{{if contains \"debian\" .OS}}#CE0056{{end}}",
            "#333333"
          ]
        },
```


gitolicious:
Potential extensions I could think of are
* a `.WSL` template property to define different colors for WSL and native consoles
* a `"color_background": true` configuration property which automatically sets pre-defined background colors for each OS

src/segment_os.go
```diff
@@ -76,6 +77,7 @@ func (n *osInfo) string() string {
 	case "linux":
 		wsl := n.env.getenv("WSL_DISTRO_NAME")
 		p := n.env.getPlatform()
+		n.OS = p
```
JanDeDobbeleer: I would opt to set this regardless of the platform:

```golang
func (n *osInfo) string() string {
	goos := n.env.getRuntimeGOOS()
	switch goos {
	case windowsPlatform:
		n.OS = windowsPlatform
		return n.props.getString(Windows, "\uE62A")
	case darwinPlatform:
		n.OS = darwinPlatform
		return n.props.getString(MacOS, "\uF179")
	case linuxPlatform:
		wsl := n.env.getenv("WSL_DISTRO_NAME")
		p := n.env.getPlatform()
		if len(wsl) == 0 {
			n.OS = p
			return n.getDistroName(p, "")
		}
		n.OS = wsl
		return fmt.Sprintf("%s%s%s",
			n.props.getString(WSL, "WSL"),
			n.props.getString(WSLSeparator, " - "),
			n.getDistroName(p, wsl))
	default:
		n.OS = goos
		return goos
	}
}
```

This also add the two platforms to `environment.go`:

```golang
	darwinPlatform  = "darwin"
	linuxPlatform   = "linux"
```

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
> a "color_background": true configuration property which automatically sets pre-defined background colors for each OS

@lnu that's solvable by using templates imho. Our go-to should be that, if the template can handle it, we don't extend the base properties. Ideally we can move 100% to templates for every segment one day.

src/segment_os.go
gitolicious: Good point, I amended the PR accordingly.

gitolicious:
> @lnu that's solvable by using templates imho. Our go-to should be that, if the template can handle it, we don't extend the base properties. Ideally we can move 100% to templates for every segment one day.

A predefined  setting would save the users looking up the appropriate colors for the 20+ distros, though finding nice defaults would be quite some effort even for the repo maintainers.

JanDeDobbeleer:
@gitolicious true, I'd like to meet the user who has them all set up though...

JanDeDobbeleer (approved):

gitolicious:
Thx for the quick merge and kudos for your automation and toolchain setup on Github! Never saw <5 mins from merging a PR to having the new version installed on my system via Scoop. 🏎

lnu:
@JanDeDobbeleer @gitolicious It's a trade-off between simplicity/ease of use/predefined settings and more advanced/customizable settings with templates. The first way means more work/test and magic, the second one if well documented(with samples in the doc) can be very powerful.
I'd like both ways but as most of omp users are already "power users", the second strategy can be more efficient on the long term. Both approaches makes sense, now it's up to the "lead" to make a decision :grin:

gitolicious:
I looked up some colors codes for the more popular choices.

`macos`: `#000000` / `#a3aaae`
image
`windows`: `#ffffff` / `#1675c7` (PowerShell logo colors)
image
`alpine`: `#ffffff` / `#00567e`
image
`debian`: `#ffffff` / `#ce0056`
image
`raspbian`: `#ffffff` / `#c51a4a`
image
`fedora`: `#ffffff` / `#294172`
image
`gentoo`: `#ffffff` / `#54487a`
image
`opensuse`: `#ffffff` / `#73ba25`
image
`ubuntu`: `#ffffff` / `#e95420`
image

PR #697, opened 2021-05-03
feat: hide_separator option for powerline; new theme: powerlevel10k_rainbow

uruz-7:
### Prerequisites

### Description

I wanted an option that could hide the separator between a powerline segment and others, so I could make a powerlevel10k rainbow style based on the fascinating jandedobbeleer theme.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -195,6 +195,63 @@ Text character to use when `"style": "powerline"`.
 If `true` this swaps the foreground and background colors. Can be useful when the character you want does not exist
 in the perfectly mirrored variant for example.
 
+### Hide Separator
+
+By default a separator between a powerline segment and a diamond/plain segment will be rendered. If `true` this removes
+the separator. Can be useful when you want a powerline that is right next to other segments.
+
+Notice: If you have segments that is NOT always enabled, be sure to set EVERY of them till the first always-enabled segment.
+
+```jsonc
+{
+  "segments": [
+    {
+      "type": "session",
+      "style": "diamond",
+      "foreground": "#ffffff",
+      "background": "#ffb300",
+      "leading_diamond": "\uE0B6",
+      // in this case a trailing_diamond is not required
+    },
+    {
+      "type": "go",
+      "style": "powerline",
+      "powerline_symbol": "\uE0B0",
+      "foreground": "#ffffff",
+      "background": "#61AFEF",
+      "hide_separator": true,
+      // hide_separator is required here
+      "properties": {
+        "always_enabled": false,
+      }
+    },
+    {
+      "type": "path",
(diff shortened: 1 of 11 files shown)


JanDeDobbeleer:
@uruz-7 can you give me a screenshot the result? Maybe we can go about this another way (or not).

uruz-7:
> @uruz-7 can you give me a screenshot the result? Maybe we can go about this another way (or not).

The result is like this.

image

What issues is that by default, a separator will be rendered between diamond and powerline 1(or 4), which I personally didn't want.

I have tried to change powerline 1 and 2 to diamond with customized color leading and trailing diamonds, but the `git` segment will change color, so the color of the separator between 1 and 2 must be adaptive => powerline seems to be a better choice.

image

Anyway, I'll appreciate it if there is a better way,.

JanDeDobbeleer:
It looks so pretty. Let me check if we can fix this without an additional property 👍🏻

JanDeDobbeleer:
@uruz-7 if I do the following change I get the same result by leaving either trailing or leading diamond empty:

```bash
diff --git a/src/block.go b/src/block.go
index d6f2f5c7..d6e444a0 100644
--- a/src/block.go
+++ b/src/block.go
@@ -89,9 +89,10 @@ func (b *Block) renderSegments() string {
 }

 func (b *Block) endPowerline() {
-   if b.activeSegment != nil &&
-       b.activeSegment.Style != Powerline &&
-       b.previousActiveSegment != nil &&
+   if b.previousActiveSegment == nil || b.activeSegment == nil {
+       return
+   }
+   if b.activeSegment.Style != Powerline &&
        b.previousActiveSegment.Style == Powerline {
        b.writePowerLineSeparator(b.getPowerlineColor(false), b.previousActiveSegment.background(), true)
    }
@@ -113,6 +114,12 @@ func (b *Block) getPowerlineColor(foreground bool) string {
    if b.previousActiveSegment == nil {
        return Transparent
    }
+   if b.previousActiveSegment.Style == Diamond && len(b.previousActiveSegment.TrailingDiamond) == 0 {
+       return b.previousActiveSegment.background()
+   }
+   if b.activeSegment.Style == Diamond && len(b.activeSegment.LeadingDiamond) == 0 {
+       return b.activeSegment.background()
+   }
    if !foreground && b.activeSegment.Style != Powerline {
        return Transparent
    }
```

uruz-7:
Wonderful job! Let me check it!

uruz-7:
@JanDeDobbeleer This worked perfect for my case. Thanks!

And it influenced some other themes which could be fixed easily in https://github.com/JanDeDobbeleer/oh-my-posh/pull/700.

This PR could be closed :)

(Please just ignore my mistake in branching above.)

PR #741, opened 2021-05-25
fix: escape \ from segment value when in bash

### LINKED ISSUE #738: Escaped characters in path are replaced with whatever they translate to, but only in git bash, windows-terminal
### Prerequisites

### Description

The clickable path in the terminal (for example `D:\dev`) works fine on a powershell core instance, but it breaks in git bash:

image
image

This path, for example, `D:\dev` shows normally on powershell, but the "\d" part of it becomes "Mon May 24", the current day.
It also does not happen when git bash is used inside vscode, only in windows-terminal. Happens with some other themes as well, but `pixelrobots` is the one I use.

### Environment

- Oh my Posh version:  3.156.0
- Theme: pixelrobots
- Operating System: Windows 10 64-bit 20H2 Build 19042.985
- Shell: Git Bash (git version 2.31.1.windows.1)
- Terminal: Windows Terminal 1.7.210419001-release1.7

### Steps to Reproduce

1. Install oh-my-posh with scoop 
2. Configure git bash by adding `eval "$(oh-my-posh --init --shell bash --config 'D:\\scoop\\apps\\oh-my-posh\\current\\themes\\pixelrobots.omp.json')"` to ~/.bashrc (D:\scoop is my scoop home)
3. Navigate to D:\dev

**Expected behavior:** Clickable path in terminal should work "unescaped" (?), like it works with powershell core

**Actual behavior:** Clickable path in terminal replaces some escaped characters (such as "\d" with current day), producing incorrect file paths

lnu:
### Prerequisites

### Description

First draft of escaping \ for bash. I don't like it since the check is done inside block.go, which should be shell agnostic.
fix #738 

### Tips

If you're not comfortable working with git, we're working a [guide][docs] to help you out.
Oh my Posh advises [GitKraken][kraken] as your preferred cross platfrom git GUI powertool.

### DIFF
--- a/src/block.go
+++ b/src/block.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 )
@@ -90,7 +91,13 @@ func (b *Block) renderSegments() string {
 		}
 		b.activeSegment = segment
 		b.endPowerline()
-		b.renderSegmentText(segment.stringValue)
+		segmentValue := segment.stringValue
+		// escape backslashes to avoid replacements
+		// https://tldp.org/HOWTO/Bash-Prompt-HOWTO/bash-prompt-escape-sequences.html
+		if b.env.getShellName() == bash {
+			segmentValue = strings.ReplaceAll(segment.stringValue, "\\", "\\\\")
+		}
+		b.renderSegmentText(segmentValue)
 	}
 	if b.previousActiveSegment != nil && b.previousActiveSegment.Style == Powerline {
 		b.writePowerLineSeparator(Transparent, b.previousActiveSegment.background(), true)


JanDeDobbeleer:
I was waiting on time to provide an answer, but isn't the most straightforward fix here to use `folder_separator` and map to `/`. This will only ever happen in git bash (which is to be avoided anyways).

lnu:
It's not only broken for the path segment, if there's a '\d' in any segment, it is expanded as the current date(you can see it by adding a text segment with '\d').

JanDeDobbeleer:
I was trying this, but that gives the same error. Would've expected this to work.

```diff
diff --git a/src/engine.go b/src/engine.go
index 49313711..41b820ed 100644
--- a/src/engine.go
+++ b/src/engine.go
@@ -25,8 +25,7 @@ func (e *engine) string() string {
 	return e.console.String()
 }
 
-func (e *engine) canWriteRPrompt() bool {
-	prompt := e.string()
+func (e *engine) canWriteRPrompt(prompt string) bool {
 	consoleWidth, err := e.env.getTerminalWidth()
 	if err != nil || consoleWidth == 0 {
 		return true
@@ -156,17 +155,24 @@ func (e *engine) debug() string {
 }
 
 func (e *engine) print() string {
+	prompt := e.string()
 	switch e.env.getShellName() {
 	case zsh:
 		if !*e.env.getArgs().Eval {
 			break
 		}
 		// escape double quotes contained in the prompt
-		prompt := fmt.Sprintf("PS1=\"%s\"", strings.ReplaceAll(e.string(), "\"", "\"\""))
-		prompt += fmt.Sprintf("\nRPROMPT=\"%s\"", e.rprompt)
-		return prompt
-	case pwsh, powershell5, bash, plain:
-		if e.rprompt == "" || !e.canWriteRPrompt() {
+		eval := fmt.Sprintf("PS1=\"%s\"", strings.ReplaceAll(prompt, "\"", "\"\""))
+		eval += fmt.Sprintf("\nRPROMPT=\"%s\"", e.rprompt)
+		return eval
+	case bash:
+		// escape backslashes to avoid replacements
+		// https://tldp.org/HOWTO/Bash-Prompt-HOWTO/bash-prompt-escape-sequences.html
+		prompt = strings.ReplaceAll(prompt, "\\", "\\\\")
+		e.rprompt = strings.ReplaceAll(e.rprompt, "\\", "\\\\")
+		fallthrough
+	case pwsh, powershell5, plain:
+		if e.rprompt == "" || !e.canWriteRPrompt(prompt) {
 			break
 		}
 		e.write(e.ansi.saveCursorPosition)
@@ -175,5 +181,5 @@ func (e *engine) print() string {
 		e.write(e.rprompt)
 		e.write(e.ansi.restoreCursorPosition)
 	}
-	return e.string()
+	return prompt
 }
diff --git a/src/engine_test.go b/src/engine_test.go
index 8b39d62c..f836bdfd 100644
--- a/src/engine_test.go
+++ b/src/engine_test.go
@@ -37,7 +37,7 @@ func TestCanWriteRPrompt(t *testing.T) {
 		}
 		engine.rprompt = strings.Repeat("x", tc.RPromptLength)
 		engine.console.WriteString(strings.Repeat("x", tc.PromptLength))
-		got := engine.canWriteRPrompt()
+		got := engine.canWriteRPrompt(engine.print())
 		assert.Equal(t, tc.Expected, got)
 	}
 }
```

lnu:
In that case you end up escaping thing that should not be escaped(I also tried it this morning):
image

src/block.go
```diff
@@ -90,7 +91,12 @@ func (b *Block) renderSegments() string {
 		}
 		b.activeSegment = segment
 		b.endPowerline()
-		b.renderSegmentText(segment.stringValue)
+		segmentValue := segment.stringValue
+		// it hurts
+		if b.env.getShellName() == "bash" {
```
JanDeDobbeleer: change to the inner variable `bash` rather than the string value

JanDeDobbeleer:
@lnu also just found out about that. Now, to be clear, the cleanest way would be to be able to turn off that functionality. If not, check my pointer and let's go with the ugly way. You're right, you need the clean string value. But add the comment I added for future reference ;-)

lnu:
> Now, to be clear, the cleanest way would be to be able to turn off that functionality

You mean disabling bash support :)

JanDeDobbeleer:
Can you rebase?

PR #776, opened 2021-06-03
Updated documentation of Path segment to reflect JSON schema

Jazzinghen:
### Prerequisites

### Description

Updated the documentation for Path Segments as it was still reporting that it accepted a map while the connected schema requires a list of 2-elements lists.

I tried to infer how to represent that as a type in documentation from other places in the documentation but I couldn't find a relevant example.

### Tips

If you're not comfortable working with git, we're working a [guide][docs] to help you out.
Oh my Posh advises [GitKraken][kraken] as your preferred cross platfrom git GUI powertool.

### DIFF
--- a/docs/docs/segment-path.md
+++ b/docs/docs/segment-path.md
@@ -47,15 +47,15 @@ locations before doing a replacement.
 
 - mapped_locations_enabled: `boolean` - replace known locations in the path with the replacements before applying the
 style. defaults to `true`
-- mapped_locations: `map[string]string` - custom glyph/text for specific paths (only when `mapped_locations_enabled`
+- mapped_locations: `[][]string, string` - custom glyph/text for specific paths (only when `mapped_locations_enabled`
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@Jazzinghen it's the schema that's outdated. I adjusted this a while back because that syntax is a bit ugly but forgot to update the schema it seems. The old syntax still works for backwards compatibility, but shouldn't be supported anymore from a docs/schema point of view.

Jazzinghen:
> 
> 
> @Jazzinghen it's the schema that's outdated. I adjusted this a while back because that syntax is a bit ugly but forgot to update the schema it seems. The old syntax still works for backwards compatibility, but shouldn't be supported anymore from a docs/schema point of view.

Oh! Haha.
I wrote it because VSCode was  complaining about the new JSON config I was writing and I checked it out. The dictionary makes sense, actually :D

PR #810, opened 2021-06-17
Remove confusing quote

boehs:
### Prerequisites

### Description

The quote here has no context and makes zero sense. I suggest removing it.


JanDeDobbeleer:
That was the point indeed. Not everything in life makes sense and sometimes you simply get annoyed being serious all the time. That said, go watch Final Space as that's how it's linked 😉

boehs:
I am not opposed to humor when it makes sense but I spent a long while trying to understand the context of the quote or why it was there, and concluded it made zero sense in the context, let alone in documentation. Assuming the reader has watched "Final Space" is bold. 

That being said, I understand if you want to keep it and I thank you for the recommendation, looks quite good and will watch after I finish rewatching _chuck_ 😉

JanDeDobbeleer:
I dare to make bold moves at times. Which in the end makes us having this conversation 😉

PR #863, opened 2021-07-19
docs: Added docs for installing on Windows/WSL specifically

jamesshenry:
### Prerequisites

### Description

Current documentation does not correctly inform users how to configure on WSL. Copied the relevant .mdx files and added  WSL specific instructions.

I edited the "Override the theme settings" section but those instructions won't work on PowerShell if the user installed via winget or scoop

### DIFF
--- a/docs/docs/install-customize-wsl.mdx
+++ b/docs/docs/install-customize-wsl.mdx
@@ -1,0 +1,84 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+<Tabs
+  defaultValue="powershell"
+  groupId="shell"
+  values={[
+    { label: 'powershell', value: 'powershell', },
+    { label: 'zsh', value: 'zsh', },
+    { label: 'bash', value: 'bash', },
+    { label: 'fish', value: 'fish', },
+  ]
+}>
+<TabItem value="powershell">
+
+```powershell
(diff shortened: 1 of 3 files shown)


JanDeDobbeleer:
Why would we create a custom WSL section rather than have people follow the unix guide?

jamesshenry:
Because the documentation is incorrect? Perhaps I should have created an issue first

JanDeDobbeleer:
> Because the documentation is incorrect? Perhaps I should have created an issue first

Because the path of the theme is wrong? Windows home vs WSL home? That seems to be the core issue unless I've missed something else.

I would rather provide that context on top than duplicate everything (this is putting quite the burden on maintenance).

JanDeDobbeleer:
@henry-js if you give me some more info I'll happily figure out how to facilitate this.

jamesshenry:
Yeah that was the essence of the issue. I was coming at it from the point of view of someone who has not much experience with Linux, but dabbling into WSL. 
If you're following the WIndows doc, everything works until you attempt to enable oh-my-posh in a WSL linux shell.
Noticed the 'Customize' section now tells you the full path to share themes:
image

So maybe something like that could be added earlier also.
e.g.
> ***Info***
> Remember, in WSL the executable is `oh-my-posh-wsl`

JanDeDobbeleer:
@All-contributors please add @henry-js for doc

PR #878, opened 2021-08-02
perf/git segment

### LINKED ISSUE #873: Performance: Command Execution Slow in Git Repositories
### Prerequisites

- [x] I have looked for duplicate issues before submitting this one

### Description
Hello. The program runs very smothly(almost instantaneously) when outside a git repository, but as soon as I enter a git repository(ir doesn't matter how big or small it is) it gets slow taking more than 300ms to execute on average wich is noticible. An example can be seen in the picture below:
slow2

I have followed the steps described in the offitial page for the installation. I've been browsing the internet trying to look for a solution or verifying if this was normal or not but couldn't find an answer. Also, on the youtube videos I visited seemed didn't seem to be any kind of slowdowns when getting in a git repository.

I want to know if this is normal and if it is not I'd like to know how I could try to fix it. Thanks in advance!

lnu:
### Prerequisites

### Description

git segment perf improvement: read origin url from config instead of calling git exe

resolves #873 

### Tips

If you're not comfortable with working with Git, we're working a [guide][docs] to help you out.
Oh My Posh advises [GitKraken][kraken] as your preferred cross platform Git GUI power tool.


src/segment_git.go
```diff
@@ -128,15 +132,20 @@ func (g *git) enabled() bool {
...
 	// handle worktree
+	g.repo.gitRootFolder = gitdir.path
 	dirPointer := g.env.getFileContent(gitdir.path)
 	dirPointer = strings.Trim(dirPointer, " \r\n")
 	matches := findNamedRegexMatch(`^gitdir: (?P<dir>.*)$`, dirPointer)
 	if matches != nil && matches["dir"] != "" {
-		g.repo.gitFolder = matches["dir"]
+		g.repo.gitWorkingFolder = matches["dir"]
+		ind := strings.Index(g.repo.gitWorkingFolder, "/.git")
+		g.repo.gitRootFolder = g.repo.gitWorkingFolder[:ind+5]
```
JanDeDobbeleer, line 150: this is magic :-)

lnu: Comments added. How would you change it?
It could break if you have a git repo in a path under .git/worktrees like `d:\.git\worktrees\myrepo`, but in that case you're looking for troubles :)

JanDeDobbeleer: I would make it explicit and trim `/worktrees/branch` in that case rather than work with indices.

lnu: we don't know yet the branch name at this stage. Would be better to have a perfect match but here if the +5 is explained, it should not hurt.

PR #892, opened 2021-08-13
fix: avoid variable expansion when using postfix in zsh

lnu:
### Prerequisites

### Description

If a `%` is used in `postfix` property, zsh try to expand the variable.  
like here(`%~` set in `postfix` property of battery segment):
image


JanDeDobbeleer:
I'm currently on holiday so I'll be reviewing this a bit slower than usual.

JanDeDobbeleer:
LGTM will merge

PR #897, opened 2021-08-13
feat: media segment

zuozishi:
### Prerequisites

### Description

Show the currently playing media info from Windows NowPlayingSession API or media-related process window title.

Groove:
groove

Spotify:
spotify

YouTube (Edge Chromium):
youtube


JanDeDobbeleer:
I'm currently on holiday so I'll be reviewing this a bit slower than usual.

src/segment_media_others.go
```diff
@@ -0,0 +1,7 @@
+// +build !windows
+
+package main
+
+func (s *media) enabled() bool {
```
JanDeDobbeleer: Rather than doing this I would for now only enable the segment on Windows by validating the platform we're on. If we ever have a fully working set on another platform, we can always extend it like this.

src/segment_media_test.go
```diff
@@ -0,0 +1,58 @@
...
+func TestStringPlayingMedia(t *testing.T) {
+	expected := "\ue602 Candlemass - Spellbreaker"
+	s := &media{}
+	s.info.MediaInfo.Title = "Candlemass"
+	s.info.MediaInfo.Artist = "Spellbreaker"
+	s.info.Playback.PlaybackState = 5
+	assert.Equal(t, expected, s.string())
+}
+
+func TestStringPausedMedia(t *testing.T) {
```
JanDeDobbeleer: We moved to a specific style of tests. See [here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/572149b176b864775edae0e531ed38153a12d66a/src/segment_battery_test.go#L17) for more details.

src/segment_media_windows.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func (s *media) enabled() bool {
+	var cmd = exec.Command("sys-media-info")
+	var str, err = cmd.Output()
+	if err == nil {
+		json.Unmarshal(str, &s.info)
+		return true
+	}
+	s.other, err = s.env.getWindowTitle("cloudmusic.exe", "^(.*\\s-\\s.*)$")
```
JanDeDobbeleer: This can be refactored to a for loop using the players and be easily extended that way too. Right now this is basically 3 times the same code.

src/segment_media.go
```diff
@@ -0,0 +1,159 @@
...
+		Subtitle            string   `json:"Subtitle"`
+		Title               string   `json:"Title"`
+		Artist              string   `json:"Artist"`
+		MediaClassPrimaryID string   `json:"MediaClassPrimaryID"`
+		Genres              []string `json:"Genres"`
+		AlbumTrackCount     int      `json:"AlbumTrackCount"`
+		TrackNumber         int      `json:"TrackNumber"`
+	} `json:"MediaInfo"`
+	Timeline struct {
+		StartTime struct {
```
JanDeDobbeleer: This and the structs below are all the same, define it above and initialize these as such.

JanDeDobbeleer (changes requested):
We would also need a section in the docs to explain the segment to end users.

src/segment_media.go
```diff
@@ -0,0 +1,134 @@
...
+	// MediaPausedIcon indicates a song is paused
+	MediaPausedIcon Property = "paused_icon"
+	// MediaStoppedIcon indicates a song is stopped
+	MediaStoppedIcon Property = "stopped_icon"
+	// MediaTrackSeparator is put between the artist and the track
+	MediaTrackSeparator Property = "track_separator"
+	// MediaTimeSeparator is put between the media position and total time
+	MediaTimeSeparator Property = "time_separator"
+	// MediaShowTime is media time show or hidden switch
+	MediaIsShowTime Property = "is_show_time"
```
JanDeDobbeleer, line 22: Our other `displaying` switches are prepended with `display_` so I would rename this to `display_time`.

src/segment_media.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type CShapTimeSpan struct {
```
JanDeDobbeleer, line 25: Typo (CSharp), but I would keep `TimeSpan` only.

src/segment_media.go
```diff
@@ -0,0 +1,134 @@
...
+	props *properties
+	env   environmentInfo
+	info  NowPlayingSessionInfo
+	other string
+}
+
+func (s *media) enabled() bool {
+	tool := "sys-media-info"
+	if s.env.isWsl() {
+		tool += ".exe"
```
JanDeDobbeleer, line 87: Why not use `.exe` always?

src/segment_media.go
```diff
@@ -0,0 +1,134 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func (s *media) string() string {
+	separator := s.props.getString(TrackSeparator, " - ")
+	time_separator := s.props.getString(MediaTimeSeparator, "/")
+	if s.other != "" {
```
JanDeDobbeleer, line 110: We always use `len(s.other) != 0`

JanDeDobbeleer, line 110: We always use `len(s.other) != 0`

src/segment_media.go
```diff
@@ -0,0 +1,134 @@
...
+		tool += ".exe"
+	}
+	if s.env.hasCommand(tool) {
+		str, err := s.env.runCommand(tool, "--json")
+		if err == nil && str != "{}" {
+			json.Unmarshal([]byte(str), &s.info)
+			return true
+		}
+	}
+	players := [...]string{"cloudmusic.exe", "qqmusic.exe", "spotify.exe"}
```
JanDeDobbeleer, line 96: I wonder what the difference is between this and the Spotify implementation on Windows? As it seems to fetch more, which is really interesting tbh.

JanDeDobbeleer (changes requested):
Very brief wording on my end due to limited connectivity. A few changes, and I do wonder if I need to structurally address the Spotify implementation on Windows.

Overall, would it make sense to rename this segment to `windows media`? As that's what it implies.

JanDeDobbeleer:
@zuozishi any plans to continue on this one? I'd love to proceed with it.

lnu:
I spent a few minutes to test the pull and update the code: https://github.com/lnu/oh-my-posh/tree/feat/media_segment

it works well. I also removed the check for "others" player. As @JanDeDobbeleer said, all the information and even more are given. The best would be to avoid the exe dependency and being able to call the dll from go but that's another (big) step.

lnu:
checked and it really slows down the prompt too much(for me):
image
image

JanDeDobbeleer:
@lnu that's also just a binary doing some win32 calls. If we decide to use this, we should call those directly.

lnu:
yes and no... it's a bit complicated. The api is a winrt api https://github.com/ADeltaX/NPSMLib

JanDeDobbeleer:
> it's a bit complicated

Saw that. Given the responsiveness I'm closing this one.

PR #915, opened 2021-08-25
feat(git): display worktree count

lnu:
### Prerequisites

Display worktree count in git segment.  
There was also an issue with stash count not being displayed when using worktree.  
worktree and stash are global object and not stored in the worktree folder but in the root folder.

[Description of the change]

### DIFF
--- a/src/segment_git.go
+++ b/src/segment_git.go
@@ -20,6 +20,7 @@ type gitRepo struct {
 	gitWorkingFolder string // .git working folder, can be different of root if using worktree
 	isWorkTree       bool
 	gitRootFolder    string // .git root folder
+	worktreeCount    int
 }
 
 type gitStatus struct {
@@ -120,6 +121,10 @@ const (
 	AheadColor Property = "ahead_color"
 	// BranchMaxLength truncates the length of the branch name
 	BranchMaxLength Property = "branch_max_length"
+	// DisplayWorktreeCount show worktree count or not
+	DisplayWorktreeCount Property = "display_worktree_count"
+	// WorktreeCountIcon shows before the worktree context
+	WorktreeCountIcon Property = "worktree_count_icon"
 )
 
 func (g *git) enabled() bool {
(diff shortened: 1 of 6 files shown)


lnu:
I tried to set `fb44` by default but it does not look nice. I set a dumb tree instead 'f1bb'

src/segment_git.go
```diff
@@ -486,3 +511,15 @@ func (g *git) getOriginURL(upstream string) string {
 	}
 	return keyVal
 }
+
+// getGitPath returns the right path depending on being in a worktree or not
+func (g *git) getGitPath(file string) string {
+	path := g.repo.gitWorkingFolder
+	if file == "config" || file == "logs/refs/stash" || file == "worktrees" {
```
JanDeDobbeleer: This feels a bit off to me. I see what you're trying to do but it feels strange. Is there another way to facilitate this somehow by specifying `root_repo` as a boolean for example?

lnu: I don't like this code.... I did not want to update all calls to `getGitFileContents` or `hasGitFile` etc...
I'll think about a cleaner way to do it

src/segment_git.go
```diff
@@ -389,25 +400,35 @@ func (g *git) hasGitFile(file string) bool {
 	return g.env.hasFilesInDir(g.repo.gitWorkingFolder, file)
 }
 
-func (g *git) hasGitFolder(folder string) bool {
-	path := g.repo.gitWorkingFolder + "/" + folder
-	return g.env.hasFolder(path)
+func (g *git) hasGitFolder(folder string, rootFolder bool) bool {
```
JanDeDobbeleer: Still feels off. I see what you meant with avoiding to add that context to every call. This way we're better off replacing the call with the actual `env` call to avoid that boolean and simply fall back to the relevant call and context in line. Thoughts?

lnu: I tried a move into that direction. Direct access to `env` for `hasGitFolder` and folder parameter added to `getGitFileContents`.  
Tell me what you think.

JanDeDobbeleer (approved):

PR #943, opened 2021-09-10
feat: add format option to owm segment

Karel-Kroeze:
adds a format property to the owm segment, allowing users to change the
format used to generate the weather label (e.g. remove the parentheses,
display weather with less than absolute accuracy).

### DIFF
--- a/docs/docs/segment-owm.md
+++ b/docs/docs/segment-owm.md
@@ -29,6 +29,7 @@ The free tier for *Current weather and forecasts collection* is sufficient.
     "location": "AMSTERDAM,NL",
     "units": "metric",
     "enable_hyperlink" : false,
+    "format": "%s (%g%s)",
     "http_timeout": 20
   }
 }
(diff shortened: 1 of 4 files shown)


Karel-Kroeze:
please note that I have no idea what I'm doing when it comes to Go, and this is the first time I've done any amount of coding in it. I've copy/pasted my way through adding tests, and things seem to run fine - but it might be worth taking some extra time to check for stupid mistakes 😇

JanDeDobbeleer:
@Karel-Kroeze I respect that. Will do a review over the weekend 👍🏻

JanDeDobbeleer:
@Karel-Kroeze thank you for your patience. I have a structural request, which will require you to rewrite it quite a bit :-) We do not make use of the standard formatting options, but rather `text/templating` functionality. You can look at the [AWS](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segment_aws.go#L82) segment to see how that works, that way everything is aligned and there's even more freedom.

JanDeDobbeleer (changes requested):
@Karel-Kroeze thank you for your patience. I have a structural request, which will require you to rewrite it quite a bit :-) We do not make use of the standard formatting options, but rather `text/templating` functionality. You can look at the [AWS](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segment_aws.go#L82) segment to see how that works, that way everything is aligned and there's even more freedom.

lnu:
https://github.com/JanDeDobbeleer/oh-my-posh/pull/1161

PR #946, opened 2021-09-11
feat: weather caching

boarder2:
### Prerequisites

- [ ] Tests for the changes have been added (for bug fixes/features) (Not sure what the best way to test this would be?)

### Description

Adds caching of weather data locally to keep the prompt more responsive with higher timeouts.

### DIFF
--- a/docs/docs/segment-owm.md
+++ b/docs/docs/segment-owm.md
@@ -30,6 +30,7 @@ The free tier for *Current weather and forecasts collection* is sufficient.
     "units": "metric",
     "enable_hyperlink" : false,
     "http_timeout": 20
+    "cachefile": ""
   }
 }
 ```
@@ -44,3 +45,7 @@ The free tier for *Current weather and forecasts collection* is sufficient.
                     Available values are standard (kelvin), metric (celsius), and imperial (fahrenheit) - defaults to `standard`
 - enable_hyperlink: `bool` - Displays an hyperlink to get openweathermap data
 - http_timeout: `int` - The default timeout for http request is 20ms.
+- cachefile: `string` - If specified, location where weather data will be cached. 
+                        Data will be cached for 10 minutes. 
+                        This will allow you to set a higher timeout and keep your prompt responsive more often. 
+                        Read/write access to the file is required.
--- a/docs/docs/segment-owm.md
+++ b/docs/docs/segment-owm.md
@@ -29,7 +29,7 @@ The free tier for *Current weather and forecasts collection* is sufficient.
     "location": "AMSTERDAM,NL",
     "units": "metric",
     "enable_hyperlink" : false,
-    "http_timeout": 20
+    "http_timeout": 20,
     "cachefile": ""
   }
 }
--- a/docs/docs/segment-owm.md
+++ b/docs/docs/segment-owm.md
@@ -45,7 +45,7 @@ The free tier for *Current weather and forecasts collection* is sufficient.
                     Available values are standard (kelvin), metric (celsius), and imperial (fahrenheit) - defaults to `standard`
 - enable_hyperlink: `bool` - Displays an hyperlink to get openweathermap data
 - http_timeout: `int` - The default timeout for http request is 20ms.
-- cachefile: `string` - If specified, location where weather data will be cached. 
-                        Data will be cached for 10 minutes. 
-                        This will allow you to set a higher timeout and keep your prompt responsive more often. 
+- cachefile: `string` - If specified, location where weather data will be cached.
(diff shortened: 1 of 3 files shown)


src/segment_owm.go
```diff
@@ -61,22 +65,58 @@ func (d *owm) string() string {
...
+func (d *owm) getResult() (*OWMDataResponse, error) {
 	apikey := d.props.getString(APIKEY, ".")
 	location := d.props.getString(LOCATION, "De Bilt,NL")
 	units := d.props.getString(UNITS, "standard")
 	timeout := d.props.getInt(HTTPTimeout, DefaultHTTPTimeout)
+	cachefile := d.props.getString(CACHEFILE, "")
+
+	if cachefile != "" {
+		stats, err := os.Stat(cachefile)
+		if err == nil {
```
JanDeDobbeleer, line 77: We'll need to remove the arrow of `if`'s here. Try to return early and only keep 1 level of indentation.

src/segment_owm.go
```diff
@@ -61,22 +65,58 @@ func (d *owm) string() string {
...
+				}
+			}
+		}
+	}
 
 	url := fmt.Sprintf("http://api.openweathermap.org/data/2.5/weather?q=%s&units=%s&appid=%s", location, units, apikey)
 	body, err := d.env.doGet(url, timeout)
 	if err != nil {
-		return err
+		return new(OWMDataResponse), err
```
JanDeDobbeleer, line 94: if there's an error, you can `return nil, err`

src/segment_owm.go
```diff
@@ -61,22 +65,58 @@ func (d *owm) string() string {
...
 	}
 	q := new(OWMDataResponse)
 	err = json.Unmarshal(body, &q)
+	if err != nil {
+		return new(OWMDataResponse), err
```
JanDeDobbeleer, line 99: if there's an error, you can `return nil, err`

src/segment_owm.go
```diff
@@ -61,22 +65,58 @@ func (d *owm) string() string {
...
+	}
+
+	if cachefile != "" {
+		err := os.WriteFile(cachefile, body, 0600)
+		if err != nil {
```
JanDeDobbeleer, line 104: When there's an error with the cache, I don't see why we do not return the data.
I would make use of `defer` to write the cache after the entire logic executes.

docs/docs/segment-owm.md
```diff
@@ -29,7 +29,8 @@ The free tier for *Current weather and forecasts collection* is sufficient.
     "location": "AMSTERDAM,NL",
     "units": "metric",
     "enable_hyperlink" : false,
-    "http_timeout": 20
+    "http_timeout": 20,
+    "cachefile": ""
```
JanDeDobbeleer, line 33: Why don't we make use of for example writing an environment variable to store this context?

JanDeDobbeleer (changes requested):

docs/docs/segment-owm.md
boarder2: Using one or two environment variables instead of a file is a good idea! I'll try to get back to it this weekend. Thanks for the suggestions!

lnu: the only upside of the cache file is when you use multiple console at the same time. But, you could aslo have multiple console with different config with different locations overriding the same cache file... :zany_face:

JanDeDobbeleer: @lnu in this use-case I feel the cache can be the same cross shell for this one.

boarder2: So I dug into this a little bit last night and if you set an environment variable from within the application, the value is not persisted outside the run of the application.

At least not when using os.Setenv() - I don't think this will work since the OMP process isn't long lived, it's just run long enough to get the prompt text.

Any other thoughts on how to do this without requiring the user to specify a file location?

lnu: You're right, you can only set a variable for a process and its children... so it's not possible to store it in an env variable. 
You could store it by default in the temp dir of the user(`os.TempDir`)

JanDeDobbeleer: I like tempdir. We would need to create a cache mechanism rather than implement it solely here though. Ideally key-value based.

lnu: good idea for a global persisted cache, in the meantime I did some changes based on the branch here(https://github.com/boarder2/oh-my-posh3/pull/1)

JanDeDobbeleer: #982 has a an implementation idea for a persisted cache

PR #963, opened 2021-09-17
feat: allow for custom text in place of shell names

asherber:
### Prerequisites

### Description

Give the `shell` segment the ability to use custom text in place of the shell name -- e.g., display `PS` instead of `pwsh`.

### DIFF
--- a/docs/docs/segment-shell.md
+++ b/docs/docs/segment-shell.md
@@ -18,7 +18,7 @@ Show the current shell name (ZSH, powershell, bash, ...).
   "foreground": "#ffffff",
   "background": "#0077c2",
   "properties": {
-    "custom_text": {
+    "mapped_shell_names": {
       "pwsh": "PS"
     }
   }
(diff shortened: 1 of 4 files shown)


docs/docs/segment-shell.md
```diff
@@ -18,7 +18,13 @@ Show the current shell name (ZSH, powershell, bash, ...).
   "foreground": "#ffffff",
   "background": "#0077c2",
   "properties": {
-    "prefix": " \uFCB5 "
+    "custom_text": {
```
JanDeDobbeleer: I would rename this to `mapped_shell_names` in line with `mapped_locations`. Other than that, really good work!!!

JanDeDobbeleer (changes requested):

docs/docs/segment-shell.md
asherber: Thanks for the quick response (and the great utility)! Since you're particular about the state of your repo, do you prefer an amended commit or a new one? And if new, what's the appropriate prefix for the commit message?

JanDeDobbeleer: You can create a new one as it will always be a single commit in the end (I can squash this one). Prefix would be `feat` in this case.

JanDeDobbeleer (approved):

PR #971, opened 2021-09-17
feat: improved theme png rendering

NickCraver:
### Prerequisites

- [x] Tests for the changes have been added (for bug fixes/features) (in PR description)

### Description

This is making the following changes to the PNG render pipeline for themes:
1. Use the latest Hack Nerd Font here (from https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/Hack), this solves many to all missing glyph issues.
2. Checks for Nerd Font ranges (defined at https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points) and assumes double with for those.

Here are some example before (based on current website) and with this change:

| Theme | Before | After |
| ----- | ------ | ----- |
| M365Princess | image | m365princess |
| agnoster | image | agnoster |
| blue-owl | image | blue-owl |
| bubbles | image | bubbles |
| craver | image | craver |
| darkblood | image | darkblood |
| material | image | material |
| nu4a | image | nu4a |
| paradox | image | paradox |
| pixelrobots | image | pixelrobots |
| powerlevel10k_classic | image | powerlevel10k_classic |
| powerlevel10k_rainbow | image | powerlevel10k_rainbow |
| slim | image | slim |
| wopian | image | wopian |

Some things I don't love:
- There may be a wildly simpler way to implement this
- I may have committed atrocities against golang and need to wear the cone of shame
- We're explicitly checking Nerd Font Unicode ranges ([which are documented](https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points)), and further excluding a few from that (known to be single width, that I ran into testing)
  - Counter-argument: this is for previews and may rarely or not ever have a negative consequence

...is this a bad idea? I'm not sure - I hope it helps or sparks another idea at least. My feeling is the _result_ is a net win (unless people hate the font!), but the implementation is icky. If the gut reaction to this is "omg Nick, wtf?!?" I don't mind it getting closed one bit, only trying to explore a _possible_ path here to make the themes a bit more accurate in the previews.


src/image.go
```diff
@@ -320,9 +372,11 @@ func (ir *ImageRenderer) SavePNG(path string) error {
 		}
 
 		w, h := dc.MeasureString(str)
+		w += (w * float64(ir.runeAdditionalWidth(runes[0])))
+
 		if ir.backgroundColor != nil {
 			dc.SetRGB255(ir.backgroundColor.r, ir.backgroundColor.g, ir.backgroundColor.b)
-			dc.DrawRectangle(x, y-h, w, h+12)
+			dc.DrawRectangle(x, y-h+3, w, h+9)
```
JanDeDobbeleer, line 389: These magic numbers, I'm equally guilty to it, can we give them a name?

JanDeDobbeleer (approved):

src/image.go
JanDeDobbeleer: As it's not "new", we'll keep it as is.

JanDeDobbeleer:
@NickCraver this isn't a bad take. To be honest, I was so happy to get this to work in the first place, so this part of the codebase can be pragmatic. Will there be a cleaner way? Probably. Is there a lot of ROI when spending the time on it? Highly unlikely.

src/image.go
```diff
@@ -320,9 +372,11 @@ func (ir *ImageRenderer) SavePNG(path string) error {
 		}
 
 		w, h := dc.MeasureString(str)
+		w += (w * float64(ir.runeAdditionalWidth(runes[0])))
```
JanDeDobbeleer, line 382: this is done to double the width or not, right?

NickCraver: Indeed! I tried to go through and add comments on all points. It was light on comments before so wasn't sure on stance there, but hopefully the additions here are good - happy to tweak more!

PR #972, opened 2021-09-18
feat(az): use template for more flexibility

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/docs/docs/segment-az.mdx
+++ b/docs/docs/segment-az.mdx
@@ -33,7 +33,23 @@ To enable this, set `$env:AZ_ENABLED = $true` in your `$PROFILE`.
 
 ## Properties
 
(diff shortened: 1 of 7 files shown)


JanDeDobbeleer:
@lnu this could also be considered "breaking", but it's easily adjusted. In hindsight, even the change in behaviour for the PR you did wasn't breaking in that sense. Thoughts?

lnu:
Not breaking in the sense of it won’t work anymore but still you have to update your template. But at the same time keeping backward compatibility for things you can easily update in 2 minutes is not really worth it. Don’t you think?

JanDeDobbeleer:
I decided the numbers are there for a reason and if we at one point have things like this that create a different behavior for an existing config, we treat that as breaking. It's just that we have 2 short after one another, but in the end it's only a number.

PR #976, opened 2021-09-18
feat(path): flexible matching for mapped-locations

asherber:
### Prerequisites

### Description

I'm not sure how you'll feel about this one. The goal was to make `mapped_locations` more portable. 

My use case is that I've got two machines with similar directory structures, except that I'm `Aaron` on one and `ASherber` on the other. I'd like to be able to put `"~\\Documents\\Projects"` in `mapped_locations` and have it work on both machines (i.e., match both `C:\Users\Aaron\Documents\Projects` and `C:\Users\ASherber\Documents\Projects`). 

While I'm at it, because I hate having to escape backslashes, and because Windows doesn't care which path separator I use, I'd like to be able to use `"~/Documents/Projects"`. And then if I make `mapped_locations` case-insensitive on Windows and Mac, I can actually use `"~/documents/projects"` and use the same theme file on all OSes.

If you agree with this approach, I've got similar changes queued up for `include_folders` and `exclude_folders` which I could either add to this PR or put into a new one.

### DIFF
--- a/docs/docs/segment-path.md
+++ b/docs/docs/segment-path.md
@@ -43,12 +43,12 @@ Display the current path.
 ## Mapped Locations
 
 Allows you to override a location with an icon. It validates if the current path **starts with** the value and replaces
-it with the icon if there's a match. To avoid issues with nested overrides, Oh my posh will sort the list of mapped
+it with the icon if there's a match. To avoid issues with nested overrides, oh-my-posh will sort the list of mapped
 locations before doing a replacement.
 
 - mapped_locations_enabled: `boolean` - replace known locations in the path with the replacements before applying the
-style. defaults to `true`
-- mapped_locations: `map[string]string` - custom glyph/text for specific paths. Works regardless of the `mapped_locations_enabled`
+style - defaults to `true`
+- mapped_locations: `object` - custom glyph/text for specific paths. Works regardless of the `mapped_locations_enabled`
 setting.
 
 For example, to swap out `C:\Users\Leet\GitHub` with a GitHub icon, you can do the following:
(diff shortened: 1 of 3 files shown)


docs/docs/segment-path.md
```diff
@@ -43,12 +43,12 @@ Display the current path.
 ## Mapped Locations
 
 Allows you to override a location with an icon. It validates if the current path **starts with** the value and replaces
-it with the icon if there's a match. To avoid issues with nested overrides, Oh my posh will sort the list of mapped
+it with the icon if there's a match. To avoid issues with nested overrides, oh-my-posh will sort the list of mapped
```
JanDeDobbeleer: The docs should always refer to Oh My Posh, to stay consistent.

docs/docs/segment-path.md
```diff
@@ -43,12 +43,12 @@ Display the current path.
...
 locations before doing a replacement.
 
 - mapped_locations_enabled: `boolean` - replace known locations in the path with the replacements before applying the
-style. defaults to `true`
-- mapped_locations: `map[string]string` - custom glyph/text for specific paths. Works regardless of the `mapped_locations_enabled`
+style - defaults to `true`
+- mapped_locations: `object` - custom glyph/text for specific paths. Works regardless of the `mapped_locations_enabled`
```
JanDeDobbeleer, line 51: Why did you change this to object?

src/segment_path.go
```diff
@@ -220,6 +220,19 @@ func (pt *path) getPwd() string {
 	return pwd
 }
 
+func (pt *path) normalize(inputPath string) string {
```
JanDeDobbeleer, line 223: this would need to be put outside of the scope of path when you want to reuse the functionality.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
Agreed with doing this on macOS and Windows given their nature. Actually like this, amazing flexibility and a real world use-case. Also well written case on why and how. Great work overall.

docs/docs/segment-path.md
asherber: Okay -- I thought I'd seen it a few ways in the current docs, and the current "Oh my posh" didn't seem right. I'll change this.

docs/docs/segment-path.md
asherber: Because it's an object in JSON, which is what we care about here, and a map in Go. This specification confused me when I first saw it. `mapped_shell_names` has the same structure, and it's specified as `object`.

src/segment_path.go
asherber: Yes, and possibly also refactored a bit.

docs/docs/segment-path.md
JanDeDobbeleer: If we now align I do believe oh-my-posh makes more sense, but that's beyond the scope of this PR IMHO.

src/segment_path.go
JanDeDobbeleer: I suppose we leave that refactoring for the following PR?

PR #979, opened 2021-09-21
feat: flexible matching for include/exclude folders

asherber:
### Prerequisites

### Description

Give include/exclude folders the same powers as mapped_locations (#976) -- can use slash instead of backslash and ~ instead of home dir, and case-insensitive on Windows and macOS.

I know in the previous PR we talked about centralizing and refactoring the "normalize" functionality. I found this tricky -- and possibly not desirable -- for a couple of reasons. First, the needs of normalizing are somewhat different in `Segment` and `path`, making it a little tricky to formulate a graceful function that they can share. And second, moving normalization into `environment`, which seemed like the most appropriate place, means that any tests that use `MockedEnvironment` and do something with the path now need to mock all the possible calls to `normalizePath()` by replicating most of its logic.

I've been working on that, and can continue down that path if you like, but in the meantime I thought I'd show you what this looks like without refactoring, with similar code in `Segment` and `path`, to see how much it bothers you. 

I also noticed that `path.replaceMappedLocations()` gets rid of `Microsoft.PowerShell.Core\FileSystem::` before doing any matching, but `Segment.cwdMatchesOneOf()` does not. I wasn't sure if this was intentional; I can add this to `cwdMatchesOneOf()` if you want.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -293,8 +293,9 @@ will not be rendered when in one of the excluded locations.
 ]
 ```
 
-You can also specify a [regular expression][regex] to create folder wildcards.
-In the sample below, any folders inside the `/Users/posh/Projects` path will be matched.
+The strings specified in these properties are evaluated as [regular expressions][regex]. You
+can use any valid regular expression construct, but the regular expression must match the entire directory
+name. The following will match `/Users/posh/Projects/Foo` but not `/home/Users/posh/Projects/Foo`.
 
 ```json
 "include_folders": [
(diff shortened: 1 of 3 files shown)


JanDeDobbeleer (approved):

asherber:
Just to clarify: Do you want me to continue refactoring this code, or are you okay with the current state? And do you want me to add the replacement of `Microsoft.PowerShell.Core\FileSystem::` to `Segment.cwdMatchesOneOf()`?

JanDeDobbeleer:
@asherber you can always refactor it if you're not happy with it, I didn't take that much time on reviewing this one as I liked your previous work. We can do the second implementation suggestion indeed.

asherber:
Thanks @JanDeDobbeleer! I'll keep looking at the refactoring, see if I can get to something I like. In the meantime, I'll also send a PR for the second thing.

PR #982, opened 2021-09-21
feature cache

JanDeDobbeleer:
- refactor: split concurrent map from cache implementation
- feat: cache capabilities

### Prerequisites

### Description

[Description of the change]


lnu:
I did something similar on my side today… 
With load/save from/to a file

src/environment_cache.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (fc *fileCache) init(home string) {
+	fc.cache = newConcurrentMap()
+	fc.home = home
+	content, err := ioutil.ReadFile(home + cachePath)
+	if err != nil {
+		return
+	}
+	for _, line := range strings.Split(string(content), "\n") {
```
lnu: I did serialize using  gob.NewEncoder/gob.NewDecoder but yours can be easily opened with a text editor. The only missing part is an expiration timeout.

src/environment_cache.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (fc *fileCache) close() {
+	if dump, err := json.Marshal(fc.cache.list()); err == nil {
+		_ = ioutil.WriteFile(fc.home+cachePath, dump, 0644)
+	}
+}
+
+func (fc *fileCache) get(key string) (*cacheObject, bool) {
+	val, found := fc.cache.get(key)
```
lnu, line 49: Maybe checking the timestamp before returning the value from the cache

src/environment_cache.go
JanDeDobbeleer: Added, but I'll refactor so you set the TTL in seconds when adding. A lot easier for consumption/handling.

src/environment_cache.go
JanDeDobbeleer: Yes. That's the idea now that I wrote this. That should be internal and not leak.

JanDeDobbeleer:
> I did something similar on my side today…
> With load/save from/to a file

@lnu I'd appreciate a review, if architecturally OK I can start adding some tests

src/environment.go
```diff
@@ -77,24 +85,24 @@ type environmentInfo interface {
...
 func (c *commandCache) get(command string) (string, bool) {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-	if cmd, ok := c.commands[command]; ok {
-		command = cmd
+	cmd, found := c.commands.get(command)
+	if !found {
+		return "", false
+	}
+	if command, ok := cmd.(string); ok {
```
JanDeDobbeleer: remove this and simply use `return cmd.(string)`

src/environment_cache.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	for key, co := range list {
+		fc.cache.set(key, co)
+	}
+}
+
+func (fc *fileCache) close() {
+	fc.set("hello", "world", 200)
+	if dump, err := json.Marshal(fc.cache.list()); err == nil {
```
JanDeDobbeleer: format the output for human readability

lnu: Is it possible to format it for human readbility with the addition of the timestamp?

lnu: On my side it was serialized and thus impossible to read for a "human". Maybe we can add an extra parameter to omp to clear or display the content of the cache?

src/environment.go
lnu: I did the same, returning the value with a bool to tell if found or not. It's easier to use after for the caller.

src/main.go
```diff
@@ -160,7 +160,7 @@ func main() {
 	flag.Parse()
 	env := &environment{}
 	env.init(args)
-	defer env.tracer.close()
+	defer env.close()
```
lnu, line 163: Good idea to have a global close

lnu:
All the code I've done is here https://github.com/boarder2/oh-my-posh3/pull/1/files
It's very rough but it's very similar to yours
I still wonder how to return a mocked cache from the mocked environment.

PR #990, opened 2021-09-23
feat: add caching for owm segment

lnu:
default timeout of 10 minutes
Original concept by boarder2

### Prerequisites

### Description

Original pull by @boarder2 (#946). Relies on the new global cache by @JanDeDobbeleer.

### DIFF
--- a/src/segment_owm.go
+++ b/src/segment_owm.go
@@ -21,8 +21,10 @@ const (
 	LOCATION Property = "location"
 	// UNITS openweathermap units
 	UNITS Property = "units"
-	// CACHETIMEOUT cache timeout
-	CACHETIMEOUT Property = "cachetimeout"
+	// CACHETimeout cache timeout
+	CACHETimeout Property = "cache_timeout"
+	// CACHEKey key used when caching the response
+	CACHEKey string = "owm_response"
 )
 
 type weather struct {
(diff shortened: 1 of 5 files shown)


src/segment_owm.go
```diff
@@ -61,22 +63,57 @@ func (d *owm) string() string {
...
 	timeout := d.props.getInt(HTTPTimeout, DefaultHTTPTimeout)
+	cachetimeout := int64(d.props.getInt(CACHETIMEOUT, 10))
+
+	d.url = fmt.Sprintf("http://api.openweathermap.org/data/2.5/weather?q=%s&units=%s&appid=%s", location, units, apikey)
+
+	if cachetimeout > 0 {
+		// check if data stored in cache
+		val, found := d.env.cache().get("owm_response")
+		// we got something from te cache
+		if found {
```
JanDeDobbeleer, line 81: Do an early return with `!found`

src/segment_owm.go
```diff
@@ -21,6 +21,8 @@ const (
 	LOCATION Property = "location"
 	// UNITS openweathermap units
 	UNITS Property = "units"
+	// CACHETIMEOUT cache timeout
+	CACHETIMEOUT Property = "cachetimeout"
```
JanDeDobbeleer: Mismatch in doc and value `cache_timeout`, would also use the underscore here.

src/segment_owm.go
```diff
@@ -61,22 +63,57 @@ func (d *owm) string() string {
...
-		return err
+		return new(OWMDataResponse), err
 	}
 	q := new(OWMDataResponse)
 	err = json.Unmarshal(body, &q)
+	if err != nil {
+		return new(OWMDataResponse), err
+	}
+
+	if cachetimeout > 0 {
```
JanDeDobbeleer: Also use an early return here

JanDeDobbeleer (changes requested):

src/segment_owm.go
lnu: my bad... and the schema.json property has also been added on the wrong segment.

JanDeDobbeleer (approved):

src/segment_owm.go
lnu: in that case, if not found in the cache we continue the flow and call the service. If found we exit. Does an early return with !found apply here?

boarder2:
Thanks for taking this and running with it @lnu ! Sorry I wasn't able to get back to it quickly!

PR #991, opened 2021-09-24
fix(pwsh): validate correct AZ context object

JanDeDobbeleer:
### Prerequisites

### Description

[Description of the change]

### DIFF
--- a/src/init/omp.ps1
+++ b/src/init/omp.ps1
@@ -58,7 +58,7 @@ function global:Initialize-ModuleSupport {
     if ($env:AZ_ENABLED -eq $true) {
         try {
             $context = Get-AzContext
-            if ($null -ne $subscription) {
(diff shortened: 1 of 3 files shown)


lnu:
I noticed the casing of const is not the same in all places. Sometimes it's all uppercase, sometimes PascalCase. You started with PascalCase I think, right?

JanDeDobbeleer:
@lnu indeed. That's why I aligned it here. That's something I didn't take into account before in reviews. Maybe the uppercase format makes more sense for visibility?

lnu:
@JanDeDobbeleer it's not really clear what should be done. Even in the golang library they don't always follow the same pattern. 
The thing is const starting with an uppercase are exported and those starting in lowercase not(in our case there's only one package).
https://golang.org/doc/effective_go#constants

PR #992, opened 2021-09-24
fix(owm): cache also the url used to get the data

lnu:
The url is used after do generate the hyperlink

### Prerequisites

### Description

The url was blank when the response was cached. There were two solutions:

1. Moving back the getprops at the top
2. Caching also the url used to call the service

I did the second one. Don't know which one is better, but caching the url seemed logic.


JanDeDobbeleer (approved):
Makes sense like this, less work, more efficiency

Issue #1017, opened 2021-10-03
Smarter handling of path for storing .omp.cache


ralish:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

PR #982 was merged recently (nice addition!) and as a result a `.omp.cache` file is stored in the user's home directory. That's fine on Unix-like systems, where it's not unusual for configuration files to be stored directly in `$HOME` (and they're hidden anyway). On Windows, configuration files are typically not stored directly under the user's home directory, nor are files or folders with a dot prefix hidden by default.

For those among us who like to keep their home directory tidy, the current behaviour of storing the cache file directly under the home directory isn't ideal. I'd suggest selecting the path to store the cache file a bit more intelligently based on the platform.

**Windows**
The "conventional" path would be something like `$env:LOCALAPPDATA\oh-my-posh\.omp.cache`, which expands to something like: `C:\Users\user\AppData\Local\oh-my-posh\.omp.cache`. The `$env:LOCALAPPDATA` path is intended for application data which is local to the system (i.e. does not roam with the user profile). Cache files are a perfect example of such data.

**Unix-like**
Less annoying perhaps, but it would be nice to conform to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). This is pretty simple with the relevant part being:
> $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

So the ideal case would be something like `$env:XDG_CACHE_HOME/oh-my-posh/.omp.cache`, where `$env:XDG_CACHE_HOME` is `$HOME/.cache` if not set.

**Anything else and/or failure cases**
Fallback to the current approach :-)

ralish:
Wow, unbelievably fast turnaround! Lgtm, with the only comment being that storing the cache in its own directory is I think slightly preferable, as it makes it unambiguous which application owns the file (i.e. `$env:LOCALAPPDATA\oh-my-posh\whatever` or `$env:XDG_CACHE_PATH/oh-my-posh/whatever`). That said, this is still a big improvement over the current setup!

JanDeDobbeleer:
@ralish it creates that folder, so it should be OK

PR #1020, opened 2021-10-03
feat(cache): set cache directory correctly

JanDeDobbeleer:
### Prerequisites

### Description

See #1017 for context

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -85,6 +85,7 @@ type environmentInfo interface {
 	isWsl() bool
 	stackCount() int
 	getTerminalWidth() (int, error)
+	getCachePath() string
 	cache() cache
 	close()
 }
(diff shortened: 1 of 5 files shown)


JanDeDobbeleer:
@ralish this should do it, can you run through the logic? I validated this to work already.

src/environment.go
```diff
@@ -501,3 +502,19 @@ func cleanHostName(hostName string) string {
 	}
 	return hostName
 }
+
+func returnOrBuildCachePath(path string) string {
+	// validate root path
+	if _, err := os.Stat(path); err != nil {
+		return ""
+	}
+	// validate oh-my-posh folder, if non existent, create it
```
JanDeDobbeleer, line 511: @ralish it will always create the oh-my-posh folder inside the cache directory. Meaning that we will write to for example `$env:XDG_CACHE_PATH/oh-my-posh/omp.cache` in this specific use-case.

PR #1028, opened 2021-10-08
fix(wsl): spotify support

bewing:
### Prerequisites

### Description

Support Spotify segment inside WSL

### DIFF
--- a/src/segment_git_test.go
+++ b/src/segment_git_test.go
@@ -58,6 +58,7 @@ func TestGetGitOutputForCommand(t *testing.T) {
 	commandArgs := []string{"symbolic-ref", "--short", "HEAD"}
 	want := "je suis le output"
 	env := new(MockedEnvironment)
+	env.On("isWsl", nil).Return(false)
 	env.On("runCommand", "git", append(args, commandArgs...)).Return(want, nil)
 	env.On("getRuntimeGOOS", nil).Return("unix")
 	g := &git{
@@ -92,6 +93,7 @@ type detachedContext struct {
 
 func setupHEADContextEnv(context *detachedContext) *git {
 	env := new(MockedEnvironment)
+	env.On("isWsl", nil).Return(false)
 	env.On("hasFolder", "/rebase-merge").Return(context.rebaseMerge)
 	env.On("hasFolder", "/rebase-apply").Return(context.rebaseApply)
 	env.On("hasFolder", "/sequencer").Return(context.sequencer)
@@ -530,6 +532,7 @@ func TestParseGitStatsInvalidLine(t *testing.T) {
 
 func bootstrapUpstreamTest(upstream string) *git {
 	env := &MockedEnvironment{}
+	env.On("isWsl", nil).Return(false)
 	env.On("runCommand", "git", []string{"--no-optional-locks", "-c", "core.quotepath=false", "-c", "color.status=false", "remote", "get-url", "origin"}).Return(upstream, nil)
 	env.On("getRuntimeGOOS", nil).Return("unix")
 	props := &properties{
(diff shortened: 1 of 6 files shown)


src/segment_git_test.go
```diff
@@ -58,6 +58,7 @@ func TestGetGitOutputForCommand(t *testing.T) {
 	commandArgs := []string{"symbolic-ref", "--short", "HEAD"}
 	want := "je suis le output"
 	env := new(MockedEnvironment)
+	env.On("isWsl", nil).Return(false)
```
JanDeDobbeleer, line 61: why is this added?

src/segment_spotify_others.go
```diff
@@ -2,6 +2,32 @@
...
+import (
+	"encoding/csv"
+	"strings"
+)
+
 func (s *spotify) enabled() bool {
+	if !s.env.isWsl() {
+		return false
+	}
+	tlist, err := s.env.runCommand("tasklist.exe", "/V", "/FI", "Imagename eq Spotify.exe", "/FO", "CSV", "/NH")
```
JanDeDobbeleer: This will only work on WSL, can we rename the file to `segment_spotify_wsl.go`?

src/segment_spotify_others.go
```diff
@@ -2,6 +2,32 @@
...
+	if err != nil || strings.HasPrefix(tlist, "INFO") {
+		return false
+	}
+	records, err := csv.NewReader(strings.NewReader(tlist)).ReadAll()
+	if err != nil || len(records) == 0 {
+		return false
+	}
+	for _, record := range records {
+		title := record[len(record)-1]
```
JanDeDobbeleer: what if `len(record) == 0`?

src/segment_spotify_others_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+type spotifyArgs struct {
+	csv      string
+	runError error
+}
+
+func bootStrapSpotifyOtherTest(args *spotifyArgs) *spotify {
```
JanDeDobbeleer: Can you use the new style we have like in [this](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segment_ruby_test.go#L10) test for example?

JanDeDobbeleer (changes requested):
A few remarks/questions. Would be good to rename the file(s).

src/segment_git_test.go
bewing: This is added as it would panic without it due to the change at https://github.com/JanDeDobbeleer/oh-my-posh/pull/1028/files#diff-937c113f57efe859eff30ab6e3674e71a981563d47c10f826aeefe8e2e0f7c1bR149

src/segment_spotify_others.go
bewing: Covered in test at https://github.com/JanDeDobbeleer/oh-my-posh/pull/1028/files#diff-4e9c0a52809c48bdd0100d81eb90ccbec75b3f047fbef841b265e0bb4cc93e73R19

src/segment_spotify_others_test.go
bewing: Will refactor this weekend

src/segment_spotify_others.go
bewing: renamed

src/segment_spotify_others_test.go
bewing: Refactored in latest commit.

JanDeDobbeleer:
@bewing the commit can't start with an upper case. So you'll have to rename to `fix(wsl): spotify support`

JanDeDobbeleer:
@all-contributors please add @bewing for code

PR #1050, opened 2021-10-16
fix(path): correctly handle replaced paths for letter style

aexvir:
### Prerequisites

### Description

Currently, when using the `letter` style on the `path` segment, together with `mapped_locations` that replace directories with icons, the icon gets malformed as the mapped locations are applied before the segment style is applied, and the segment code naively gets the first character of the folder for every level.

This also affects directories that have characters on the extended unicode range, like `äcker`, as getting the first character from that string doesn't pick up the whole letter `ä`.

example:
Screenshot 2021-10-16 at 08 32 35

This merge request addresses this problem by extracting from the folder name the first letter, taking in account the unicode extended range, and not doing any replacement in case there is no letter on the folder at all, which would happen when the icon gets replaced by the mapped locations processor, or if the folder name would be exclusively formed by non-letters nor numbers, which I think it's unlikely to happen

so now it looks like:
Screenshot 2021-10-16 at 08 46 57

---

note:
currently, directories like python's `__pycache__` are get shortened to just `_`; as I wanted to handle the case where the replacement would include an icon and some text, e.g. `➼ something`, I've changed this logic, and shorten to until the first letter that was found instead
* for `__pycache__` it would get shortened to `__p`
* for `➼ something` it would get shortened to `➼ s`

### DIFF
--- a/docs/docs/segment-path.md
+++ b/docs/docs/segment-path.md
@@ -111,4 +111,10 @@ for the folders to display is governed by the `mixed_threshold` property.
 
 ### Letter
 
-Works like `Full`, but will write every subfolder name using the first letter only.
+Works like `Full`, but will write every subfolder name using the first letter only, except when the folder name
+starts with a symbol or icon.
+
+- `folder` will be shortened to `f`
+- `.config` will be shortened to `.c`
+- `__pycache__` will be shortened to `__p`
+- `➼ folder` will be shortened to `➼ f`
--- a/src/segment_path.go
+++ b/src/segment_path.go
@@ -155,12 +155,20 @@ func (pt *path) getLetterPath() string {
 		if len(folder) == 0 {
 			continue
 		}
-		var letter string
-		if strings.HasPrefix(folder, ".") && len(folder) > 1 {
-			letter = folder[0:2]
-		} else {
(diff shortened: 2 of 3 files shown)


JanDeDobbeleer (approved):

JanDeDobbeleer:
@aexvir how did you create those screenshots?

JanDeDobbeleer (changes requested):
@aexvir only the CI reported issues need to be addressed and we're good to go!

aexvir:
> @aexvir how did you create those screenshots?

it's just [iterm2](https://gitlab.com/gnachman/iterm2) on macOS; made the font bigger so the prompt could be easily visible, then I did a [window screenshot](https://support.apple.com/en-us/HT201361#:~:text=How%20to%20capture%20a%20window%20or%20menu)

> @aexvir only the CI reported issues need to be addressed and we're good to go!

should be fixed now 🙂

JanDeDobbeleer:
> then I did a [window screenshot](https://support.apple.com/en-us/HT201361#:~:text=How%20to%20capture%20a%20window%20or%20menu)

I had no idea that was possible, thanks!

PR #1130, opened 2021-10-28
chore: exclude themes from release

JanDeDobbeleer:
### Prerequisites

### Description

Baed on some discussions, @lnu thoughts?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,6 @@ on:
     paths:
       - 'src/**'
       - 'packages/**'
-      - 'themes/**'
 
 jobs:
   release:


lnu:
It’s getting out of control with themes these days :).
Should we provide a cmdlet to list all available themes maybe? Can be done later.

lnu:
I did something months ago: https://github.com/lnu/oh-my-posh/commit/c1981babc73334adb32b0da472ed713f06bb06c8

but I think there was a github api limit in terms of request

JanDeDobbeleer:
@lnu the issue is that I can see things get out of sync really fast that way. Suppose you do not upgrade but still pull in latest things are broken. Now, obviously this can be resolved if we only fetch the themes for your release at which point it's completely bonkers and you might as well just ship them...

JanDeDobbeleer:
@lnu merging this. It will only release code changes and still bundle latest themes, so that might be enough to avoid releases with "only themes". We'll still release a ton though :-)

SharakPL:
Thanks :) Fixes and new features are good. Though if you could make bigger updates less frequently, milestones maybe and update once a week/month then it would be much better. Or maybe make the updates work like with oh-my-zsh, where I get prompted right on terminal start about new update and just hit Enter. I don't have to manually `Update-Module` and then `Uninstall-Module oh-my-posh -RequiredVersion 5.9.0` to free the space. It is quite big ~66 MB each so I bet with just a week of regular updates I'd hit 1 GB 😅 

...or make it 3 days! I've just noticed that unpacked each version is almost 130 MB 😱 Why all this in /bin when each user needs just 1 of them? I know oh-my-posh is cross platform but isn't there a way to make it smaller for each user?
obraz

JanDeDobbeleer:
A few things:

- if you want to solve that right now, install using [winget or scoop](https://ohmyposh.dev/docs/windows) as that doesn't have the disadvantage of storing previous versions
- I'm not going to solve PowerShell's crazy version management solution, there are [scripts](https://luke.geek.nz/powershell/remove-old-powershell-modules-versions-using-powershell/) to facilitate that
- we could indeed only download the relevant binary rather than ship all options

SharakPL:
Actually Scoop also keeps old versions, but there's `scoop clean` command to remove all not linked as current. It even keeps all installers, but `scoop cache rm` removes them. I'm not yet familiar with Winget, but I suspect it's all the same.

JanDeDobbeleer:
@SharakPL Winget can maybe cache them, but the installer overrides. So that should never result in multiple binaries. But once again, ecosystem problem, not the tool's in my opinion.

PR #1160, opened 2021-11-02
devcontainer: Deploy OMP to all shells, expose build arg for timezone & other minor enhancements

darkvertex:
### Prerequisites

### Description

I thought it was a shame that the devcontainer for a custom prompt project wasn't using its own prompt. Here's a breakdown of the changes:

#### Minor features / improvements:

* Deploy OMP prompt to Powershell, Bash, Fish and ZSH default shells.
* Deploys latest OMP themes at build time.
* Add build arg to choose a default OMP theme for your prompt.
* Add build arg to override devcontainer timezone since not everyone is on UTC.

#### Minor fixes:

* One RUN for Microsoft GPG keys import (no leftover .deb, smaller final image.)
* Clean apt-get cache (for a smaller final image.)
* Edit devcontainer name ("Go" was too vague.)
* devcontainer.json mentioned "tmux" but it wasn't getting installed, so I added it to the apt-get packages.
* Your repo has a `.editorconfig` file so I added the Editorconfig vscode extension to suggested devcontainer extensions.

Let me know what you think and if you want any changes or omissions I'm all ears.


darkvertex:
(Rebased to appease the commitlint gremlins. Sorry for the noise.)

JanDeDobbeleer:
@darkvertex noise is good 😃. Love this. Has been on my list to check for a while so kudos on this obvious improvement 🙏🏻

I'll check things in the morning as I'm a bit tired after traveling 😅

lnu:
Nice work. It's something I wondered when I pushed the devcontainer config: should we use omp inside the devcontainer or not?  Or should we keep it "clean" for dev only?  
Thanks for the pull :)

.devcontainer/Dockerfile
```diff
@@ -8,16 +8,74 @@ FROM mcr.microsoft.com/vscode/devcontainers/go:0-${VARIANT}
...
-    && apt-get install -y --no-install-recommends fish\
-    && apt-get install -y --no-install-recommends powershell
+    && apt-get install -y --no-install-recommends \
+        fish \
+        powershell \
+        tmux \
+    && apt-get clean
 
+# Deploy latest stable oh-my-posh as a starting point:
+RUN pwsh -Command Install-Module oh-my-posh -Scope AllUsers -Force && \
```
JanDeDobbeleer, line 25: Should we do this or symlink/volume mount the `packages/powershell/oh-my-posh` folder?

.devcontainer/Dockerfile
```diff
@@ -8,16 +8,74 @@ FROM mcr.microsoft.com/vscode/devcontainers/go:0-${VARIANT}
...
+    ln -sf $POSH_BIN /usr/bin/oh-my-posh
+# NOTE: devcontainers are Linux-only at this time but when
+# Windows or Darwin is supported someone will need to improve
+# the code logic above.
+
+# Deploy all latest oh-my-posh themes at devcontainer build time:
+ARG USERNAME=vscode
+ARG POSHTHEMES_ROOT=/home/${USERNAME}/.poshthemes
+ENV POSHTHEMES_ROOT=${POSHTHEMES_ROOT}
+RUN mkdir -m=777 -p ${POSHTHEMES_ROOT} && \
```
JanDeDobbeleer, line 48: Same for themes, we can make use of `docker-compose` to volume mount the themes inside the container's folder.

.devcontainer/Dockerfile
darkvertex: We'd need to add a compilation step because that folder doesn't have any of the binaries.

Alternatively I could download the good binary via wget and plop it in the right place; it just seemed simpler this way.

If you wanted technically we could compile OMP at build time, but I opted against it because I figured it was safer to grab the latest "stable" released version instead of the freshest main branch.

.devcontainer/Dockerfile
darkvertex: From your perspective, what would be the benefit vs unzipping latest at build time?

lnu: For me, the devcontainer is a tool to help developing/testing omp. I'm not sure it's the best thing to set up omp in it and use it directly. I would be better to give automation through tasks, like building and executing omp in a specific shell. Because if omp is configured with the latest version and I work on it, how do I test the version I'm working on?

JanDeDobbeleer: What I would expect, not knowing the details yet (I should really look into this) is the following:
- container gets built with all dependencies to develop
- the relevant parts are volume mounted (see docker-compose), like the PowerShell module (no need to install, always latest code when changing things)
- it builds the latest code live upon start (so we have oh-my-posh working based on the code)
- I can trigger a new build for all shells/ps module whenever I want

lnu: > What I would expect, not knowing the details yet (I should really look into this) is the following:
> 
> * container gets built with all dependencies to develop
_**ok**_
> * the relevant parts are volume mounted (see docker-compose), like the PowerShell module (no need to install, always latest code when changing things)
**_can symlink from the user profile to the scripts in the folder_**
> * it builds the latest code live upon start (so we have oh-my-posh working based on the code)
**_not sure about this one. Unless you restart the container, you always use th same instance. We can build each time by using `onCreateCommand` in devcontainer.json but it slows down the container start and probably that yo know when you want to build_**  
> * I can trigger a new build for all shells/ps module whenever I want
_**can create a new command in `tasks.json`**_

JanDeDobbeleer: @lnu we could indeed skip the third one as long as you have the task to make it up-to-date. This was me writing what I'd expect without any constraints :-)

darkvertex: >> if omp is configured with the latest version and I work on it, how do I test the version I'm working on?

So, the way the PR is at the time I write this, you could technically enter Bash and rebuild over the symlink source path, so something like this should work:

```shell
bash
cd src
go build -o $(readlink /usr/bin/oh-my-posh)
```
(I ran "bash" because it's currently configured to start your shell in Powershell by default.)

Doing so would overwrite your live install and immediately replace your working shell with your own current version.

Could also predeploy some aliases to do this for you at build time, if you think that's a good idea. Adding to tasks.json is an option as well, but remember you can't make tasks appear *only* within the devcontainer, so folks who clone this in VSCode and don't use the remote containers extension may accidentally run the task they weren't supposed to, so we'd probably want to name such a task something obvious like "_devcontainer: rebuild installed_".

JanDeDobbeleer:
How do we bring this one forward? Merge it as is, or continue with some of the additional improvements if feasible?

darkvertex:
I'm personally happy with the state of the PR but if you'd like any changes I'm happy to oblige.

I strongly believe starting from the latest stable release at build time is better than dealing with mounts or rebuilding from the repo directly, and it ensures we begin from a safe stable starting point onto which one may rebuild and experiment, and if you ever mess up horribly just rebuild your devcontainer and voilà, all stable and good again.

Would you like me to add that vscode Task to rebuild the devcontainer's OMP install as per that earlier comment? It would be convenient.

JanDeDobbeleer:
@darkvertex yes, let's add the task and then merge this very useful change!

darkvertex:
> @darkvertex yes, let's add the task and then merge this very useful change!

I went ahead and added it. 👍

I also took the liberty of adding the handy (but not very originally-named) [_Tasks_](https://marketplace.visualstudio.com/items?itemName=actboy168.tasks) extension to the suggested devcontainer extensions. -- It adds VSCode _tasks_ to the statusbar as buttons and can be configured per task, which I did for the devcontainer one. It is gently colored and has a helpful tooltip on hover:

image

I think will be especially handy for first-time contributors as it brings an obvious build button right to their face. :) -- Also I configured the default devcontainer preferences to hide tasks buttons by default unless a task defines differently, so the original default build task won't be in the statusbar causing confusion.

Last but not least, I took the liberty of adding a new section in CONTRIBUTING.md about the codespaces environment explaining the basics. If you wish to rephrase or restyle anything, note I have "_Allow edits by maintainers_" enabled in the PR so go right ahead.

CONTRIBUTING.md
```diff
@@ -33,23 +69,23 @@ diverse, inclusive, and healthy community.
 Examples of behavior that contributes to a positive environment for the
 project include:
 
-* Demonstrating empathy and kindness toward other people
-* Being respectful of differing opinions, viewpoints, and experiences
-* Giving and gracefully accepting constructive feedback
-* Accepting responsibility and apologizing to those affected by our mistakes,
+- Demonstrating empathy and kindness toward other people
```
darkvertex, line 72: I didn't want to touch this section but there was a Github Actions markdownlint workflow blocking up things if I didn't correct it. 🤷‍♂️

lnu:
I got your changes. I tried some other things:

- mount themes and pwsh modules from local workspace to the container
- build and symlink  omp in postCreateCommand

This way, we can modify themes or debug powershell modules in the workspace and test it immediately in the shell.  
What do you think?

We should also add a bash and zsh config file like the fish one

darkvertex:
Mounts [don't work](https://code.visualstudio.com/docs/remote/devcontainerjson-reference) in _Github Codespaces_ though:
image

How could we support both devcontainer environments? 🤔 What if we kept the stable deployment but adjusted it to not use a versioned path and configured the mounts to mount on top of it? That way in _Github Codespaces_ you get something functional, and in VSCode straight you get your direct mount magic cherry on top. Good idea, I think?


> We should also add a bash and zsh config file like the fish one

We could but since _VSCode Remote Containers_ bundles its own bashrc and zshrc, I opted to patch them in case they came with something useful (as opposed to doing a full file copy like with the Fish config.) VSCode actually patches its own custom $PS1 shell configuration (and I'm not sure about Github) so it seemed more correct and less disruptive to comment it out via patching than to re-set OMP over it at the end of the file.

We could do a full bashrc and zshrc file copy replace if we're super duper sure both VSCode Remote Containers and Github Codespaces didn't put anything of importance in it. I haven't spent time comparing them to give you an educated answer on that.

lnu:
> Mounts [don't work](https://code.visualstudio.com/docs/remote/devcontainerjson-reference) in _Github Codespaces_ though: image
> 
> How could we support both devcontainer environments? 🤔 What if we kept the stable deployment but adjusted it to not use a versioned path and configured the mounts to mount on top of it? That way in _Github Codespaces_ you get something functional, and in VSCode straight you get your direct mount magic cherry on top. Good idea, I think?
> 
> > We should also add a bash and zsh config file like the fish one
> 
> We could but since _VSCode Remote Containers_ bundles its own bashrc and zshrc, I opted to patch them in case they came with something useful (as opposed to doing a full file copy like with the Fish config.) VSCode actually patches its own custom $PS1 shell configuration (and I'm not sure about Github) so it seemed more correct and less disruptive to comment it out via patching than to re-set OMP over it at the end of the file.
> 
> We could do a full bashrc and zshrc file copy replace if we're super duper sure both VSCode Remote Containers and Github Codespaces didn't put anything of importance in it. I haven't spent time comparing them to give you an educated answer on that.

I did not understand you tried to make it work both for devcontainer and codespaces. It makes sense now.  
But codespaces is not free at all and I'm on the beta list for ages now 🤔.

darkvertex:
> But codespaces is not free at all and I'm on the beta list for ages now 🤔.

Yeah. I've been on that beta list for ages now too. Not sure when it'll run out of free VMs but I'm enjoying it while it's there. 🤷‍♂️

JanDeDobbeleer:
Alright, so what do we do now? Merge it and continue from there if we see improvements? I didn't validate this one, relying in your expertise here (will look at it asap though, been on my list way too long).

lnu:
let's merge it. It works as is anyway and it won't trigger a build. So we can still adapt it when needed without disturbing anyone.

PR #1174, opened 2021-11-05
fix: Apply consistent casing for drive letters and only drive letters

### LINKED ISSUE #1173: Inconsistent capitalization for path
### Code of Conduct

### What happened?

When formatting the path, inconsistent capitalization is used, with `c:`  and only `c:` replaced with `C:` indiscrimately.

This is the table of what I'm getting and what I should be getting (note that it's possible to have custom PS drives):

This, surprisingly  only happens when testing (e,g. with `Get-PoshThemes`) not when actually using the program.

In all cases, the command is `.\oh-my-posh.exe -config ..\themes\hunk.omp.json -pwd ENTER_PATH_HERE`. I'm using hunk as the simplest demo, this happens to many other themes. I have also removed extraneous powerline characters, for clarity.

| path | result | expected | status |
|---|---|---|---|
|C:\Windows\ |C:\Windows\ |C:\Windows\ | :white_check_mark: |
|src:\test\test|srC:\test\test |src:\test\test | :x: |
|g:\Other\ | g:\Other\ | G:\Other\ | :x: |
|test:\test\ | test:\test\ | test:\test\ |  :white_check_mark: |
|testb:\test\ | testb:\test\ | testb:\test\ |  :white_check_mark: |
|testc:\test\ | testC:\test\ | testc:\test\ |  :x: |
|HKLM:\SOFTWARE\testc:test\ | HKLM:\SOFTWARE\testC:test\ | HKLM:\SOFTWARE\testc:test\ | :x: |

I've located the problem [at the `environment.go` file, line 188](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/environment.go#L188), where the `correctPath` function is being too naïve about Windows paths. Other drive letters than `C:` are obviously possible, you can make custom PSDrives that end in `c` (obviously the full name is greater) and you can have colons in the middle of registry keys.

The correct way would be to replace "Letter + : at the start of the string only" (if this is even necessary), not just `c:` to `C:` indiscriminately 

### Version

5.18.0

### Theme

hunk

Windows

powershell

### Relevant log output

_No response_

gschizas:
resolved #1173

### Prerequisites

### Description

Apply proper casing for all drive letters, not just `C:`. Also, ignores anything that's not really a pure drive letter, such as all the rest of the custom PSDrives (it was only by luck that none of the default non-drive-letters PSDrives end in `c`), and `c:` in the middle of the path.

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -186,7 +186,7 @@ func (env *environment) getcwd() string {
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
 		driveLetter := getCompiledRegex(`^[a-z]:`)
-		return driveLetter.ReplaceAllStringFunc(pwd, func(letter string) string {return strings.ToUpper(letter)})
+		return driveLetter.ReplaceAllStringFunc(pwd, strings.ToUpper)
 	}
 	if env.args != nil && *env.args.PWD != "" {
 		env.cwd = correctPath(*env.args.PWD)
--- a/src/environment.go
+++ b/src/environment.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -185,7 +186,8 @@ func (env *environment) getcwd() string {
 	}
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
-		return strings.Replace(pwd, "c:", "C:", 1)
+		driveLetter := regexp.MustCompile(`^[a-z]:`)
+		return driveLetter.ReplaceAllStringFunc(pwd, func(letter string) string {return strings.ToUpper(letter)})
 	}
 	if env.args != nil && *env.args.PWD != "" {
 		env.cwd = correctPath(*env.args.PWD)
--- a/src/environment.go
+++ b/src/environment.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -186,7 +185,7 @@ func (env *environment) getcwd() string {
 	}
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
-		driveLetter := regexp.MustCompile(`^[a-z]:`)
+		driveLetter := getCompiledRegex(`^[a-z]:`)
 		return driveLetter.ReplaceAllStringFunc(pwd, func(letter string) string {return strings.ToUpper(letter)})
 	}
 	if env.args != nil && *env.args.PWD != "" {
--- a/src/environment_test.go
+++ b/src/environment_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"flag"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -22,30 +21,3 @@ func TestHostNameWithLan(t *testing.T) {
 	cleanHostName := cleanHostName(hostName)
 	assert.Equal(t, "hello", cleanHostName)
 }
-
-func TestPathWithDriveLetter(t *testing.T) {
-	args := &args{
-		Debug: flag.Bool(
-			"debug",
-			false,
-			"Print debug information"),
-		PWD: flag.String(
-			"pwd",
-			"",
-			"the path you are working in"),
-	}
-
-	testPath := func(t *testing.T, inputPath string, expectedResult string)  {
-		env := &environment{}
-		env.init(args)
-		env.args.PWD = &inputPath
-		assert.Equal(t, env.getcwd(), expectedResult)
-	}
-	testPath(t, `C:\Windows\`, `C:\Windows\`)
-	testPath(t, `c:\Windows\`, `C:\Windows\`)
-	testPath(t, `p:\other\`, `P:\other\`)
-	testPath(t, `other:\other\`, `other:\other\`)
-	testPath(t, `src:\source\`, `src:\source\`)
-	testPath(t, `HKLM:\SOFTWARE\magnetic:test\`, `HKLM:\SOFTWARE\magnetic:test\`)
-}
-
--- a/src/environment_test.go
+++ b/src/environment_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


src/environment.go
```diff
@@ -185,7 +186,8 @@ func (env *environment) getcwd() string {
 	}
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
-		return strings.Replace(pwd, "c:", "C:", 1)
+		driveLetter := regexp.MustCompile(`^[a-z]:`)
```
JanDeDobbeleer: you can use getCompiledRegex(`^[a-z]:`) here. We have functionality that caches the regex once compiled.

src/environment_test.go
```diff
@@ -21,3 +22,30 @@ func TestHostNameWithLan(t *testing.T) {
 	cleanHostName := cleanHostName(hostName)
 	assert.Equal(t, "hello", cleanHostName)
 }
+
+func TestPathWithDriveLetter(t *testing.T) {
```
JanDeDobbeleer: Our main test cases are the other way around. See [this](https://github.com/JanDeDobbeleer/oh-my-posh/blob/d96cb7989a759996ae75e5ee45b95d6b798a7f96/src/segment_crystal_test.go#L10) example.

JanDeDobbeleer (changes requested):
Small remarks for consistency, but otherwise 👍🏻

src/environment_test.go
gschizas: What do you mean "other way around"?

gschizas: I'm guessing I should incorporate the tests inside segment_path_test.go instead of doing this in environment_test.go. Given that there are already some similar tests there.

gschizas: I'm having trouble running the tests. I'm getting the wrong answer for drive letters. It seems as if the drive letters are being left alone in the tests.

This is my test so far (that is ⅔ / 67% failing):

```go
func TestWritePathInfoWindowsOtherDriveLetters(t *testing.T) {
	home := homeBillWindows
	want := "C: > Windows"
	got := testWritePathInfo(home, "C:\\Windows\\", "\\")
	assert.Equal(t, want, got)

	want = "C: > Windows"
	got = testWritePathInfo(home, "c:\\Windows\\", "\\")
	assert.Equal(t, want, got)

	want = "P: > Other"
	got = testWritePathInfo(home, "p:\\Other\\", "\\")
	assert.Equal(t, want, got)
}
```

The 2nd and 3rd test are failing (but they were failing even before my patch)

src/environment.go
```diff
@@ -185,7 +185,8 @@ func (env *environment) getcwd() string {
 	}
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
-		return strings.Replace(pwd, "c:", "C:", 1)
+		driveLetter := getCompiledRegex(`^[a-z]:`)
+		return driveLetter.ReplaceAllStringFunc(pwd, func(letter string) string {return strings.ToUpper(letter)})
```
lnu: you can do this directly: `return driveLetter.ReplaceAllStringFunc(pwd, strings.ToUpper)`

src/environment_test.go
lnu: I took your changes and all tests are green on my side.

src/environment.go
gschizas: Much nicer, thanks. I'll make the change and push it again.

(I _did_ say I don't know Go 😄 )

src/environment_test.go
gschizas: Weird... I've added my standalone tests, and I'm guessing these will go red instead. I think there's something going on with the testAgnosterPath, but I haven't been able to debug it.

JanDeDobbeleer:
@gschizas about the tests, you can refactor them entirely like this:

```go
func TestWritePath(t *testing.T) {
	cases := []struct {
		Case          string
		Expected      string
		Home          string
		PWD           string
		PathSeparator string
	}{
		{Case: "Windows outside home", Expected: "C: > f > f > location", Home: homeBillWindows, PWD: "C:\\Program Files\\Go\\location", PathSeparator: "\\"},
		{Case: "Windows oustide home", Expected: "~ > f > f > location", Home: homeBillWindows, PWD: homeBillWindows + "\\Documents\\Bill\\location", PathSeparator: "\\"},
		{Case: "Windows inside home zero levels", Expected: "C: > location", Home: homeBillWindows, PWD: "C:\\location", PathSeparator: "\\"},
		{Case: "Windows inside home one level", Expected: "C: > f > location", Home: homeBillWindows, PWD: "C:\\Program Files\\location", PathSeparator: "\\"},
		{Case: "Windows lower case drive letter", Expected: "C: > Windows", Home: homeBillWindows, PWD: "c:\\Windows\\", PathSeparator: "\\"},
		{Case: "Windows lower case drive letter (other)", Expected: "P: > Other", Home: homeBillWindows, PWD: "P:\\Other\\", PathSeparator: "\\"},
		{Case: "Windows lower case drive letter (other, lower case)", Expected: "p: > Other", Home: homeBillWindows, PWD: "P:\\Other\\", PathSeparator: "\\"},
		{Case: "Windows lower word drive", Expected: "some: > some", Home: homeBillWindows, PWD: "some:\\some\\", PathSeparator: "\\"},
		{Case: "Windows lower word drive (ending with c)", Expected: "src: > source", Home: homeBillWindows, PWD: "src:\\source\\", PathSeparator: "\\"},
		{Case: "Windows lower word drive (arbitrary cases)", Expected: "sRc: > source", Home: homeBillWindows, PWD: "sRc:\\source\\", PathSeparator: "\\"},
		{Case: "Windows registry drive", Expected: "\uf013 > f > magnetic:test", Home: homeBillWindows, PWD: "HKLM:\\SOFTWARE\\magnetic:test\\", PathSeparator: "\\"},
		{Case: "Windows registry drive case sensitive", Expected: "\uf013 > f > magnetic:TOAST", Home: homeBillWindows, PWD: "HKLM:\\SOFTWARE\\magnetic:TOAST\\", PathSeparator: "\\"},
		{Case: "Unix outside home", Expected: "mnt > f > f > location", Home: homeJan, PWD: "/mnt/go/test/location", PathSeparator: "/"},
		{Case: "Unix inside home", Expected: "~ > f > f > location", Home: homeJan, PWD: homeJan + "/docs/jan/location", PathSeparator: "/"},
		{Case: "Unix outside home zero levels", Expected: "mnt > location", Home: homeJan, PWD: "/mnt/location", PathSeparator: "/"},
		{Case: "Unix outside home one level", Expected: "mnt > f > location", Home: homeJan, PWD: "/mnt/folder/location", PathSeparator: "/"},
	}

	for _, tc := range cases {
		props := &properties{
			values: map[Property]interface{}{
				FolderSeparatorIcon: " > ",
				FolderIcon:          "f",
				HomeIcon:            "~",
			},
		}
		env := new(MockedEnvironment)
		env.On("homeDir", nil).Return(tc.Home)
		env.On("getPathSeperator", nil).Return(tc.PathSeparator)
		env.On("getcwd", nil).Return(tc.PWD)
		env.On("getRuntimeGOOS", nil).Return("")
		args := &args{
			PSWD: &tc.PWD,
		}
		env.On("getArgs", nil).Return(args)
		path := &path{
			env:   env,
			props: props,
		}
		got := path.getAgnosterPath()
		assert.Equal(t, tc.Expected, got, tc.Case)
	}
}
```

JanDeDobbeleer:
@gschizas I was going through the changes because I wanted to check why the tests fail. You need to test this in environment as it's mocked in the `path` segment (due to obvious reasons). I'll push a change and you'll stay the contributor ;-)

JanDeDobbeleer:
#1176

JanDeDobbeleer:
@all-contributors please add @gschizas for code

Issue #21, opened 2020-09-23
Support for CMD


rashil2000:
Since this is a (almost) platform-agnostic shell themer, is there a way to support Command Prompt in Windows? 

For inspiration, you can check out how Cmder does it. 
For the most part, it just requires changing the `prompt` environment variable. Like this, for example: 
```
SETX /M PROMPT "$E[1;32;40m→ $E[1;36;40m$p$E[1;35;40m› $E[1;37;40m" 
```

JanDeDobbeleer:
In theory you should be able to call the executable just like you would on Powershell. Let me check and add that info to the README

JanDeDobbeleer:
Never mind the previous activity, that was me fighting the Github Android app 😅

rashil2000:
> In theory you should be able to call the executable just like you would on Powershell. 

That'd be really slick!

JanDeDobbeleer:
You'd think this should be possible, right? Well, turns out my friend `CMD` only works with strings... I'm looking for a way to bypass that, it can print the prompt though :-)

rashil2000:
Yes I tried fighting with it too. Calling the `.exe` momentarily printed the prompt, but the PROMPT variable takes only strings and some escape characters, because specifying the path of `go-my-posh` in PROMPT just printed out the path :)

JanDeDobbeleer:
It's more limited to my friend `CMD` than anything else, but maybe there's an out-of-the-box solution nobody's thinking off.

JanDeDobbeleer:
Doesn't seem to be possible yet with everything I've tried, I'll keep this open in case I (or someone else) has a brilliant idea that might work.

rashil2000:
Sure!
Thanks for your effort!

JanDeDobbeleer:
Had a look once again and it's just not possible as far as I can tell.

rashil2000:
@JanDeDobbeleer Yes, it seemed really difficult. 

Any idea how Cmder achieved a custom prompt though? Was it terminal-specific?

JanDeDobbeleer:
You can change it once, but not change it on the fly and as elaborate as oh-my-posh does.

I will have a look though. If what conemu does works on every prompt invocation, it might work.

JanDeDobbeleer:
<img width="433" alt="image" src="https://user-images.githubusercontent.com/2492783/96491048-fa831180-1241-11eb-86b9-021f6cc126d3.png">
[...]

PR #1209, opened 2021-11-11
docs(installation): add guide to using oh-my-posh with CLink-based terminals on Windows

AbdelrahmanHafez:
re #21

### DIFF
--- a/docs/docs/install-shells.mdx
+++ b/docs/docs/install-shells.mdx
@@ -19,6 +19,7 @@ oh-my-posh --print-shell
     { label: 'bash', value: 'bash', },
     { label: 'fish', value: 'fish', },
     { label: 'nu', value: 'nu', },
+    { label: 'Cmder/Clink/ConEmu', value: 'Cmder/Clink/ConEmu', },
   ]
 }>
 <TabItem value="powershell">
@@ -109,5 +110,19 @@ config set prompt  "(oh-my-posh --config ~/.poshthemes/jandedobbeleer.omp.json |
 
 Restart nu shell for the changes to take effect.
 
+</TabItem>
+
+<TabItem value="Cmder/Clink/ConEmu">
+
+Create a new file in your Cmder/Clink/ConEmu installation directory `vendor\clink-completions\oh-my-posh.lua` with the following content, which assumes you have a config file in your `C:` partition called `oh-my-posh-config.json`.
+
+```lua
+local custom_prompt = clink.promptfilter(50)
+function custom_prompt:filter(prompt)
+    return io.popen("oh-my-posh --config C:\\oh-my-posh-config.json"):read("*a")
+end
+```
+
+For more information check [CLink's documentation on how to add oh-my-posh to your terminal.](https://chrisant996.github.io/clink/clink.html#oh-my-posh)
 </TabItem>
 </Tabs>


docs/docs/install-shells.mdx
```diff
@@ -109,5 +110,19 @@ config set prompt  "(oh-my-posh --config ~/.poshthemes/jandedobbeleer.omp.json |
 
 Restart nu shell for the changes to take effect.
 
+</TabItem>
+
+<TabItem value="Cmder/Clink/ConEmu">
```
JanDeDobbeleer, line 115: Why `Cmder/Clink/ConEmu` and not CMD?

AbdelrahmanHafez: Because there is currently no way to make oh-my-posh work with native CMD AFAIK, we have to use it with something that uses CLink.

JanDeDobbeleer: Help me understand why we can't add that as a prerequisite? Because we're now mixing terminals with shells, that's the issue I see here. As far as I'm aware, only CMD needs a specific clink integration as all other shells have the ability to run oh-my-posh (regardless of the terminal).

AbdelrahmanHafez: Excuse me if I'm mixing up the terminology, I am not aware of the difference between a terminal/shell.

I got my information from the discussion in #21, I understood from your comments that it's not possible to replace cmd native shell(?) with oh-my-posh. Is this not true anymore? Can we do that with good old CMD now?

AbdelrahmanHafez: I think I misunderstood your last comment, if what you mean that the real requirement here is CLink (shell), and ConEmu and Cmder  (terminals?) both use CLink, therefore the tab label should be just CLink, I think we should still keep all 3, because even though it would be technically correct, I anticipate that many users who use Cmder/ConEmu wouldn't know that they use CLink, and mentioning all 3 would make it easier for them to find it.

JanDeDobbeleer: What I meant is that the only shell where this is relevant (or useful) is CMD. For every other shell running inside ConEmu, CMDer that isn't very useful as you can set up oh-my-posh "natively" in the shell itself (Bash, PowerShell, WSL shells, ...). So the actual use-case is "How do I use oh-my-posh for CMD?" which has nothing to do with CMDer/Conemu.

JanDeDobbeleer: What we need is the following:
- Change the label to CMD
- Explain that CMD doesn't have native support for prompts, but you can have the same effect using clink

````md
CMD does not come with native support to change the prompt. You can however make use of clink
to display oh-my-posh.

Integrating oh-my-posh with Clink is easy: just save the following text to an oh-my-posh.lua file in your 
Clink scripts directory (run clink info to find that), and make sure the oh-my-posh.exe program is in a
directory listed in the %PATH% environment variable. (Or edit the script below to provide a fully qualified
path to the oh-my-posh.exe program.)

 ```lua
 local custom_prompt = clink.promptfilter(50)
 function custom_prompt:filter(prompt)
     return io.popen("oh-my-posh --config ~\\jandedobbeleer.omp.json"):read("*a")
 end
```

Restart CMD for the changes to take effect.
````

JanDeDobbeleer:
I have made it supercharged in #1221

JanDeDobbeleer:
@all-contributors please add @AbdelrahmanHafez for docs

PR #1228, opened 2021-11-15
feat: sysinfo segment

lnu:
### Prerequisites

### Description

SysInfo segments based on gopsutil. Replaces the current memory segment. By default, looks like the old memory segment.


src/segment.go
```diff
@@ -263,6 +265,7 @@ func (segment *Segment) mapSegmentWithWriter(env environmentInfo) error {
 		Memory:        &memory{},
 		Angular:       &angular{},
 		PHP:           &php{},
+		Processor:     &processor{},
```
JanDeDobbeleer: Do we want to add this here or rename the `memory` one to `device`?

lnu: It's pure WIP. I have no idea for the name. We could merge the host information(cpu, load, memory) into one segment with all properties available for the template. It would be easier for the maintainer to have only one segment but maybe someone wants to split the info into multiple and don't want to pay the price of getting all info for each segments.

JanDeDobbeleer: When using a plain text template that could be possible even without all the styling when they're sequential. To facilitate non-sequential segments, we could also cache the response for one run in environment?

lnu: of course, if we cache it we could have one segment usable multiple times without the burden of getting data.

lnu: opted for a sysinfo segment. Used gopsutil mem package. Still WIP.

lnu:
image

```
        {
          "type":"sysinfo",
          "powerline_symbol": "",
          "foreground": "#ffffff",
          "background": "#819c09",
          "invert_powerline": true,
          "properties": {
            "prefix": " \uf9c4 ",
            // "template":"{{ round .Times 2 }}({{ .Cores }} {{.Mhz }}Mhz)"
            "template":"{{ round .Times 2 }}({{ .Cores }} cores {{ round .Mhz 2 }}Mhz)"
          },
          "style":"powerline"
        },
        {
          "type":"sysinfo",
          "powerline_symbol": "",
          "foreground": "#ffffff",
          "background": "#9c1a09",
          "invert_powerline": true,
          "properties": {
            "postfix": " ",
            // "template":"{{ round .Times 2 }}({{ .Cores }} {{.Mhz }}Mhz)"
            "template":"{{ round .Load1 2 }} {{round .Load5 2}} {{round .Load15 2}}"
          },
          "style":"powerline"
        },
        {
          "type":"sysinfo",
          "powerline_symbol": "",
          "foreground": "#ffffff",
          "background": "#8f43f3",
          "invert_powerline": true,
          "properties": {
            "prefix": " \uf85a ",
            "postfix": "% ",
            // "template":"{{ round .Load 2 }}({{ .Cores }} {{.Mhz }}Mhz)"
            "template":"{{ round .PhysicalPercentUsed 2 }}"
          },
          "style":"powerline"
        },
```

lnu:
~~still have to update all memory segment used in themes~~
image

PR #1229, opened 2021-11-15
fix: node segment enable_version_mismatch

decart:
### Prerequisites

### Description

The `.nvmrc` file can contain a node version in [different formats](https://github.com/nvm-sh/nvm#nvmrc), but currently this behavior is not supported. This fix adds more formats of `.nvmrc` file content.

### DIFF
--- a/src/segment_node_test.go
+++ b/src/segment_node_test.go
@@ -7,15 +7,26 @@ import (
 )
 
 func TestNodeMatchesVersionFile(t *testing.T) {
+	nodeVersion := version{
+		Full:  "1.2.3",
+		Major: "1",
+		Minor: "2",
+		Patch: "3",
+	}
 	cases := []struct {
 		Case      string
 		Expected  bool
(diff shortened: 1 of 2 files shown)


src/segment_node_test.go
```diff
@@ -25,9 +36,7 @@ func TestNodeMatchesVersionFile(t *testing.T) {
 			language: &language{
 				env: env,
 				activeCommand: &cmd{
-					version: &version{
-						Full: tc.Version,
-					},
+					version: &tc.Version,
```
JanDeDobbeleer: if it's always the same, why not add the above version here directly?

JanDeDobbeleer (changes requested):

src/segment_node_test.go
decart: Yes, of course...

JanDeDobbeleer (approved):

PR #1250, opened 2021-11-18
cmd extended

JanDeDobbeleer:
### Prerequisites

### Description

- feat(cmd): add tooltip functionality
- feat(cmd): transient prompt

### DIFF
--- a/src/init/omp.lua
+++ b/src/init/omp.lua
@@ -47,7 +47,7 @@ local function error_level_option()
 end
 
 local function get_posh_prompt(rprompt)
-    local prompt_exe = string.format('::OMP:: --config="::CONFIG::" %s %s --rprompt=%s', execution_time_option(), error_level_option(), rprompt)
+    local prompt_exe = string.format('::OMP:: --shell=cmd --config="::CONFIG::" %s %s --rprompt=%s', execution_time_option(), error_level_option(), rprompt)
     prompt = io.popen(prompt_exe):read("*a")
     return prompt
 end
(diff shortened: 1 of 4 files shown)


src/init/omp.lua
```diff
@@ -56,7 +57,14 @@ function p:filter(prompt)
     return get_posh_prompt(false)
 end
 function p:rightfilter(prompt)
-    return get_posh_prompt(true), false
+    if tip_word == nil then
+        return get_posh_prompt(true), false
+    end
+    local prompt_exe = string.format('::OMP:: --config="::CONFIG::" --command="%s"', tip_word)
+    tooltip = io.popen(prompt_exe):read("*a")
+    if tooltip ~= nil and tooltip ~= "" then
```
chrisant996: Probably it should always `return tooltip, false`.  Otherwise if there's no rprompt string then it will allow other prompt filter functions to run, and something else's rprompt could get used by mistake.

I.e. I think remove lines 65 and 67 (but keep like 66).

chrisant996: On second thought it needs to return a string rather than nil, otherwise it won't override any prompt filters that might run before it.  So the return line should become `return (tooltip or ""), false`.

src/init/omp.lua
```diff
@@ -60,12 +60,19 @@ function p:rightfilter(prompt)
...
-    local prompt_exe = string.format('::OMP:: --config="::CONFIG::" --command="%s"', tip_word)
+    local prompt_exe = string.format('::OMP:: --shell=cmd --config="::CONFIG::" --command="%s"', tip_word)
     tooltip = io.popen(prompt_exe):read("*a")
     if tooltip ~= nil and tooltip ~= "" then
         return tooltip, false
     end
 end
+function p:transientfilter(prompt)
+    prompt = io.popen('::OMP:: --shell=cmd --config="::CONFIG::" --print-transient'):read("*a")
+    return prompt
```
chrisant996, line 75: If Clink has transient prompt turned on but the theme doesn't define a transient prompt, then an empty string is used, which causes the prompt to malfunction (e.g. <kbd>Enter</kbd> by itself erases the prompt and draws a new prompt in the same spot, making the key appear to have no effect).

Inserting the following between lines 70 and 71 (i.e. immediately preceding the `return prompt`) appears to fix that:

```lua
    if prompt == "" then
        prompt = nil
    end
```

src/init/omp.lua
JanDeDobbeleer: resolved it by using `return (tooltip or get_posh_prompt(true)), false`

chrisant996 (approved):

PR #1255, opened 2021-11-18
feature: Palette - a map of named Standard color values that can be referenced in theme segments

yehorb:
### Prerequisites

### Description

Introduces the Palette feature - a way to maintain a map of named Standard color values and use references to these colors in places where a Standard color is expected. Example:

```json
{
  "palette": {
    "darkgrey3": "#4C566A",
    "lightgrey2": "#ECEFF4",
    "red": "#BF616A",
    "orange": "#D08770",
    "yellow": "#EBCB8B",
    "green": "#A3BE8C",
    "purple": "#B48EAD",
    ...
  },
  "blocks": [
    {
      "segments": [
        {
          "type": "git",
          "style": "powerline",
          "powerline_symbol": "\uE0B0",
          "foreground": "p:lightgrey2",
          "background": "p:darkgrey3",
          "background_templates": [
            "{{ if or (.Working.Changed) (.Staging.Changed) }}p:yellow{{ end }}",
            "{{ if and (gt .Ahead 0) (gt .Behind 0) }}p:purple{{ end }}",
            "{{ if gt .Ahead 0 }}p:green{{ end }}",
            "{{ if gt .Behind 0 }}p:orange{{ end }}"
          ]
        },
        ...
        {
          "type": "battery",
          "style": "powerline",
          "invert_powerline": true,
          "powerline_symbol": "\uE0B2",
          "foreground": "p:lightgrey2",
          "background": "p:darkgrey3",
          "properties": {
            "color_background": true,
            "charged_color": "p:green",
            "charging_color": "p:yellow",
            "discharging_color": "p:red",
          }
        },
        ...
      ]
    }
  ]
}

```

I was trying to build my dream posh theme for the last few days and quickly understood that using only Hex color values was very limiting. They are not glanceable (I do not understand what they represent immediately), changing the same color in a lot of places was tedious, importing and maintaining a consistent palette was also a pain.

So I built the Palette feature and want to share it. The implementation solves all the issues I had - allows named colors and allows importing existing color themes more easily. Palette colors can be used in most of the places where the Standard color is expected - foreground, background, their template variant, and properties.

I've been using the updated binary in my PowerShell module for the last day, and it works well. Palette has no measurable impact on performance, both in benchmarks and in real-world usage, and does not affect anything but colors.

Implementation is mostly a decorator for Segment's #backgroud() and #foreground() methods (that is why colors from properties also work) and some refactoring to engine and Block. It is still a fairly large PR due to a lot of added scaffolding - mainly tests and benchmarks.

Hope you will find this feature useful. If not, it was a fun thing to do anyway, so no worries if you find these changes to be too much to approve.


JanDeDobbeleer:
@yehorb this has been a request before and I've been thinking about it again lately so I'm sure we'll be able to merge this. Allow me to find some time to review and we'll get this sorted 😉

JanDeDobbeleer:
Based on this I've written my own implementation. It's simpler and has logical limitations which I feel are OK but make the whole architectural design a lot less impactful.

I'll share it asap.

JanDeDobbeleer:
I did the following:

```diff
diff --git a/src/config.go b/src/config.go
index 738475bc..f3e72d67 100644
--- a/src/config.go
+++ b/src/config.go
@@ -29,6 +29,7 @@ type Config struct {
 	Blocks               []*Block          `config:"blocks"`
 	Tooltips             []*Segment        `config:"tooltips"`
 	TransientPrompt      *TransientPrompt  `config:"transient_prompt"`
+	Palette              map[string]string `config:"palette"`
 }
 
 type TransientPrompt struct {
diff --git a/src/main.go b/src/main.go
index b5c45108..31350371 100644
--- a/src/main.go
+++ b/src/main.go
@@ -212,6 +212,7 @@ func main() {
 		writer = &AnsiWriter{
 			ansi:               ansi,
 			terminalBackground: getConsoleBackgroundColor(env, cfg.TerminalBackground),
+			palette:            cfg.Palette,
 		}
 	}
 	title := &consoleTitle{
diff --git a/src/writer_ansi.go b/src/writer_ansi.go
index caf8706c..61b67247 100644
--- a/src/writer_ansi.go
+++ b/src/writer_ansi.go
@@ -60,6 +60,7 @@ type promptWriter interface {
 type AnsiWriter struct {
 	builder            strings.Builder
 	ansi               *ansiUtils
+	palette            map[string]string
 	terminalBackground string
 	Colors             *Color
 	ParentColors       *Color
@@ -158,17 +159,35 @@ func (a *AnsiWriter) write(background, foreground, text string) {
 	}
 
 	getAnsiColors := func(background, foreground string) (string, string) {
+		getPaletteColor := func(color string) string {
+			mapPaletteColor := func(color string) string {
+				color = strings.TrimLeft(color, "p:")
+				if color, found := a.palette[color]; found {
+					return color
+				}
+				return "#111111"
+			}
+			for strings.HasPrefix(color, "p:") {
+				color = mapPaletteColor(color)
+			}
+			return color
+		}
 		getColorString := func(color string) string {
+			color = getPaletteColor(color)
 			if color == Background {
-				color = a.Colors.Background
-			} else if color == Foreground {
-				color = a.Colors.Foreground
-			} else if color == ParentBackground && a.ParentColors != nil {
-				color = a.ParentColors.Background
-			} else if color == ParentForeground && a.ParentColors != nil {
-				color = a.ParentColors.Foreground
-			} else if (color == ParentForeground || color == ParentBackground) && a.ParentColors == nil {
-				color = Transparent
+				return getPaletteColor(a.Colors.Background)
+			}
+			if color == Foreground {
+				return getPaletteColor(a.Colors.Foreground)
+			}
+			if color == ParentBackground && a.ParentColors != nil {
+				return getPaletteColor(a.ParentColors.Background)
+			}
+			if color == ParentForeground && a.ParentColors != nil {
+				return getPaletteColor(a.ParentColors.Foreground)
+			}
+			if (color == ParentForeground || color == ParentBackground) && a.ParentColors == nil {
+				return Transparent
 			}
 			return color
 		}
diff --git a/themes/schema.json b/themes/schema.json
index 252d81f4..cba467a8 100644
--- a/themes/schema.json
+++ b/themes/schema.json
@@ -7,7 +7,7 @@
   "definitions": {
     "color": {
       "type": "string",
-      "pattern": "^(#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})|black|red|green|yellow|blue|magenta|cyan|white|default|darkGray|lightRed|lightGreen|lightYellow|lightBlue|lightMagenta|lightCyan|lightWhite|transparent|parentBackground|parentForeground|background|foreground)$",
+      "pattern": "^(#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})|p:\\w*|black|red|green|yellow|blue|magenta|cyan|white|default|darkGray|lightRed|lightGreen|lightYellow|lightBlue|lightMagenta|lightCyan|lightWhite|transparent|parentBackground|parentForeground|background|foreground)$",
       "title": "Color string",
       "description": "https://ohmyposh.dev/docs/config-colors",
       "format": "color"
```

This works according to the expectations you wrote (and theme you shared), plus it has the advantage to work for overrides as well. The problem I always had with the palette is that I also went in your direction, which made it leak everywhere for no apparent reason. I only came to this solution because I could see a different path while going through this PR, so you made it a reality. Maybe I missed something, I'm looking forward to feedback, ideas :-)

yehorb:
@JanDeDobbeleer cool, nice to hear I helped you to find the solution.
I have been tinkering with the code today, and also came to the conclusion that the Palette should be constrained to AnsiWriter, as it is the only part of the project that does "real" color processing anyway. I do agree that leaking it as deep as Blocks never felt right to me too.
Just as I was writing an update to say that Palette color will not work with Color overrides, I thought to myself "Wait, it doesn't sound right - palette colors not working with color overrides" - and dug through the code a bit more to find the color processing in AnsiWriter.
So how should we proceed? Should I close the pull request, and leave the implementation up to you? Or I can integrate the Palette class I have into AnsiWriter, just as you did, and you can review that version of the implementation.

JanDeDobbeleer:
@yehorb you can apply my diff and write the tests for it if you're up for it, it's fine to start from my branch. Would love to have you be able to finish this!

yehorb:
@JanDeDobbeleer, great, thank you for the feedback. I will work on the updated implementation then.

JanDeDobbeleer (changes requested):
As indicated by the initial review, but really great addition!

yehorb:
@JanDeDobbeleer, I believe I implemented the Palette according to your suggestion the best I could.
Palette is fully constrained to AnsiWriter as a part of a pluggable interface for color solving. The interface of AnsiWriter changed a bit, but this time it is much more constrained. Overall, I feel much better about this implementation, than the previous one.
I'm sorry for doing more than a minimal amount of changes (again), and all those force-pushes. I did not want to throw away the original code, so I just rebased the excessive changes away.
I hope you will find this implementation more manageable.

src/palette_colors_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/alecthomas/assert"
+)
+
+// FIXME: test is not completely deterministic, engineReder calls can have different
+// outputs due to time - there can be 1 second difference on the clocks during
+// rendering. The test is useful, and removing it completely will be incorrect - having
+// a close-to-real test of no palette vs palette rendering parity is better than not
+// having one. Not sure what to do with this one.
+func TestEngineRendersSamePrompt(t *testing.T) {
```
yehorb: Also, @JanDeDobbeleer, seeking some advice about this test. Do you have any ideas on how to make it correct?

.vscode/settings.json
```diff
@@ -2,7 +2,7 @@
   "go.lintTool": "golangci-lint",
   "go.useLanguageServer": true,
   "go.languageServerExperimentalFeatures": {
-    "diagnostics": false,
+    "diagnostics": false
```
JanDeDobbeleer, line 5: these changes are linter related, we might want to remove them :-)

docs/docs/config-colors.md
```diff
@@ -51,5 +51,15 @@ To change *only* the background color, just omit the first color from the above
...
 ```
 
+## Palette reference
+
+If your theme has a defined [Palette][palette], you can use the Palette reference in places where the
+__Standard color__ is expected.
+
+Palette references have 2 possible formats: `p:<palette key>` and `palette:<palette key>`.
+
+Read more about [Palette][palette].
```
JanDeDobbeleer: I would include palette here, rather than having a separate section.

docs/docs/config-palette.md
```diff
@@ -0,0 +1,94 @@
+---
+id: config-palette
+title: Palette
+sidebar_label: Palette
+---
+
+Palette allows you to define a set of named colors and use references to these colors in your theme.
+Palette reference can be used everywhere where a [Standard color][colors] is expected.
```
JanDeDobbeleer: references

docs/docs/config-colors.md
```diff
@@ -51,5 +51,15 @@ To change *only* the background color, just omit the first color from the above
 "prefix": "<,#FFFFFF>┏[</>",
 ```
 
+## Palette reference
+
+If your theme has a defined [Palette][palette], you can use the Palette reference in places where the
+__Standard color__ is expected.
+
+Palette references have 2 possible formats: `p:<palette key>` and `palette:<palette key>`.
```
JanDeDobbeleer: let's remove `palette:<palette key>`, not necessary.

src/cached_colors.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+// CachedColors is the AnsiColors Decorator that does simple color lookup caching.
+// AnsiColorFromSrting calls are cheap, but not free, and having a simple cache in
```
JanDeDobbeleer: typo: AnsiColorFromSrting -> AnsiColorFromString

src/config.go
```diff
@@ -29,6 +29,7 @@ type Config struct {
 	Blocks               []*Block          `config:"blocks"`
 	Tooltips             []*Segment        `config:"tooltips"`
 	TransientPrompt      *TransientPrompt  `config:"transient_prompt"`
+	Palette              Palette           `config:"palette"`
```
JanDeDobbeleer, line 32: this needs a pointer

src/engine_test.go
```diff
@@ -41,3 +43,81 @@ func TestCanWriteRPrompt(t *testing.T) {
 		assert.Equal(t, tc.Expected, got, tc.Case)
 	}
 }
+
+func BenchmarkEngineRender(b *testing.B) {
```
JanDeDobbeleer, line 47: isn't this more like a main test -> or easier to setup when you call main with args?

src/palette_colors_test.go
JanDeDobbeleer: I don't see why this is necessary, if it renders the colors as we expect, no need to compare with or without. This is an e2e test, for which I don't see the added benefit if your low level functions are properly unit tested.

JanDeDobbeleer (changes requested):
What I like:

- the color cache (valid point on speed increase)
- the color type, it helps with the `IsTransparent()` etc..

I still feel it's way too overengineered. I like code simple and readable, here I go from one file to another to understand what's going on. Palette can be part of `ansi_colors` (which we can name color as we'll always needs ANSI anyhow), the lookup, cache and method extensions can be done there as well.

I feel like we can meet in the middle of my simplistic implementation and your very useful solutions. But we're not quite there yet. Before we end up going in all directions (and you get demotivated), OK if I give it a go to see if we could agree on a certain middle ground?

src/palette_colors_test.go
yehorb: Makes sense, I will remove it then.

.vscode/settings.json
yehorb: I do see that this "fix" is not really relevant, but I doubt that I am the only one who will open the settings.json and do formatting subconsciously. This will pop up later. I think leaving it fixed is a better option.

docs/docs/config-colors.md
yehorb: Makes sense, I will change the docs accordingly.

src/config.go
yehorb: I thought about it, and... Does it really need a pointer? I'm just not sure. I looked this question up during development and found a few sources that state it is not required. As far as I understand, map types are reference types, they are already pointers. https://stackoverflow.com/questions/28384343/accessing-a-map-using-its-reference, https://go.dev/blog/maps. The same as Blocks, for example. It is []*Block, not *[]*Block, event though it can be. But still,  If a pointer is a requirement of the Config module or just a required convention, I can do that.

src/engine_test.go
yehorb: This is more like the main test. But I did not want to do the trick where the test binary calls itself with additional arguments, so I just copied most of the setup code into the test. Making it a dedicated main test is going to be better, but I am not sure how exactly to do that.

yehorb:
@JanDeDobbeleer, I understand. I will try to rehash the current implementation a bit, to make it easier to follow.

src/cached_colors.go
yehorb: Fixed.

docs/docs/config-colors.md
yehorb: Removed.

docs/docs/config-palette.md
yehorb: Fixed.

docs/docs/config-colors.md
yehorb: Moved palette docs to Palette section in config-colors.md.

src/palette_colors_test.go
yehorb: Removed.

yehorb:
@JanDeDobbeleer, I did the refactoring of the Palette and Colors. Now everything relevant to color solving resides in colors.go. I also refactored Palette docs - they are now included in the Palette section in config-colors.md. Fixed a few typos along the way. Ready for another review.

JanDeDobbeleer:
@yehorb I'm having a look tomorrow 👍🏻

docs/docs/config-colors.md
```diff
@@ -51,5 +51,93 @@ To change *only* the background color, just omit the first color from the above
...
+
+Should you use an invalid Palette reference as a color (for example typo `p:bleu` instead of `p:blue`),
+the Pallete engine will use the Transparent keyword as a fallback value. So if you see your prompt segments
+rendered with incorrect colors, and you are using a Palette, be sure to check the correctness of your references.
+
+### Limitations
+
+#### (Non) Recursive resolution
+
+Palette does not allow for recursive reference resolution. You should not reference Palette colors in other
```
JanDeDobbeleer: This is easily solveable though. You can see that in my initial sample, that was perfectlly possible.

src/config.go
JanDeDobbeleer: You are correct, I mistook it for a struct. Should've looked better.

src/properties.go
```diff
@@ -71,11 +71,10 @@ func (p *properties) getColor(property Property, defaultValue string) string {
 		return defaultValue
 	}
 	colorString := parseString(val, defaultValue)
-	_, err := getColorFromName(colorString, false)
-	if err == nil {
+	if IsAnsiColorName(colorString) {
 		return colorString
 	}
-	values := findNamedRegexMatch(`(?P<color>#[A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})`, colorString)
+	values := findNamedRegexMatch(`(?P<color>#[A-Fa-f0-9]{6}|[A-Fa-f0-9]{3}|(p|palette):.*)`, colorString)
```
JanDeDobbeleer: This one still references `palette:`

src/palette.go
```diff
@@ -0,0 +1,101 @@
...
+
+type Palette map[string]string
+
+const (
+	paletteKeyError          = "palette: requested color %s does not exist in palette of colors %s"
+	paletteRecursiveKeyError = "palette: resolution of color %s returned palette reference %s; recursive references are not supported"
+)
+
+var (
+	paletteColorPrefixes = [...]string{"palette:", "p:"}
```
JanDeDobbeleer: I would remove this and only use `p:`

src/palette.go
```diff
@@ -0,0 +1,101 @@
...
+			prefixFound = true
+			break
+		}
+	}
+
+	if !prefixFound {
+		return "", false
+	}
+
+	key := strings.ReplaceAll(colorName, prefix, "")
```
JanDeDobbeleer: this can be a simple `strings.TrimPrefix(colorName, prefix)`

src/palette.go
```diff
@@ -0,0 +1,101 @@
...
+		return colorName, nil
+	}
+
+	color, ok := p[key]
+	if !ok {
+		return "", &PaletteKeyError{Key: key, palette: p}
+	}
+
+	if _, ok = p.asPaletteKey(color); ok {
+		return "", &PaletteRecursiveKeyError{Key: key, Value: color}
```
JanDeDobbeleer: this is actually solvable by looping the above as long as you find the prefix and a match. In my sample I wrote the following:

```golang
getPaletteColor := func(color string) string {
    mapPaletteColor := func(color string) string {
        color = strings.TrimLeft(color, "p:")
        if color, found := a.palette[color]; found {
		    return color
	    }
	    return "#111111"
    }
    for strings.HasPrefix(color, "p:") {
	    color = mapPaletteColor(color)
    }
    return color
}
```

I know you could argue why, but call this experience, this is going to happen anyways. So we better get prepared if possible ;-)

JanDeDobbeleer (changes requested):
A few small remarks, we're almost there.

docs/docs/config-colors.md
yehorb: Not that it is not possible, I just think that is a bad idea. I was worried about performance initially because someone somewhere will build a long lookup chain just for the sake of it. With caching this is not that big of a threat, though, and can be enabled.

JanDeDobbeleer: I'm not expecting this to go fully crazy, a simple loop isn't going to do much imho.

yehorb: Added recursion and updated docs.

src/properties.go
yehorb: Removed "palette:" from valid references.

src/palette.go
yehorb: Removed "palette:" from valid references.

src/palette.go
yehorb: Correct, changed it.

src/palette.go
yehorb: This will run into infinite recursion in case of, for example, `"palette": {"blue": "p:blue"}`. I added the recursion (with depth check), but that actually made the code more complicated than it can be.

src/engine_test.go
JanDeDobbeleer: me neither, but we could try to find out :-)

src/palette.go
yehorb: And, in the case of `for` loop, infinite recursion will never terminate.

JanDeDobbeleer: that's solvable by adding 1 check:

```golang
getPaletteColor := func(color string) string {
    mapPaletteColor := func(color string) string {
        color = strings.TrimLeft(color, "p:")
        if color, found := a.palette[color]; found {
		    return color
	    }
	    return "#111111"
    }
    for strings.HasPrefix(color, "p:") {
	    refColor = mapPaletteColor(color)
            if refColor == color {
                break
            }
    return color
}
```

src/engine_test.go
JanDeDobbeleer: Let's keep it here for the time being, I'll try to refactor that later.

src/palette.go
```diff
@@ -0,0 +1,101 @@
...
+	errorStr := fmt.Sprintf(paletteRecursiveKeyError, p.Key, p.Value, p.depth)
+	return errorStr
+}
+
+// maybeResolveColor wraps resolveColor and silences possible errors, returning
+// Transparent color by default, as a Block does not know how to handle color errors.
+func (p Palette) MaybeResolveColor(colorName string) string {
+	color, err := p.ResolveColor(colorName)
+	if err != nil {
+		return Transparent
```
JanDeDobbeleer: In my tests, that was't the best idea. Ended up returning black.

yehorb: Do you mean that the tests failed on your machine? When was black returned instead of Transparent?

src/palette.go
JanDeDobbeleer: You're right. Did some tests, there's no simple around this without caching all entries in the path. I would set the depth to 3 through. No need to go crazy here.

src/palette.go
```diff
@@ -0,0 +1,101 @@
...
+	"sort"
+	"strings"
+)
+
+type Palette map[string]string
+
+const (
+	paletteKeyPrefix         = "p:"
+	paletteKeyError          = "palette: requested color %s does not exist in palette of colors %s"
+	paletteMaxRecursionDepth = 9 // allows 10 recusive resolutions
```
JanDeDobbeleer: I would set this to 3

JanDeDobbeleer (changes requested):
2 remarks and we can merge this little beauty.

src/palette.go
JanDeDobbeleer: No, transparent gives weird effects as the full default bg color on Powerline segments for example. Purely UI related.

src/palette.go
yehorb: What is the second remark?

JanDeDobbeleer: transparent as the default and this

src/palette.go
yehorb: So it should be "black" instead of Transparent? Or I can remove this altogether, this method is a remnant from the first implementation. It is not used anywhere but in tests.

JanDeDobbeleer: I would return `#1111111` I don't think we default to a color somewhere in case there's no match right?

yehorb: Maybe `""`, then? It will be transformed into `emptyAnsiColor` and then processed in `AnsiWriter#writeColoredText` accordingly. Will become white if it is for foreground or not used if it is background.

yehorb: And yes, there are no explicit "default" color returns.

JanDeDobbeleer: Perfect, let's go for `""`

yehorb: Done.

src/palette.go
yehorb: Reduced to depth 3.

yehorb:
@JanDeDobbeleer looks like all of the fixes for now.
And also looks like `commitlint` is marking commits with any capital letters in the subject line as invalid. Should I rebase to make them all lowercase?

JanDeDobbeleer:
> Should I rebase to make them all lowercase?

@yehorb I'm going to squash this, don't worry ;-)

PR #1261, opened 2021-11-20
oh-my-posh Atomic Style theme

lfelipe1501:
New Atomic Style Theme by lfelipe1501

### Prerequisites

### Description

I have created my own theme for this framework and I would like to share it with the entire community of this wonderful project.
This theme is based on the original JanDeDobbeleer theme with some slight modifications and following the guide on the official page to create all the segments.

Official capture from my windows terminal

Config1

### DIFF
--- a/themes/atomic.omp.json
+++ b/themes/atomic.omp.json
@@ -38,9 +38,9 @@
           "background": "#FF9248",
           "properties": {
             "home_icon": "",
-            "folder_icon": "\uf413 ",
+            "folder_icon": " \uf413 ",
             "style": "folder",
-            "prefix": "<parentBackground>\uE0B0</> \ufc6e\uf553",
+            "prefix": " \ufc6e\uf553",
             "enable_hyperlink": true
           }
         },
@@ -71,7 +71,7 @@
           "foreground": "#ffffff",
           "background": "#83769c",
           "properties": {
-            "prefix": "<parentBackground>\uE0B0</>羽",
+            "prefix": "羽",
             "postfix": "\u2800",
             "threshold": 0,
             "style": "roundrock"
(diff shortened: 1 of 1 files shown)


themes/atomic.omp.json
```diff
@@ -0,0 +1,326 @@
...
+          "type": "path",
+          "style": "powerline",
+          "powerline_symbol": "\uE0B0",
+          "foreground": "#2d3436",
+          "background": "#FF9248",
+          "properties": {
+            "home_icon": "",
+            "folder_icon": "\uf413 ",
+            "style": "folder",
+            "prefix": "<parentBackground>\uE0B0</> \ufc6e\uf553",
```
lnu: why do you need the `<parentBackground>\uE0B0</>` here? That's the reason for the blue arrow in the path segment. Same for the time segment.

JanDeDobbeleer: That was going to be remark indeed, something is off there (looks pretty col other than that).

JanDeDobbeleer (changes requested):
As indicated by @lnu , changing the two segments will do the trick

themes/atomic.omp.json
lfelipe1501: It's correct, just to style the theme and make it cool, I really left it that way because it seems quite elegant to me.

In this case, is it wrong that I put this?

themes/atomic.omp.json
```diff
@@ -0,0 +1,326 @@
...
+          "properties": {
+            "prefix": "\uFD31 ",
+            "postfix": "",
+            "template": "{{.Context}}{{if .Namespace}} :: {{.Namespace}}{{end}}"
+          }
+        },
+        {
+          "type": "os",
+          "style": "diamond",
+          "leading_diamond": "\uE0B6",
```
JanDeDobbeleer: I would add `"leading_diamond": " \uE0B6",` as without it, you get the previous language segment attached to this one while every other one has a space separation. But other than that 👍🏻

JanDeDobbeleer (changes requested):

themes/atomic.omp.json
lfelipe1501: done

JanDeDobbeleer (approved):

PR #1270, opened 2021-11-21
feature: wifi segment

mikesigs:
### Prerequisites

### Description

This is a WiFi segment to display details of the current WiFi connection. I've only tested it so far on Windows, but the `netsh` command that makes it work there, also works in WSL. So I am pretty sure it will work in WSL too, but obviously need to test.

Any suggestions for how to do that? I was thinking I'd just re-clone the repo in WSL and install another Go environment there, build it, and test.

Also, for the Linux functionality, I tried to use the Dev Container in this project, but after much Googling it turns out virtual environments like that don't actually have a WiFi network interface. The data should be in `/proc/net/wireless`, but as I don't have a physical Linux machine to test with, I cannot develop this functionality. So I'm hoping it's cool to submit this with just Windows and WSL for now. And perhaps a kind contributor could help with the Linux and Darwin code!

Please let me know what you think of the PR. I am very happy to make any changes you request.

Here's a sample of what it looks like with the default template (yes that's my actual SSID):

image

### DIFF
--- a/docs/docs/segment-wifi.md
+++ b/docs/docs/segment-wifi.md
@@ -1,0 +1,47 @@
+---
+id: wifi
+title: WiFi
+sidebar_label: WiFi
+---
+
+## What
+
+Show details about the connected WiFi network.
+
+## Sample Configuration
+
+```json
+{
+  "type": "wifi",
+  "style": "powerline",
+  "background": "#8822ee",
+  "foreground": "#eeeeee",
+  "powerline_symbol": "\uE0B0",
+  "properties": {
+    "template": "{{if .Connectected}}\uFAA8{{else}}\uFAA9{{end}}{{if .Connected}}{{.SSID}} {{.Signal}}% {{.ReceiveRate}}Mbps{{else}}{{.State}}{{end}}"
+  }
+}
+```
+
+## Properties
+
+- connected_icon: `string` - the icon to use when WiFi is connected - defaults to `\uFAA8`
+- connected_icon: `string` - the icon to use when WiFi is disconnected - defaults to `\uFAA9`
+- template: `string` - A go [text/template][go-text-template] template extended with [sprig][sprig]
+utilizing the properties below -
+defaults to `{{if .Connectected}}\uFAA8{{else}}\uFAA9{{end}}{{if .Connected}}{{.SSID}} {{.Signal}}% {{.ReceiveRate}}Mbps{{else}}{{.State}}{{end}}`
+
+## Template Properties
+
+- `.Connected`: `bool` - if WiFi is currently connected
+- `.State`: `string` - WiFi connection status - e.g. connected or disconnected
+- `.SSID`: `string` - the SSID of the current wifi network
+- `.RadioType`: `string` - the radio type - e.g. 802.11ac, 802.11ax, 802.11n, etc.
+- `.Authentication`: `string` - the authentication type - e.g. WPA2-Personal, WPA2-Enterprise, etc.
+- `.Channel`: `int` - the current channel number
+- `.ReceiveRate`: `int` - the receive rate (Mbps)
+- `.TransmitRate`: `int` - the transmit rate (Mbps)
+- `.Signal`: `int` - the signal strength (%)
+
+[go-text-template]: https://golang.org/pkg/text/template/
+[sprig]: https://masterminds.github.io/sprig/
--- a/docs/docs/segment-wifi.md
+++ b/docs/docs/segment-wifi.md
@@ -6,7 +6,11 @@ sidebar_label: WiFi
 
 ## What
 
-Show details about the connected WiFi network.
+Show details about the currently connected WiFi network.
+
+::info
+Currently only supports Windows and WSL. Pull requests for Darwin and Linux support are welcome :)
+:::
 
 ## Sample Configuration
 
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -158,3 +158,5 @@ Output/
 
 # images
 *.png
+
+local/
```
JanDeDobbeleer: I don't think this is related somewhere, right?

src/segment_wifi.go
```diff
@@ -0,0 +1,118 @@
...
+	RadioType      string
+	Authentication string
+	Channel        int
+	ReceiveRate    int
+	TransmitRate   int
+	Signal         int
+}
+
+const (
+	ConnectedIcon    Property = "connected_icon"
```
JanDeDobbeleer: These need to disappear in favor of a template that can handle swapping the icon with a simple if statement. We're trying to get rid of as muych icons as we can bny leveraging templates (you actually already have the logic in `defaultTemplate`)

src/segment_wifi.go
```diff
@@ -0,0 +1,118 @@
...
+		displayError := w.props.getBool(DisplayError, false)
+		if err != nil && displayError {
+			w.State = "WIFI ERR"
+			return true
+		}
+		if err != nil {
+			return false
+		}
+
+		regex := regexp.MustCompile(`(.+) : (.+)`)
```
JanDeDobbeleer: please use the helper functions found in `regex.go`
- [getCompiledRegex](https://github.com/JanDeDobbeleer/oh-my-posh/blob/accf46420ff2924e863f5a76179abc92b2212535/src/regex.go#L15)
- [findNamedRegexMatch](https://github.com/JanDeDobbeleer/oh-my-posh/blob/accf46420ff2924e863f5a76179abc92b2212535/src/regex.go#L35)

src/segment_wifi.go
```diff
@@ -0,0 +1,118 @@
...
+	DisconnectedIcon Property = "disconnected_icon"
+)
+
+const defaultTemplate string = "{{if .Connected}}{{.SSID}} {{.Signal}}% {{.ReceiveRate}}Mbps{{else}}{{.State}}{{end}}"
+
+func (w *wifi) enabled() bool {
+	// If in Linux, who wnows. Gonna need a physical linux machine to develop this since the wlan interface isn't available in wsl
+	// But possible solution is to cat /proc/net/wireless, and if that file doesn't exist then don't show the segment
+	// In Windows and in wsl you can use the command `netsh.exe wlan show interfaces`. Note the .exe is important for this to worw in wsl.
+	if w.env.getPlatform() == windowsPlatform || w.env.isWsl() {
```
JanDeDobbeleer: Invert the if statement to have an early return rather than a full block indented.

src/segment_wifi.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+			case "Signal":
+				if v, err := strconv.Atoi(strings.TrimRight(value, "%")); err == nil {
+					w.Signal = v
+				}
+			}
+		}
+	}
+
+	return true
```
JanDeDobbeleer, line 49: shouldn't this be `false`?

src/segment_wifi_test.go
```diff
@@ -0,0 +1,194 @@
...
+	Transmit rate (Mbps)   : %d
+	Signal                 : %d%%
+	Profile                : ohsiggy
+
+	Hosted network status  : Not available`
+
+	return fmt.Sprintf(netshString, args.state, args.ssid, args.radioType, args.authentication, args.channel, args.receiveRate, args.transmitRate, args.signal)
+}
+
+func bootStrapEnvironment(args *wifiArgs) *wifi {
```
JanDeDobbeleer: We've stepped away from this in favor of using test cases, see [here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/accf46420ff2924e863f5a76179abc92b2212535/src/segment_aws_test.go#L9) for example.

JanDeDobbeleer (changes requested):
Am I correct in thinking this now only works in Windows? If so, please also add that context to the docs, we can always extend Unix (I could at least). Adding a few architectural requests, the test might be the most impactful changes, everything else should be feasible.

.vscode/settings.json
```diff
@@ -15,5 +15,27 @@
   "go.formatTool": "gofmt",
   "go.formatFlags": [
     "-s"
-  ]
+  ],
+  "workbench.colorCustomizations": {
```
JanDeDobbeleer: this needs to be removed, it's a bit too much :-)

.gitignore
mikesigs: I put a local/ folder in the root to contain a theme for testing, then I also outputed the binary there. I can remove it. But how do you handle testing a new segment? Do you just use your jandedobbeleer theme all the time?

.vscode/settings.json
mikesigs: Sorry, this is from the Peacock extension. I will remove it.

src/segment_wifi_test.go
mikesigs: I tried to see how to write these tests using test cases. But I really didn't like the structure of it. Two things I couldn't really grasp:

1. How do you handle mocking dependencies? 
2. How do you handle test cases that don't assert just the simple output of the function?

Too much logic and indirection in testing code makes tests much harder to maintain imo. This is your repository, so I'll check out the link you provided and do my best to restructure.

src/segment_wifi.go
mikesigs: Do you mean, it'd be up to the user to use the template to prefix with their desired icon? And I should just bake this into the default template like `{{if .Connectected}}\uFAA8{{else}}\uFAA9{{end}}`?

.gitignore
JanDeDobbeleer: I have my own config in the home folder. I adjust that one while debugging (and leave the tasks file assumed unchanged at all times).

src/segment_wifi.go
JanDeDobbeleer: Exactly!

src/segment_wifi_test.go
JanDeDobbeleer: I can always help out. Haven't found a downside as it only moves the bootstrap inside the for loop rather than have x functions. It also helps extending use-cases a lot faster.

.vscode/settings.json
JanDeDobbeleer: It was pretty colorful, I admit 😄

src/segment_wifi.go
mikesigs: Cool. This is much cleaner. Got rid of an oddity in the tests I didn't like too.

src/segment_wifi.go
mikesigs: I wound up just removing the regex altogether and using `strings.Split(" : ")` instead

src/segment_wifi.go
mikesigs: If you get past all the other gates in the method (which return false) then at this point it should be true. I refactored this method with your early exit suggestion, as well as pulled out the actual parsing logic into a new method. Makes it much more readable.

src/segment_wifi_test.go
mikesigs: I will take a kick at it. I've pushed up the other changes you requested and suggested. This is the last piece.

mikesigs:
So general question: How should I go about testing this in WSL? So far I've only run the code in Windows, but I have a high level of confidence it'll work in WSL as it's just going to use the `netsh.exe` form of the command, rather than trying to do it the Linux way (wsl itself doesn't have a wlan device). Can I just call `/path/to/my/compiled/oh-my-posh.exe` from wsl, or should I recompile the binary in wsl itself and use that?

mikesigs:
Also, this branch is now out of date with main. Do you prefer I rebase onto main? Or update with a merge commit from main?

src/segment_wifi_test.go
mikesigs: Done. And I see the light :) It's way cleaner.

mikesigs:
This is my first time using Go. It was a fun experience. I think things are pretty much ready now, other than proper WSL testing and the lack of *nix support.

JanDeDobbeleer:
> Do you prefer I rebase onto main

@mikesigs if you can that would be awesome, otherwise I just squash this.

docs/docs/segment-wifi.md
```diff
@@ -0,0 +1,50 @@
...
+id: wifi
+title: WiFi
+sidebar_label: WiFi
+---
+
+## What
+
+Show details about the currently connected WiFi network.
+
+_Note: Currently only supports Windows and WSL. Pull requests for Darwin and Linux support are welcome :)_
```
JanDeDobbeleer: You can turn this into an [admonition](https://docusaurus.io/docs/markdown-features/admonitions):

```markdown
:::info
Note...
:::
```

themes/schema.json
```diff
@@ -1685,6 +1686,38 @@
...
+              "type": { "const": "wifi" }
+            }
+          },
+          "then": {
+            "title": "WiFi Segment",
+            "description": "https://ohmyposh.dev/docs/wifi",
+            "properties": {
+              "properties": {
+                "properties": {
+                  "connected_icon": {
```
JanDeDobbeleer: we still need to remove the icons, right?

JanDeDobbeleer:
2 comments, really great work. For your first attempt at go, I'm super proud of what you did here.

docs/docs/segment-wifi.md
mikesigs: Done

themes/schema.json
mikesigs: and done

mikesigs:
Alrighty. I rebased and squashed most of the commits. Thanks for taking all of this time to review the PR each day. Your quick feedback has been really appreciated! I'm still curious how I should go about testing this in WSL, and if you have any recommendations. Also, I should say thanks for creating this tool. I used oh-my-posh v2 for far far too long. I had a bunch of local customizations I had made to the PowerShell module and was afraid to upgrade. Fortunately, you had basically implemented everything I had done in v2, and the v3 upgrade has been pretty seamless. 

The only thing lacking from my previous customizations was displaying the git branch name in the window/tab title. Perhaps I'll take a crack at that next, if it's something you think would be useful.

JanDeDobbeleer:
> displaying the git branch name in the window/tab title

@mikesigs that's a tricky one. Although we could find a way I guess. We'd need to cache the context.

PR #1273, opened 2021-11-21
feat: Enable kubectl segment to read kubeconfig

hacst:
### Prerequisites

### Description
This extends the kubectl segment to optionally not rely on the
kubectl command and instead to directly parse kubeconfig files like
kubectl does. This is meant as a performance optimization similar
to how the git segment can determine the current branch itself
without calling to git. Especially on Windows and in the presence
other factors slowing process creation like AntiVirus this
can make shells using the segment considerably more responsive.

On my Windows 10 laptop execution time of the segment drops from
around 100ms to 0ms (Write-PoshDebug) noticeably speeding up my
prompt. On another machine with an overly aggressive AV segment
execution goes from seconds to a few milliseconds.
The latter example is admittedly ridiculous but it is what I measure.

The functionality is enabled using the new parse_kubeconfig prop.
It defaults to false to prevent breaking existing users in case
there are any unanticipated behavioral changes.

Additionally the new template properties Cluster and User were
added as they are easily available and helpful in kubectl
setups with more elaborate configuration.

### DIFF
--- a/src/segment_kubectl.go
+++ b/src/segment_kubectl.go
@@ -1,14 +1,34 @@
 package main
 
 import (
+	"path/filepath"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 )
 
+// Whether to use kubectl or read kubeconfig ourselves
+const ParseKubeConfig Property = "parse_kubeconfig"
+
 type kubectl struct {
-	props     properties
-	env       environmentInfo
-	Context   string
-	Namespace string
+	props   properties
+	env     environmentInfo
+	Context string
+	KubeConfigContext
+}
+
+type KubeConfigContext struct {
+	Cluster   string `yaml:"cluster"`
+	User      string `yaml:"user"`
+	Namespace string `yaml:"namespace"`
+}
+
+type KubeConfig struct {
+	CurrentContext string `yaml:"current-context"`
+	Contexts       []struct {
+		Context KubeConfigContext `yaml:"context"`
+		Name    string            `yaml:"name"`
+	} `yaml:"contexts"`
 }
 
 func (k *kubectl) string() string {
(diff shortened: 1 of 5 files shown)


src/segment_kubectl.go
```diff
@@ -31,15 +51,80 @@ func (k *kubectl) init(props *properties, env environmentInfo) {
...
+	// Follow kubectl search rules (see https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable)
+	// TL;DR: KUBECONFIG can contain a list of files. If it's empty ~/.kube/config is used. First file in list wins when merging keys.
+	kubeconfigs := filepath.SplitList(k.env.getenv("KUBECONFIG"))
+	if len(kubeconfigs) == 0 {
+		kubeconfigs = []string{filepath.Join(k.env.homeDir(), ".kube/config")}
+	}
+	contexts := make(map[string]KubeConfigContext)
+	k.Context = ""
+	for _, kubeconfig := range kubeconfigs {
+		if kubeconfig == "" {
```
JanDeDobbeleer: I always default to using `len(kubeconfig) == 0`

src/segment_kubectl.go
```diff
@@ -31,15 +51,80 @@ func (k *kubectl) init(props *properties, env environmentInfo) {
...
+	if displayError {
+		message := "KUBECONFIG ERR"
+		if k.Context == "" {
+			k.Context = message
+		}
+		k.Namespace = message
+		k.User = message
+		k.Cluster = message
+		return true
+	} else {
```
JanDeDobbeleer: the else can be removed due to returning early in the if. I would swap them in that case to remove the need for the indentation:

```golang
if !displayError {
    return false
}
message := "KUBECONFIG ERR"
if k.Context == "" {
	k.Context = message
}
k.Namespace = message
k.User = message
k.Cluster = message
return true
```

src/segment_kubectl.go
```diff
@@ -31,15 +51,80 @@ func (k *kubectl) init(props *properties, env environmentInfo) {
...
 	if !k.env.hasCommand(cmd) {
 		return false
 	}
-	result, err := k.env.runCommand(cmd, "config", "view", "--minify", "--output", "jsonpath={..current-context},{..namespace}")
+	result, err := k.env.runCommand(cmd, "config", "view", "--minify", "--output", "jsonpath={..current-context},{..namespace},{..context.user},{..context.cluster}")
 	displayError := k.props.getBool(DisplayError, false)
 	if err != nil && displayError {
 		k.Context = "KUBECTL ERR"
 		k.Namespace = k.Context
+		k.User = k.Context
```
JanDeDobbeleer: This block and the above statement are identical, it might be useful to have a simple helper function to set the error context?

JanDeDobbeleer (changes requested):
Really good and useful addition, a few small remarks. Great work!

src/segment_kubectl.go
hacst: Personally I prefer `==""` as it's clear I'm checking a string and not an array or anything else. But when in Rome... ;) I replaced all occurances of `== ""` with the preferred pattern.

src/segment_kubectl.go
hacst: Done

src/segment_kubectl.go
hacst: I extracted it into a helper as suggested.

hacst:
I tried to address all the review comments and failing checks. Unfortunately I had to force push to be able to clear the commit message linter error. Sorry about that.

JanDeDobbeleer:
@hacst that's fine, I force push all the time. GitHub should handle that.

src/segment_kubectl_test.go
```diff
@@ -63,15 +85,127 @@ func TestKubectlSegment(t *testing.T) {
...
+			kubectlOutCluster:   tc.Cluster,
+			kubectlErr:          tc.KubectlErr,
 		}
 		kubectl := bootStrapKubectlTest(args)
 		assert.Equal(t, tc.ExpectedEnabled, kubectl.enabled(), tc.Case)
 		assert.Equal(t, tc.ExpectedString, kubectl.string(), tc.Case)
 	}
 }
+
+func TestKubectlSegmentDefaultKubeConfig(t *testing.T) {
```
JanDeDobbeleer: The above was refactord into the test format, but the tests below weren't. Any pointers as to why not?

JanDeDobbeleer:
One question remains

src/segment_kubectl_test.go
hacst: When I initially though about the additional parameters required in the configuration matrix for the parameterization as well as what I wanted to test I felt extending it with all the new parameters (especially file contents) would make it unmanageable. For that reason I chose to add separate tests for the new code path and use the existing pattern for the existing ones. I guess now that it's written down it doesn't look that crazy anymore to extend the parameter list. It could be integrated if that is wanted.

Generally speaking I still feel ambiguous towards having a lot of tests in this pattern. As go test does not natively support parameterized tests I find the workflows around what is basically a long for loop in a single test to be quite clunky. E.g. with a separate test I can easily use my IDE (or console) to selectively run and debug a single test. That is not easily possible for the for loop pattern.

JanDeDobbeleer: @hacst that's a discussion we're not having in this repo as there are too many benefits when it comes to this pattern. If I want to debug a single test, I add that case to the top and we're good to go.

hacst: Fine with me. I will adjust the PR later today.

JanDeDobbeleer: @hacst rereading that I sound like a complete cunt. Wasn't the intention, I love this addition, but when it comes to tests and how they work for me (as I'll be the one here for quite a while) it's an important concept.

hacst: No worries. Different projects have different preferred coding styles. I am just not yet familiar with the preferences in this project. I definitely know the pains of being a maintainer (props for that by the way 👍 ).

hacst: Extended previous test pattern with new tests

hacst:
Addressed test feedback. PTAL.

JanDeDobbeleer (approved):

PR #1288, opened 2021-11-23
Feat: Initial commit of a Nightscout segment to show Blood Sugar

shanselman:
### Prerequisites

### Description

This adds an initial segment to support the display of Nightscout near-realtime blood sugar data for diabetics. This was inspired by the work of @TooTallNate [(blog and PS support here)](https://www.hanselman.com/blog/visualizing-your-realtime-blood-sugar-values-and-a-git-prompt-on-windows-powershell-and-linux-bash) and reimagines it as a segment in OhMyPosh that doesn't require an external service or installation. 

image


docs/docs/segment_nightscout.md
```diff
@@ -0,0 +1,60 @@
...
+```json
+{
+  "type": "nightscout",
+  "style": "diamond",
+  "foreground": "#ffffff",
+  "background": "#ff0000",
+  "background_templates": [
+    "{{ if gt .Sgv 150 }}#FFFF00{{ end }}",
+    "{{ if lt .Sgv 60 }}#FF0000{{ end }}",
+    "#00FF00"
```
JanDeDobbeleer, line 33: this will be more efficient by adding it to the `background` property directly. Saves evaluating one template with the same effect.

docs/docs/segment_nightscout.md
```diff
@@ -0,0 +1,60 @@
...
+    "template": " {{.Sgv}}{{.TrendIcon}}"
+  }
+},
+```
+
+## Properties
+
+- url: `string` - Your Nightscout URL, inclding the full path to entries.json AND count=1 AND token. Example above. You'll know this works if you can curl it yourself and get a single value. - defaults to ``
+- http_timeout: `number` - How long do you want to wait before you want to see your prompt more than your sugar? I figure a half second is a good default - defaults to 500ms
+- template: a go text/template. See the example above where I added a syringe. You can change the icon, put the trend elsewhere, add text, however you like! Make sure your NerdFont has the glyph you want or search for one at nerdfonts.com
```
JanDeDobbeleer: I would add the links to the text template logic like the other segments docs:

```markdown
- template: `string` - a go [text/template][go-text-template] template extended with [sprig][sprig] utilizing the properties below. See the example above where I added a syringe. You can change the icon, put the trend elsewhere, add text, however you like! Make sure your NerdFont has the glyph you want or search for one at nerdfonts.com

[go-text-template]: https://golang.org/pkg/text/template/
[sprig]: https://masterminds.github.io/sprig/
```

docs/docs/segment_nightscout.md
```diff
@@ -0,0 +1,60 @@
+---
+id: nightscout
+title: Nightscout
+sidebar_label: Nightscout
+---
+
+## What
+
+[Nightscout](http://www.nightscout.info/) (CGM in the Cloud) is an open source, DIY project that allows real time access to a CGM data via an HTTP REST API. It is used for secure remote viewing of blood sugar data from anywhere...including OhMyPosh segments on the command line!
```
JanDeDobbeleer: Can't enforce this (yet) via the markdown linter, we have an unwritten convention of adding url's at the bottom of the page:

```markdown
[Nightscout][nightscout] (CGM in the Cloud)

...

[nightscout]: http://www.nightscout.info/
```

docs/docs/segment_nightscout.md
```diff
@@ -0,0 +1,60 @@
...
+- SingleUpIcon - defaults to ↑
+- FortyFiveUpIcon - defaults to ↗
+- FlatIcon - defaults to →
+- FortyFiveDownIcon - defaults to ↘
+- SingleDownIcon - defaults to ↓
+- DoubleDownIcon - defaults to ↓↓
+
+## Template Properties
+
+- .Sgv - Your Serum Glucose Value (your sugar)
```
JanDeDobbeleer: we need to add the types so people know what they can expect

src/segment_nightscout.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err.Error()
+	}
+
+	return text
+}
+
+func (ns *nightscout) getResult() (*nightscoutData, error) {
+	url := ns.props.getString(URL, "")
+	//cacheTimeout := int64(ns.props.getInt(CacheTimeout, DefaultCacheTimeout))
```
JanDeDobbeleer: just in case you forget: remove the hardcoded value ;-)
about: //nightscout can only give you data every 5 min, todo: make it configurable?

If you do this, that's what will happen, timeout is in minutes:

`cacheTimeout := ns.props.getInt(CacheTimeout, 5)`

src/segment_nightscout.go
```diff
@@ -0,0 +1,131 @@
...
+		return &nightscoutData{}, err
+	}
+	var arr []*nightscoutData
+	err = json.Unmarshal(body, &arr)
+	if err != nil {
+		return &nightscoutData{}, err
+	}
+
+	firstelement := arr[0]
+	firstData, err := json.Marshal(firstelement)
```
JanDeDobbeleer, line 119: to avoid doing this, we're better off storing the entire response (`body`). Only means we need a smell helper function (can be scoped inside this function) to parse the response and return the first entity.

JanDeDobbeleer (changes requested):
A few pointers, But overall great work. Feels like we went through this already :-) (now get some sleep 😴)

docs/docs/segment_nightscout.md
shanselman: I'm confused, we want folks to have the ability to change colors, and this is three colors. What would we add to the background property? Having the two offers a lot of flexibility. background template is likely to be the only one used, and foreground is likely to be hard coded by the user to white or black.

docs/docs/segment_nightscout.md
shanselman: cool

src/segment_nightscout.go
shanselman: I looked at this, and tried a few different ways and I'm not seeing how it adds a lot (makes it harder to read for me) and it only runs every 5 min.

docs/docs/segment_nightscout.md
shanselman: Here what it looks like at the moment:

```
{
    "type": "nightscout",
    "style": "powerline",
    "powerline_symbol": "",
    "invert_powerline": false,
    "foreground": "#000000",
    "background": "#ff0000",
    "background_templates": [
      "{{ if gt .Sgv 150 }}#FFFF00{{ end }}",
      "{{ if lt .Sgv 60 }}#FF0000{{ end }}",
      "#00FF00"
    ],          
    "leading_diamond": "",
    "trailing_diamond": "",
    "properties": {
      "url": "MINE",
      "http_timeout": 2000,
      "template": " {{.Sgv}}{{.TrendIcon}}"
    }
  },
```

image

JanDeDobbeleer: good question, using the following will result in the `background` never being used. It evaluates the templates sequentially and in case of no match it will always return `#00FF00` and never `#ff0000`. Adding `#ff0000` and removing it from the templates has the same result as `background` is the always the fallback color. But I can merge it like that.

JanDeDobbeleer (approved):

JanDeDobbeleer:
@All-contributors please add @shanselman for code and docs

PR #1260, opened 2021-11-19
feat(segment_regquery): segment to query and display a registry key on Windows.

memcpy-rand-rand-rand:
### Prerequisites

- [X] I have read and understood the `CONTRIBUTING` guide
- [X] The commit message follows the [conventional commits][cc] guidelines
- [X] Tests for the changes have been added (for bug fixes/features)
- [X] Docs have been added / updated (for bug fixes/features)

### Description

New segment, tests and docs which will allow the user to specify a Windows registry path and key and display the result (as the string returned if registry key type is REG_SZ or as an 0x%08X capital hex number for type REG_DWORD).  Ability to specify different behaviours for failure conditions (do not display, use supplied fallback string or show debug info).

build: tags used by OS to supply bypass implementation for non-Windows.

Also, in this PR - minor fix to the segment contribution doc

Hadn't used CC msgs while playing about in the fork, most recent commits updated to do so.

[cc]: https://www.conventionalcommits.org/en/v1.0.0/#summary

PR #1297, opened 2021-11-24
feat(windows): registry query segment

JanDeDobbeleer:
### Prerequisites

### Description

closes #1260

### DIFF
--- a/docs/docs/segment-winreg.md
+++ b/docs/docs/segment-winreg.md
@@ -1,0 +1,46 @@
+---
+id: winreg
+title: Windows Registry Key Query
+sidebar_label: Windows Registry Key Query
+---
+
+## What
+
+Display the content of the requested Windows registry key.
+
+Supported registry key types:
+
+- String
+- DWORD (displayed in upper-case 0x hex)
+
+## Sample Configuration
+
+```json
+  {
+    "type": "winreg",
+    "style": "powerline",
+    "powerline_symbol": "\uE0B0",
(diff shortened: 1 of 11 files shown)


memcpy-rand-rand-rand:
@JanDeDobbeleer, thanks for the edits.  Still wrangling with merging myself, but noticed that the template call moved out of enabled()?  I believe this breaks the "if I fail to get the value from the registry display this instead" template scenario.

JanDeDobbeleer:
@memcpy-rand-rand-rand ah yes, the fallback value. That was necessary in your case, right? Let me add that back! That means that we don't error but return a value of that's filled in.

JanDeDobbeleer:
@memcpy-rand-rand-rand incoming via #1306

memcpy-rand-rand-rand:
Actually, I had changed the logic away from the fallback value using the template.

It evaluated the template in IsEnabled() such that if the user had configured something in the config json to be displayed when '. Value' is null, the string to use for the segment would be non-empty and therefore segment enabled.  If no template or result of the template is empty then isenabled returned false.

JanDeDobbeleer:
@memcpy-rand-rand-rand gotcha, looking at the case though it makes it easier to have that value directly and do some of the lifting for the end user. If that's OK with you.

memcpy-rand-rand-rand:
sure thing, I like the explicit approach.  will have a play when I finally get my merging sorted out :)

PR #1324, opened 2021-11-26
chore(devcontainer): fix path

lnu:
### Prerequisites

### Description

the path is wrong(at least does not work here). With the fix the prompt is ok.


JanDeDobbeleer:
@lnu now I'm puzzled as to why this worked in my container...it shouldn't have...

lnu:
Probably you had the executable in your path from a previous build? For testing the best is to do a rebuild without cache to start with a fresh image/container.

JanDeDobbeleer:
I did that. So I don't get it. Either way, this fix is correct.

PR #1326, opened 2021-11-26
theme: add festivetech

JanDeDobbeleer:
### Prerequisites

### Description

this theme is a bit special and needs the following in your
PowerShell profile to work correctly:

function Set-EnvVars {
    $now = [DateTime]::Now
    $env:DAYS = ([Datetime](“12/25/” + $now.Year) – $Now).Days
}
New-Alias -Name 'Set-PoshContext' -Value 'Set-EnvVars' -Scope Global -Force

happy holidays!

closes #1236

### DIFF
--- a/themes/festivetech.omp.json
+++ b/themes/festivetech.omp.json
@@ -1,0 +1,94 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "final_space": true,
+  "blocks": [
+    {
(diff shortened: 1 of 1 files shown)


BoseSj:
Is it already added to the documentation? If not, shouldn't we add this to then? As that will be more convenient for the users.

JanDeDobbeleer:
@BoseSj there's no way to add this anywhere. I'll remove the theme as it was for a specific event a while back.

BoseSj:
Okay, cool!

PR #1338, opened 2021-11-27
pwsh module basic wrapper

JanDeDobbeleer:
### Prerequisites

### Description

There's quite a bit of confusion around the module vs the binary etc. This can all be fixed if we fallback to
using the module as an alternative download method rather than a full blown setup and align everything else.


JanDeDobbeleer:
@lnu thoughts? This will annoy quite a bit of people but once that's done it should make our lives easier? I still need to validate the PATH adding on a clean machine.

JanDeDobbeleer:
> @lnu thoughts? This will annoy quite a bit of people but once that's done it should make our lives easier? I still need to validate the PATH adding on a clean machine.

I was in doubt of we should keep the `Set-PoshPrompt` function as is or not. Maybe it's better to keep it, but don't document it anymore. That way we're not annoying anyone and we can move forward.

yehorb:
@JanDeDobbeleer why are you removing Get-PoshThemes from PowerShell module? It is a great way to see how themes will render in my current context. I used it quite a lot in the beginning, to see how different themes look at different directory depth, conda environments etc.

JanDeDobbeleer:
@yehorb it's a draft, I'm going to add it back in, but it needs to know where to look and move outside of the module.

lnu:
I always use the powershell commands. It's way easier to list all themes or call debug print. But on the other side it brings lots of troubles(issues, discussions, ...).  
But pwsh module install is really specific and not really "cross platform". So, maybe we could have standard install for all platforms and an extra section for people who wants to use/keep using pwsh(at your own risk :)).

JanDeDobbeleer:
@lnu the change I'm working on here will keep all module "advantages " while also stating the disadvantage (add to PATH manually for other shells). So I believe this is the best of everything and allows to align the docs (and reduce overhead). Let me tweak it a little bit more and we're good to go.

PR #1344, opened 2021-11-29
Jdd main

memcpy-rand-rand-rand:
### Prerequisites

- [X] Docs have been added / updated (for bug fixes/features)

### Description

feat(winreg): update return logic to allow successfully retrieved empty value to be used without fallback

fixing my commit tree with rebase and squash

fix schema.json winreg entries

undo erroneous prior fix to the segment contrib doc 


src/segment_winreg.go
```diff
@@ -32,11 +32,17 @@ func (wr *winreg) enabled() bool {
 
 	var err error
 	wr.Value, err = wr.env.getWindowsRegistryKeyValue(registryPath, registryKey)
-	if len(fallback) > 0 && (err != nil || len(wr.Value) == 0) {
-		wr.Value = fallback
-		return true
+
+	if err != nil {
```
JanDeDobbeleer: Can you do an early return here? `if err == nil`, then we can continue the logic in this block unindented below.

JanDeDobbeleer (changes requested):

src/segment_winreg.go
memcpy-rand-rand-rand: Gotcha, like this:

```
	if err == nil {
		return true
	}

	if len(fallback) > 0 {
		wr.Value = fallback
		return true
	}

	return false

```

Feels like there's almost a razor here - i.e. if you find yourself writing "else" to go through codeblocks that contain return, look again you can likely optimize the early-out structure.

JanDeDobbeleer: Exactly. I also didn't invent this strategy, it comes from [object calisthenics](https://gist.github.com/bobuss/6534934). While not every rule is always applicable or relevant, it still produces very readable code when you keep it in mind.

memcpy-rand-rand-rand: Interesting, thanks.  I was always a "single return path" guy when I was engineering full-time, it's fun to challenge that!

JanDeDobbeleer:
@memcpy-rand-rand-rand I'm so proud of how you've grown here in just two PR's. Really well done 👍🏻

PR #1377, opened 2021-12-04
Env reg variant

memcpy-rand-rand-rand:
### Prerequisites

- [] Docs have been added / updated (for bug fixes/features)

### Description

refactor(winreg): change getWindowsRegistryKeyValue fn to be more generic, simplify winreg path


memcpy-rand-rand-rand:
Draft PR for discussion.  Added variant-style code to the getWindowsRegistryKeyValue function to make it more useful should others with to use it.

src/environment.go
```diff
@@ -56,6 +56,21 @@ type cache interface {
 	set(key, value string, ttl int)
 }
 
+type windowsRegistryValueType int
+
+const (
```
JanDeDobbeleer, line 61: should we move these to `environment_windows` so they don't get added for other platforms?

src/environment_windows.go
```diff
@@ -105,31 +105,57 @@ func (env *environment) getCachePath() string {
...
+	//
+	// 1. Root key - extract the root HKEY string and turn this into a handle to get started
+	// 2. Path - open this path
+	// 3. Key - get this key value
+	//
+	// If 3 is "" (i.e. the path ends with "\"), then get (Default) key.
+	//
+	regPathParts := strings.SplitN(path, "\\", 2)
 
 	regRootHKeyHandle := getHKEYHandleFromAbbrString(regPathParts[0])
```
JanDeDobbeleer, line 140: I didn't catch this at first, it assumes there's an element in the array.

src/environment_windows.go
```diff
@@ -105,31 +105,57 @@ func (env *environment) getCachePath() string {
...
 	if regRootHKeyHandle == 0 {
 		errorLogMsg := fmt.Sprintf("Error, Supplied root HKEY value not valid: '%s'", regPathParts[0])
 		env.log(Error, "getWindowsRegistryKeyValue", errorLogMsg)
-		return "", errors.New(errorLogMsg)
+		return nil, errors.New(errorLogMsg)
 	}
 
+	// Strip key off the end.
+	lastSlash := strings.LastIndex(regPathParts[1], "\\")
```
JanDeDobbeleer, line 148: same here, we expect as second element, I would validate if we have all entries before continuing.

src/environment_windows.go
```diff
@@ -105,31 +105,57 @@ func (env *environment) getCachePath() string {
...
+
+	regKey := regPathParts[1][lastSlash+1:]
+	regPath := regPathParts[1][0:lastSlash]
+
+	regKeyLogged := regKey
+	if regKeyLogged == "" {
```
JanDeDobbeleer: purely style, I always use `len(regKeyLogged) == 0`

src/environment_windows.go
```diff
@@ -105,31 +105,57 @@ func (env *environment) getCachePath() string {
...
+		regKeyLogged = "(Default)"
+	}
+	env.log(Error, "getWindowsRegistryKeyValue", fmt.Sprintf("getWindowsRegistryKeyValue: root:\"%s\", path:\"%s\", key:\"%s\"", regPathParts[0], regPath, regKeyLogged))
+
 	// Second part of split is registry path after HK part - needs to be UTF16 to pass to the windows. API
-	regPathUTF16, regPathUTF16ConversionErr := windows.UTF16FromString(regPathParts[1])
+	regPathUTF16, regPathUTF16ConversionErr := windows.UTF16FromString(regPath)
```
JanDeDobbeleer: pls rename to `err` for consistency, you can reuse it.

src/environment_windows.go
```diff
@@ -105,31 +105,57 @@ func (env *environment) getCachePath() string {
...
+	lastSlash := strings.LastIndex(regPathParts[1], "\\")
+
+	regKey := regPathParts[1][lastSlash+1:]
+	regPath := regPathParts[1][0:lastSlash]
+
+	regKeyLogged := regKey
+	if regKeyLogged == "" {
+		regKeyLogged = "(Default)"
+	}
+	env.log(Error, "getWindowsRegistryKeyValue", fmt.Sprintf("getWindowsRegistryKeyValue: root:\"%s\", path:\"%s\", key:\"%s\"", regPathParts[0], regPath, regKeyLogged))
```
JanDeDobbeleer: why is this line here? Seems to always log that.

src/segment_winreg.go
```diff
@@ -62,3 +74,45 @@ func (wr *winreg) templateString(segmentTemplate string) string {
 	}
 	return text
 }
+
+func (wr winreg) GetRegistryString(path string) (string, error) {
```
JanDeDobbeleer, line 78: Oh boy do I want generics in go ASAP...

src/segment_winreg.go
```diff
@@ -62,3 +74,45 @@ func (wr *winreg) templateString(segmentTemplate string) string {
...
+	regValue, err := wr.env.getWindowsRegistryKeyValue(path)
+
+	if regValue != nil {
```
JanDeDobbeleer: Inverting this if allows the send if to be on the same level below.

JanDeDobbeleer (changes requested):
A few remarks, nothing major. Pretty cool stuff tbh.

src/environment.go
memcpy-rand-rand-rand: I was wondering about that too - if possible to have the same interface, but on windows the windowsRegistryValueType and interface etc is full featured where on other platforms it's an interface with no info... I'll have a try.

src/environment_windows.go
memcpy-rand-rand-rand: yes - I'm making the assumption that the split works - will add some validation.

src/environment_windows.go
memcpy-rand-rand-rand: yup - validation needed.

src/environment_windows.go
memcpy-rand-rand-rand: yes - mostly for debug implementation help.  Likely end-users wouldn't need it I guess.

src/segment_winreg.go
memcpy-rand-rand-rand: ooh, good spot - yeah, makes it a lot cleaner 👍

src/environment_windows.go
memcpy-rand-rand-rand: ah whoops - I think I see what you mean, it's always logging an error not debug.

src/environment.go
memcpy-rand-rand-rand: ah - problem with that, segment_winreg.go has a dependency on that interface and enum now, so to make it windows-only at an environment level means the segment would then have to have different platform impls to avoid using the interface/enum.

JanDeDobbeleer: Gotcha!

memcpy-rand-rand-rand:
@JanDeDobbeleer I branched this and squashed commits to remove the commit lint issue; it has the changes as reviewed in this PR.  Would you prefer I create a PR on the new branch?

JanDeDobbeleer:
@memcpy-rand-rand-rand works for me.

PR #1385, opened 2021-12-07
feat(language): parameter for version detection

lnu:
### Prerequisites

read angular version from angular project file. New parameter added for language  for custom version reading.

### DIFF
--- a/src/segment_language.go
+++ b/src/segment_language.go
@@ -10,6 +10,7 @@ type loadContext func()
 
 type inContext func() bool
 
+type getVersion func() (string, error)
 type matchesVersionFile func() bool
 
 type version struct {
@@ -25,6 +26,7 @@ type cmd struct {
 	executable string
 	args       []string
 	regex      string
+	getVersion getVersion
 }
 
 func (c *cmd) parse(versionInfo string) (*version, error) {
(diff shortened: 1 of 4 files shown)


src/segment_language.go
```diff
@@ -166,13 +168,22 @@ func (l *language) hasLanguageFiles() bool {
...
+		if command.getVersion == nil {
+			if !l.env.hasCommand(command.executable) {
+				continue
+			}
+			versionStr, err = l.env.runCommand(command.executable, command.args...)
+			if exitErr, ok := err.(*commandError); ok {
+				l.exitCode = exitErr.exitCode
+				return fmt.Errorf("err executing %s with %s", command.executable, command.args)
+			}
+		} else {
```
JanDeDobbeleer, line 182: if we swap the if, we can simply return early or continue in case we have an error

JanDeDobbeleer (changes requested):
One remark

JanDeDobbeleer:
@lnu question, do we ever see the need for additional logic to add that to cmd but rather have the function live on language level and only execute it once?

lnu:
> @lnu question, do we ever see the need for additional logic to add that to cmd but rather have the function live on language level and only execute it once?

I put it at the command level because we still use the regex parsing part of the command.

JanDeDobbeleer:
@lnu fair point, can you have a look at my comment? Then we can merge this in.

src/segment_language.go
lnu: I don't really get it. If `getVersion` is not null we use it to read the version, if not we use the command. Both will output a `versionStr` that needs to be parsed using the regex. I probably missed your point, sorry.
image

JanDeDobbeleer: you're right.

PR #1392, opened 2021-12-08
feat(plastic): added Plastic SCM segment

Khaos66:
### Prerequisites

### Description

This adds [Plastic SCM](https://www.plasticscm.com/) as a new `plastic` segment. It took inspiration from the `git` segment. 

### Todo

These are points to work on:

- [x] rename `g` to `p`
- [x] remove `StatusColorsEnabled` check
- [x] remove `getCmCommand`
- [x] use `findNamedRegexMatch` more defensively
- [x] refactor `truncateBranch` into new `scm` base type
- [x] refactor `gitStatus` to `scmStatus`
   - [x] switch to new `add` loop logic
- [x] fix `unmerged` counter to reflect files with unhandled merge conflicts
- [x] create new page in `contribute section` with information about setup and manual test of Plastic SCM

### DIFF
--- a/src/segment_plastic.go
+++ b/src/segment_plastic.go
@@ -6,27 +6,37 @@ import (
 	"strings"
 )
 
+type PlasticStatus struct {
+	ScmStatus
+}
+
+func (s *PlasticStatus) add(code string) {
+	switch code {
+	case "LD":
+		s.Deleted++
+	case "AD", "PR":
+		s.Added++
+	case "LM":
+		s.Moved++
+	case "CH":
+		s.Modified++
+	}
+}
+
 type plastic struct {
-	props properties
-	env   environmentInfo
+	scm
 
-	Moved    int
-	Deleted  int
-	Added    int
-	Modified int
-	Unmerged int
-	Changed  bool
-
+	Status   *PlasticStatus
 	Behind   bool
 	Selector string
 
 	plasticWorkspaceFolder string // root folder of workspace
 }
 
-const (
-	// FullBranchPath displays the full path of a branch
-	FullBranchPath Property = "full_branch_path"
-)
+func (p *plastic) init(props properties, env environmentInfo) {
+	p.props = props
+	p.env = env
+}
 
 func (p *plastic) enabled() bool {
 	if !p.env.hasCommand("cm") {
@@ -48,21 +58,12 @@ func (p *plastic) enabled() bool {
 	return false
 }
 
-func (p *plastic) shouldIgnoreRootRepository(rootDir string) bool {
-	value, ok := p.props[ExcludeFolders]
-	if !ok {
-		return false
-	}
-	excludedFolders := parseStringArray(value)
-	return dirMatchesOneOf(p.env, rootDir, excludedFolders)
-}
-
 func (p *plastic) string() string {
 	displayStatus := p.props.getOneOfBool(FetchStatus, DisplayStatus, false)
 
-	p.Selector = p.getSelector()
+	p.setSelector()
 	if displayStatus {
-		p.getPlasticStatus()
+		p.setPlasticStatus()
 	}
 
 	// use template if available
@@ -88,23 +89,7 @@ func (p *plastic) templateString(segmentTemplate string) string {
 	return text
 }
 
-func (p *plastic) Status() string {
-	var status string
-	stringIfValue := func(value int, prefix string) string {
-		if value > 0 {
-			return fmt.Sprintf(" %s%d", prefix, value)
-		}
-		return ""
-	}
-	status += stringIfValue(p.Added, "+")
-	status += stringIfValue(p.Modified, "~")
-	status += stringIfValue(p.Deleted, "-")
-	status += stringIfValue(p.Moved, ">")
-	status += stringIfValue(p.Unmerged, "x")
-	return strings.TrimSpace(status)
-}
-
-func (p *plastic) getPlasticStatus() {
+func (p *plastic) setPlasticStatus() {
 	output := p.getCmCommandOutput("status", "--all", "--machinereadable")
 	splittedOutput := strings.Split(output, "\n")
 	// compare to head
@@ -113,6 +98,7 @@ func (p *plastic) getPlasticStatus() {
 	p.Behind = headChangeset > currentChangeset
 
 	// parse file state
+	p.Status = &PlasticStatus{}
 	p.parseFilesStatus(splittedOutput)
 }
 
@@ -124,23 +110,9 @@ func (p *plastic) parseFilesStatus(output []string) {
 		if len(line) < 3 {
 			continue
 		}
-		if matchString(`(?i)merge\s+from\s+[0-9]+\s*$`, line) {
-			p.Unmerged++
-			continue
-		}
-		code := line[0:2]
-		switch code {
-		case "LD":
-			p.Deleted++
-		case "AD", "PR":
-			p.Added++
-		case "LM":
-			p.Moved++
-		case "CH":
-			p.Modified++
-		}
+		code := line[:2]
+		p.Status.add(code)
 	}
-	p.Changed = p.Added > 0 || p.Deleted > 0 || p.Modified > 0 || p.Moved > 0 || p.Unmerged > 0
 }
 
 func (p *plastic) parseStringPattern(output, pattern, name string) string {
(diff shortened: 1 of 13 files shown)


Khaos66:
Hi! Please let me know if I missed anything. I'll be happy to contribute =)

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+
+	cmCommand string
+}
+
+const (
+	// FullBranchPath displays the full path of a branch
+	FullBranchPath Property = "full_branch_path"
+)
+
+func (g *plastic) enabled() bool {
```
JanDeDobbeleer: Can you rename `g` to `p` ;-)

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+	value, ok := g.props[ExcludeFolders]
+	if !ok {
+		return false
+	}
+	excludedFolders := parseStringArray(value)
+	return dirMatchesOneOf(g.env, rootDir, excludedFolders)
+}
+
+func (g *plastic) string() string {
+	statusColorsEnabled := g.props.getBool(StatusColorsEnabled, false)
```
JanDeDobbeleer: This is old and needs to be removed, it's only in git for backwards compatibility. We now color the segment based on the color templates.

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+		Env:      g.env,
+	}
+	text, err := template.render()
+	if err != nil {
+		return err.Error()
+	}
+	return text
+}
+
+func (g *plastic) Status() string {
```
JanDeDobbeleer: This is still based on the previous version of the git segment, it would be interesting to see if the `gitStatus` struct could be reused (and renamed obviously in that case).

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+	// compare to head
+	currentChangeset := g.parseStatusChangeset(splittedOutput[0])
+	headChangeset := g.getHeadChangeset()
+	g.Behind = headChangeset > currentChangeset
+
+	// parse file state
+	g.parseFilesStatus(splittedOutput)
+}
+
+func (g *plastic) parseFilesStatus(output []string) {
```
JanDeDobbeleer: Same remark as for `gitStatus`, we can make use (extend) the `add` function there

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+		case "CH":
+			g.Modified++
+		}
+	}
+	g.Changed = g.Added > 0 || g.Deleted > 0 || g.Modified > 0 || g.Moved > 0 || g.Unmerged > 0
+}
+
+func (g *plastic) parseStatusChangeset(status string) int {
+	var csRegex = `STATUS\s+(?P<cs>[0-9]+?)\s`
+	cs, _ := strconv.Atoi(findNamedRegexMatch(csRegex, status)["cs"])
```
JanDeDobbeleer: what if there's no match?

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+	return cs
+}
+
+func (g *plastic) getHeadChangeset() int {
+	output := g.getCmCommandOutput("status", "--head", "--machinereadable")
+	var csRegex = `\bcs:(?P<cs>[0-9]+?)\s`
+	cs, _ := strconv.Atoi(findNamedRegexMatch(csRegex, output)["cs"])
```
JanDeDobbeleer: same remark as above, no match can crash this

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+	symbol := g.props.getString(TruncateSymbol, "")
+	return branch[0:maxLength] + symbol
+}
+
+func (g *plastic) getCmCommandOutput(args ...string) string {
+	val, _ := g.env.runCommand(g.getCmCommand(), args...)
+	return val
+}
+
+func (g *plastic) getCmCommand() string {
```
JanDeDobbeleer: This is a duplicate function, if we feel like we need this maybe it's better to add this logic to `runCommand` IF this is an issue for all executables (git is slow in this case so we fallback to the Windows variant)

src/segment_plastic.go
```diff
@@ -0,0 +1,245 @@
...
+	var branchRegex = `branch "(?P<branch>[\/a-zA-Z0-9\-\_]+?)"`
+	return findNamedRegexMatch(branchRegex, selector)["branch"]
+}
+
+func (g *plastic) init(props properties, env environmentInfo) {
+	g.props = props
+	g.env = env
+}
+
+func (g *plastic) truncateBranch(branch string) string {
```
JanDeDobbeleer: Duplicate function

JanDeDobbeleer (changes requested):
All in all, really good addition. Some duplications and old style (I admit, I did refactor the git segment like a madman the last few days). About `findNamedRegexMatch`, that's being used a bit too freely and is subject to break if there are no matches.

Khaos66:
Wow, thank you for the fast review!
I agree with all you points. I see your changes to `git` in #1389 and will apply them to this segment as well.
It may (or may not) take some time, as I'm doing this in my free time. Cheers

src/segment_plastic.go
Khaos66: Yes, I was wondering what it's used for and forgot to remove it. After all, it's not really used anyway. 
Will be removed!

src/segment_plastic.go
Khaos66: I agree. It could be renamed to `scmStatus`. What do you think?
One difference is that I counted `moved` files separatly and in `git` they are counted as `modified`.

src/segment_plastic.go
Khaos66: I'll have a look at the new `add` logic.
By the way, what does `unmerged` translate to in git? Is it 'files with unhandled merge conflicts' or 'locally merged but uncommited files'? (I implemented the later one for plastic)

src/segment_plastic.go
Khaos66: Oh. So this might not be needed at all? 
`cm.exe` (windows) and `cm` (others) can both be executed via `cm`. So maybe you'r right. I will test that.

src/segment_plastic.go
Khaos66: True. Where would you refactor this function to? (I need a little guidance here)

Khaos66:
Also, is there a place where I could document steps for manual testing? I guess it would be helpfull to document the steps to setup a Plastic SCM server and client with repos and workspaces (as this is not common knowlege).
Maybe in a github issue?

src/segment_plastic.go
JanDeDobbeleer: the first option

src/segment_plastic.go
JanDeDobbeleer: Good question, we could create an `scm` struct that has the functions (also status) and use struct inheritance in `git` and `plastic`

JanDeDobbeleer:
> Also, is there a place where I could document steps for manual testing? I guess it would be helpfull to document the steps to setup a Plastic SCM server and client with repos and workspaces (as this is not common knowlege).
Maybe in a github issue?

@Khaos66 I would create a separate page in the[ contributing section](https://ohmyposh.dev/docs/contributing_started)

src/segment_plastic.go
Khaos66: This worked out well =) See [edb0fec](https://github.com/JanDeDobbeleer/oh-my-posh/pull/1392/commits/edb0fece458381486862e9e400feb375388231f9)

Khaos66:
I'm trying to get the icon for Plastic SCM added to FontAwesome and ultimately to Nerd-Fonts.
Please help me up-vote the issue
https://github.com/FortAwesome/Font-Awesome/issues/18504

Khaos66:
@JanDeDobbeleer I think I'm done. Can you please review again?

src/segment_plastic.go
```diff
@@ -0,0 +1,192 @@
...
+	output := p.getCmCommandOutput("status", "--head", "--machinereadable")
+	return p.parseIntPattern(output, `\bcs:(?P<cs>[0-9]+?)\s`, "cs", 0)
+}
+
+func (p *plastic) setSelector() {
+	var ref string
+	selector := p.getFileContents(p.plasticWorkspaceFolder+"/.plastic/", "plastic.selector")
+	// changeset
+	ref = p.parseChangesetSelector(selector)
+	if ref != "" {
```
JanDeDobbeleer: if `len(ref) == 0` (same as everywhere else)

src/segment_plastic.go
```diff
@@ -0,0 +1,192 @@
...
+		p.Selector = fmt.Sprintf("%s%s", p.props.getString(CommitIcon, "\uF417"), ref)
+		return
+	}
+	// fallback to label
+	ref = p.parseLabelSelector(selector)
+	if ref != "" {
```
JanDeDobbeleer: if `len(ref) == 0` (same as everywhere else)

src/segment_plastic.go
```diff
@@ -0,0 +1,192 @@
...
+		p.Selector = fmt.Sprintf("%s%s", p.props.getString(TagIcon, "\uF412"), ref)
+		return
+	}
+	// fallback to branch/smartbranch
+	ref = p.parseBranchSelector(selector)
+	if ref != "" {
```
JanDeDobbeleer: if `len(ref) == 0` (same as everywhere else)

src/segment_plastic.go
```diff
@@ -0,0 +1,192 @@
...
+	match := findNamedRegexMatch(pattern, output)
+	if sValue, ok := match[name]; ok {
+		return sValue
+	}
+	return ""
+}
+
+func (p *plastic) parseIntPattern(output, pattern, name string, defValue int) int {
+	sValue := p.parseStringPattern(output, pattern, name)
+	if sValue != "" {
```
JanDeDobbeleer: if `len(sValue) == 0` (same as everywhere else)

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@Khaos66 only 1 styling comment, amazing PR!

Khaos66:
> @Khaos66 only 1 styling comment, amazing PR!

@JanDeDobbeleer Thank you very much! I'm doing what I can =D (first time in `go` though...)

It's done btw ;)

Khaos66:
<3

Issue #885, opened 2021-08-06
Ability to add text decoration to segment contents


vidhanio:
### Prerequisites

- [x] I have read and understood the `CONTRIBUTING` guide
- [x] I have looked for duplicate issues before submitting this one

### Description

Is it possible to apply text decorations (e.g. italics, bold) to the actual content of the segments? It seems to only be possible to add them to the pre/postfix and was wondering if it was possible to apply to the content.

### Environment

- Oh My Posh version: 3.171.2
- Theme: Custom
- Operating System: Windows 10
- Shell: Powershell
- Terminal: Windows Terminal

JanDeDobbeleer:
@vidhanlol segments which have a `template` can be text decorated (see [kubectl](https://ohmyposh.dev/docs/kubectl) for example). Not every segment has this yet, something I wanted to do for V4 as it could break existing themes (so we need an upgrade functionality).

vidhanio:
Interesting, I'll look into that. Thanks!

JanDeDobbeleer:
@vidhanlol for which segment would you like this? Because we can extend without breaking for some.

vidhanio:
Hey! Sorry, wasn't checking my notifications often. My current custom theme is using `git`, `python`, `session`, `time`, `path`, and `spotify`.

tylerbutler:
For `path`, I took a quick look at the code and I think it would be safe to add a `template` property that, when set, overrides the other format related properties. This is how `template` behaves in most (all?) cases today.

This may be true for other segments; I didn't look.

@JanDeDobbeleer is my reading of the code correct? Do you foresee any compatibility issues with this change for `path`? I am interested in opening a PR if you're open to the change.

JanDeDobbeleer:
@tylerbutler all segments would need to be adjusted so that templating works. So yes, we should do that 😅 There's a file called `segment_deprecated.go` where we've added properties which we silently support for backwards compatibility in case the usage of template would render those obsolete, not sure if it applies here but just in case.

JanDeDobbeleer:
All segments now have templating capabilities.

bckelley:
@JanDeDobbeleer  I'm trying to do this too, using a customized Nordtron just can't figure out how to make it ```<b>bold</b>``` I did see this in the docs [themes docs](https://ohmyposh.dev/docs/configuration/templates#text-decoration) 
small snippet
```JSON
{
  "foreground": "#FF0000",
  "style": "plain",
[...]

PR #1404, opened 2021-12-10
Path segment: add support for template property

tylerbutler:
_I'm opening this as a draft to verify the approach and get feedback before I go too far._

### Prerequisites

Not all of these are done, but I'll complete them iteratively in the PR.

### Description

This PR adds support for a `template` property in the Path segment. See #885 for more context.

#### Template properties

The template receives the following properties:

These four properties contain the string values of the path with various styles:

- Full
- Folder
- Letter
- Mixed

The `style` property is ignored. Instead, the strings passed to the template take into account the other properties that affect how the styles are rendered. For example, this path configuration produces the following:

```json
{
  "type": "path",
  "style": "plain",
  "foreground": "#ffffff",
  "background": "#ff479c",
  "properties": {
    "style": "agnoster",
    "folder_separator_icon": ">",
    "template": "<b>{{ .Full }} {{ .Folder }} {{ .Letter }} {{ .Mixed }}</b>"
  }
}
```

image

These four properties contain icons (as strings):

- FolderSeparatorIcon
- HomeIcon
- FolderIcon
- WindowsRegistryIcon

#### Behavior notes

When the template property is provided, the `style` property is ignored since the strings for all styles are passed to the template. This is consistent with other segments that support a template property. It should also be backwards-compatible. Existing themes won't have a template property and thus will continue to work as they do today. Users can opt into the template by adding it to their theme.

### DIFF
--- a/src/segment_path.go
+++ b/src/segment_path.go
@@ -10,6 +10,17 @@ import (
 type path struct {
 	props properties
 	env   environmentInfo
+
+	// Full displays the full path
+	Full   string
+	Folder string
+	Letter string
+	Mixed  string
+
+	FolderSeparatorIcon string
+	HomeIcon            string
+	FolderIcon          string
+	WindowsRegistryIcon string
 }
 
 const (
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@tylerbutler good idea, I'll have a look tomorrow 👍🏻

JanDeDobbeleer:
Rather than prepopulate every style, I would keep style as a property. That gives the option to only format the directory once, and we can expose that as `.Path` for example. It would also be a good idea to expose the actual path `.PWD` so there can be colouring options based on that value too.

JanDeDobbeleer:
Just delivered it

tylerbutler:
@JanDeDobbeleer Nice! Sorry I didn't get back to this PR quickly. I'm glad to see the feature make it in though! :)

Issue #737, opened 2021-05-24
Battery info not displaying when unplugged


justin-vogt:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

When I unplug my laptop, the battery segment indicates an error state.
When I run this powershell command locally I get a valid reponse:

`(Get-WmiObject win32_battery).estimatedChargeRemaining`

### Environment

- Oh my Posh version: 3
- Theme: Custom
- Operating System: Windows 10
- Shell: Powershell
- Terminal: Windows Terminal

### Steps to Reproduce

1. Unplug laptop from powersource
2. Have battery the following battery segment in your theme

```
{
          "type": "battery",
          "style": "plain",
          "foreground": "#193549",
          "background": "#2f2f2f",
          "properties": {
              "battery_icon": "",
              "discharging_icon": "\uE231 ",
              "charging_icon": "\uE234 ",
              "charged_icon": "\uE22F ",
              "color_background": true,
              "charged_color": "#4caf50",
              "charging_color": "#40c4ff",
              "discharging_color": "#ff5722",
              "prefix": "<#4caf50,transparent>\uE0B6</> ",
              "postfix": "\u003c#193549\u003e%\u003c/\u003e\u003c#FFFFFF\u003e \u003c/\u003e",
              "display_charging": true,
              "display_error": true
            }
        },

```        
3. Execute command in shell

**Expected behavior:** [What you expected to happen]
I expect to see battery level information displayed.
I expect to see that the battery is not currently charging.

**Actual behavior:** [What actually happened]
Error
`[{ChargeRate:Unknown value received}]% `

lnu:
Works fine here with your config. But it can happen that sometimes you receive a wrong value and get the message(I had it once when testing then working fine). Is it all the time?

I received this error while testing(not charging status) in wsl:
https://github.com/distatus/battery/pull/16

Issue with distatus/battery

![image](https://user-images.githubusercontent.com/1829553/119474758-ced28780-bd4c-11eb-930a-801f9712feb9.png)
![image](https://user-images.githubusercontent.com/1829553/119475013-03464380-bd4d-11eb-985b-127d4e69c15c.png)
![image](https://user-images.githubusercontent.com/1829553/119475300-528c7400-bd4d-11eb-8f16-ca302fb8cd4d.png)


[...]

Issue #1405, opened 2021-12-10
Wifi segment not working when system language isn't English


Khaos66:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

The `wifi` segment is not showing on my German system

### Theme

Any theme using `wifi`

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

```shell
.
```

Khaos66:
This is the output of `netsh.exe wlan show interfaces`
![image](https://user-images.githubusercontent.com/4013009/145643228-e9039c1e-b16b-490d-9c37-96c2008928d2.png)

but `parseNetshCmdResult` only accepts english text

Khaos66:
`netsh` seems to be a tricky one. It only returns localized output no matter what.
There is a `wifi` go lib: https://github.com/mdlayher/wifi
But it only supports unix

On windows there is an API for getting the network and wifi state
https://docs.microsoft.com/en-us/windows/win32/api/netlistmgr/nn-netlistmgr-inetworklistmanager

Khaos66:
Using the win32 api would be the cleanest way, but maybe it is enought to have [all constants here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segment_wifi.go#L87) configurabe by the user

JanDeDobbeleer:
Win32 it is 😬 I'm not pushing that burden to the user.

For reference: https://github.com/Tviljan/WifiHotspotStatus/blob/master/WifiHotspotStatus/Program.cs

Khaos66:
It would be nice to contribute to https://github.com/mdlayher/wifi and use this lib. Then wifi will work on Linux as well

JanDeDobbeleer:
@Khaos66 really good point. But we need to get it to work first 😉

Khaos66:
> Win32 it is 😬 I'm not pushing that burden to the user.
> 
> For reference: https://github.com/Tviljan/WifiHotspotStatus/blob/master/WifiHotspotStatus/Program.cs

The API used in this reference is used to monitor a wifi hotspot **hosted on the current system**. That's not really what we want here, right? 
The `wifi` segment shows information about the currently **connected** (not hosted) wifi network, correct?

Update:
Maybe this is a better reference:
https://github.com/dahall/Vanara/blob/master/PInvoke/NetListMgr/NetListMgr.cs

Khaos66:
[...]

PR #1407, opened 2021-12-11
fix(battery): ignore UnknownChargeRate error

Khaos66:
This error is returned when Windows can't get the charge rate from the battery

### Prerequisites

### Description

On some Windows systems the battery doesn't supply a `charge rate value`. Then an error is returned by the `battery` lib.
This error can be ignored. The battery stats are still correct.

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -399,15 +399,42 @@ func (env *environment) getBatteryInfo() ([]*battery.Battery, error) {
 			validBatteries = append(validBatteries, batt)
 		}
 	}
+	// clean minor errors
 	unableToRetrieveBatteryInfo := "A device which does not exist was specified."
+	unknownChargeRate := "Unknown value received"
+	var fatalErr battery.Errors
+	ignoreErr := func(err error) error {
+		if e, ok := err.(battery.ErrPartial); ok {
+			// ignore errors unknown charge rate value
+			if e.Current == nil &&
+				e.Design == nil &&
+				e.DesignVoltage == nil &&
+				e.Full == nil &&
+				e.State == nil &&
+				e.Voltage == nil &&
+				e.ChargeRate.Error() == unknownChargeRate {
+				return nil
+			}
+		}
+		return err
+	}
+	if batErr, ok := err.(battery.Errors); ok {
+		for _, err := range batErr {
+			err = ignoreErr(err)
+			if err != nil {
+				fatalErr = append(fatalErr, err)
+			}
+		}
+	}
+
 	// when battery info fails to get retrieved but there is at least one valid battery, return it without error
-	if len(validBatteries) > 0 && err != nil && strings.Contains(err.Error(), unableToRetrieveBatteryInfo) {
+	if len(validBatteries) > 0 && fatalErr != nil && strings.Contains(fatalErr.Error(), unableToRetrieveBatteryInfo) {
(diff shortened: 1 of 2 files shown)


Khaos66:
@JanDeDobbeleer This fixes #737 for me. Please review
May still need to add a unit test though

Khaos66:
Ok I tried to create a unit test, but there are very few tests for `environment.go`. Maybe because you can't use `MockedEnvironment` to test `environment` =D
Not sure how to mock `battery.getAll()`
@JanDeDobbeleer  Any suggestions?

src/environment.go
```diff
@@ -399,15 +399,42 @@ func (env *environment) getBatteryInfo() ([]*battery.Battery, error) {
...
+	ignoreErr := func(err error) error {
+		if e, ok := err.(battery.ErrPartial); ok {
+			// ignore errors unknown charge rate value
+			if e.Current == nil &&
+				e.Design == nil &&
+				e.DesignVoltage == nil &&
+				e.Full == nil &&
+				e.State == nil &&
+				e.Voltage == nil &&
+				e.ChargeRate.Error() == unknownChargeRate {
```
Khaos66, line 416: Does it require a `nil` check here?

src/environment.go
```diff
@@ -399,15 +399,42 @@ func (env *environment) getBatteryInfo() ([]*battery.Battery, error) {
 			validBatteries = append(validBatteries, batt)
 		}
 	}
+	// clean minor errors
 	unableToRetrieveBatteryInfo := "A device which does not exist was specified."
+	unknownChargeRate := "Unknown value received"
+	var fatalErr battery.Errors
+	ignoreErr := func(err error) error {
```
JanDeDobbeleer: Why not return a bool? Saves some space below, you already have the error.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@Khaos66 I wouldn't test this. It's only reliable from an integration perspective anyways.

Khaos66:
> @Khaos66 I wouldn't test this. It's only reliable from an integration perspective anyways.

Good point =D I copied this from the battery lib itself and changed it a little bit to fit this case. But a bool makes more sense.

src/environment.go
Khaos66: Good point =D I copied this from the battery lib itself and changed it a little bit to fit this case. But a bool makes more sense.

Khaos66: @JanDeDobbeleer  Done

src/environment.go
JanDeDobbeleer: No, not to my understanding

Khaos66:
@JanDeDobbeleer Thanks for merging! 
I installed the new version and it works like a charm <3
Now I only need to sort out #1405 and I'm all good ;)

Khaos66:
@JanDeDobbeleer Sorry! I just found that I had a typo in my commit message =/ 
Can you please fix the release notes of the current version? 
It should be `UnknownChargeRate` not `UnknownChareRate` ... would be great if you could fix that
Thanks <3

JanDeDobbeleer:
@Khaos66 done!

JanDeDobbeleer:
@All-contributors please add @Khaos66 for code and doc

PR #1410, opened 2021-12-11
feat(wakatime): added Wakatime segment

Khaos66:
### Prerequisites

### Description

This adds a new segment for [wakatime](https://wakatime.com/). It displays the tracked time of the current day


Khaos66:
Sorry, I forgot to change some comments when copying owm... but it should be all clear now

src/segment_wakatime.go
```diff
@@ -0,0 +1,103 @@
...
+	Minutes      int
+	MinutesTotal int
+}
+
+const (
+	// CacheKeyResponse key used when caching the response
+	WTCacheKeyResponse string = "wt_response"
+)
+
+type wtTotals struct {
```
JanDeDobbeleer, line 42: I would expose this as well, no? We can use inheritance to automaticaly add these after parsing the response.

src/segment_wakatime.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return response, nil
+		}
+	}
+
+	apikey := w.props.getString(APIKey, ".")
+	httpTimeout := w.props.getInt(HTTPTimeout, DefaultHTTPTimeout)
+	url := fmt.Sprintf("https://wakatime.com/api/v1/users/current/summaries?start=today&end=today&api_key=%s", apikey)
```
JanDeDobbeleer: We could also expose the url including the `api_key` like the `nightscout` segment?

JanDeDobbeleer (changes requested):
Two remarks, really well done.

src/segment_wakatime.go
Khaos66: Not totally sure what you mean by inheritance, but i cound expose this, yes. Will do and we'll see if that's hat you expected ;)

src/segment_wakatime.go
Khaos66: I don't see the point in doing so, but sure I will

JanDeDobbeleer: That's a question more than a remark. I was wondering if people might want to use different time spans (hardcoded to a day now).

src/segment_wakatime.go
JanDeDobbeleer: By adding that struct to the segment struct. You can then also parse the response with that embedded struct (like with git and scm).

Khaos66:
@JanDeDobbeleer should be good now

Khaos66:
hmm.. I crashed the linter with my comment... again ... will fix

Khaos66:
@JanDeDobbeleer can you plz help me? Whats the problem of the code validator now?
Edit: 
I found the problem

JanDeDobbeleer:
I'm reviewing this in the morning, need my desk for this. I think I made things worse by not being clear enough, I'll fix that together with you 😅

Khaos66:
> I'm reviewing this in the morning, need my desk for this. I think I made things worse by not being clear enough, I'll fix that together with you 😅

Ok. Feels strange being on the other side of this for once. In my day to day business I'm the one giving ~_unclear_~ **high level** advice 😅 
Take your time. We'll get there

Khaos66:
@JanDeDobbeleer It's feature complete now. Please elaborate the inheritance issue. Then we can merge I guess =) 
You don't need to hurry. Take your time.

JanDeDobbeleer:
@Khaos66 I'm writing an implementation based on your commits. I had what seems to be a brilliant idea.

Khaos66:
Great news. I can't wait to see it 😃

PR #1418, opened 2021-12-13
feat(wakatime): add segment

JanDeDobbeleer:
### Prerequisites

### Description

Different approach, leverages temple functions (which can be reused by the `executiontime` segment in templates)

### DIFF
--- a/docs/docs/segment-wakatime.md
+++ b/docs/docs/segment-wakatime.md
@@ -1,0 +1,57 @@
+---
+id: wakatime
+title: Wakatime
+sidebar_label: Wakatime
+---
+
+## What
+
+Shows the tracked time on [wakatime][wt] of the current day
+
+:::caution
+
+You **must** request an API key at the [wakatime][wt] website.
+The free tier for is sufficient. You'll find the API key in your profile settings page.
+
+:::
+
+## Sample Configuration
+
+```json
+{
+    "type": "wakatime",
+    "style": "powerline",
+    "powerline_symbol": "\uE0B0",
+    "foreground": "#ffffff",
+    "background": "#007acc",
+    "properties": {
+        "prefix": " \uf7d9  ",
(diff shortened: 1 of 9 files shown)


JanDeDobbeleer:
@Khaos66 this has a few advantages:

- no longer using the complex settings (just add the URL with your params)
- leverage a template function to format the Seconds

Khaos66:
@JanDeDobbeleer I get your point now ;D
Yes that's simpler on this side. Though, the user has to read the API docs and build the URL on its own now.
But we're serving devs here, so this should be ok

docs/docs/segment-wakatime.md
```diff
@@ -0,0 +1,56 @@
...
+        "cache_timeout": 10,
+        "http_timeout": 500
+    }
+},
+```
+
+## Properties
+
+- url: `string` - Your Wakatime summaries URL, inclding the API key. Example above. You'll know this works if you can
+- curl it yourself and get a single value. - defaults to ``
```
Khaos66: You could add a link to the actual API doc: 
https://wakatime.com/developers#summaries

Khaos66:
@JanDeDobbeleer This segment is kind of generic now. Apart from the response structs there is not much left, that is very wakatime specific...
Would it be possible to remove these structs as well? Then this would be a generic `webapi` segment and could be used for ANY rest api =)
I'm not sure how to document wakatime then. Because I would want to point users on that great service ;D 
Maybe a sub-section in the docs where all kinds of webapis are documented, so the user can copy&paste an api easily?
What do you think?

JanDeDobbeleer:
@Khaos66 there's a difference between architecture for oh-my-posh and how we expose that to the outside. I would wait to streamline this once we have generics as the parsing of the json is going to be cumbersome otherwise. But you're right, we can reuse this logic.

PR #1432, opened 2021-12-14
feat(brewfather): brewfather segment work squashed

memcpy-rand-rand-rand:
segment code and docs
modification to doGet to support request modifier funcs

### Prerequisites

- [X] Docs have been added / updated (for bug fixes/features)

feat(brewfather): show brew batch information from Brewfather's REST API

new segment code and docs
additional functionality on env.doGet to add request modifiers

### DIFF
--- a/src/environment.go
+++ b/src/environment.go
@@ -56,6 +56,8 @@ type cache interface {
 	set(key, value string, ttl int)
 }
 
+type doGetRequestModifier func(request *http.Request)
+
 type windowsRegistryValueType int
 
 const (
@@ -96,7 +98,7 @@ type environmentInfo interface {
 	getShellName() string
 	getWindowTitle(imageName, windowTitleRegex string) (string, error)
 	getWindowsRegistryKeyValue(path string) (*windowsRegistryValue, error)
-	doGet(url string, timeout int) ([]byte, error)
+	doGet(url string, timeout int, requestModifiers ...doGetRequestModifier) ([]byte, error)
 	hasParentFilePath(path string) (fileInfo *fileInfo, err error)
 	isWsl() bool
 	stackCount() int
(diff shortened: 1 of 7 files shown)


src/segment_brewfather.go
```diff
@@ -0,0 +1,341 @@
...
+
+	return bf.props.getString(BFFlatIcon, "→")
+}
+
+func (bf *brewfather) getBatchStatusIcon(batchStatus string) string {
+	switch batchStatus {
+	case BFStatusPlanning:
+		return bf.props.getString(BFPlanningStatusIcon, "")
+	case BFStatusBrewing:
+		return bf.props.getString(BFBrewingStatusIcon, "")
```
JanDeDobbeleer: the defaults need the `\uxxxx` notation

src/segment_brewfather.go
```diff
@@ -0,0 +1,341 @@
...
+		bf.URL = fmt.Sprintf("https://web.brewfather.app/tabs/batches/batch/%s", batchID)
+	}
+
+	bf.DefaultString = bf.formatDefaultString()
+
+	return true
+}
+
+// build a default string for the segment using Sprintf for speed and to provide a simple starting point without complex template logic
+func (bf *brewfather) formatDefaultString() string {
```
JanDeDobbeleer: You could also add default template? I want to move away from the string based implementations. I see you have both but it doesn't seem like `DefaultString` is used anywhere, right? If it's up to me, I would not have this function and only rely on the default template.

src/segment_brewfather.go
memcpy-rand-rand-rand: That makes more sense than the evil glyph boxes of doom!

src/environment.go
```diff
@@ -56,6 +56,8 @@ type cache interface {
 	set(key, value string, ttl int)
 }
 
+type doGetRequestModifier func(request *http.Request)
```
JanDeDobbeleer: I like the idea, would rename to `HTTPRequestModifier`

JanDeDobbeleer (changes requested):
I love this addition. It almost too crazy to believe that there's a brewing segment. Just a few remarks, let's do this :-)

src/segment_brewfather.go
memcpy-rand-rand-rand: I pondered this one for a bit - figured a quick coded version that might be the golden path for many would be more efficient that trying to replicate with template logic.  Mainly coming at it from a perf angle, but I haven't profiled it.

It does feel a bit janky to have both.

I'll switch to default template for consistency.

src/segment_brewfather.go
memcpy-rand-rand-rand: fixed and pushed.

src/segment_brewfather.go
memcpy-rand-rand-rand: fixed and pushed

src/environment.go
memcpy-rand-rand-rand: done - pushed.

src/segment_brewfather.go
memcpy-rand-rand-rand: on that topic - I notice some spacing misbehaviours depending on the glyph, are you tracking issues on that already?

Edit: never mind, it's our old friend the Terminal bug.

PR #1505, opened 2021-12-27
Added Random theme loader for Windows Powershell & CMD

MRDGH2821:
### Prerequisites

### Description

Added scripts to load random theme on new terminal instance.
Scripts added only for Windows Powershell & Command prompt.


MRDGH2821:
I tried to figure out how to put it into tabbed way, but it was too complicated to understand hence I wrote the things as it is

MRDGH2821:
And I have access only to Windows, so the random theme loader is limited to Windows only 😅

lnu:
should not, powershell runs on macos and linux as well. We should check how to avoid the hardcoded appdata path.
If installed using brew: gci "$(brew --prefix oh-my-posh)/themes" | get-random
All installers have a way to find the theme path.

JanDeDobbeleer:
This is an interesting idea, but a terrible one to support. I'm not in favor of merging this myself. Not that I want to undo the effort, but it's a very minimal use case with a potentially high maintenance effort on my end if things don't work as expected.

mattcargile:
I would think one could relegate this work to the applicable "dotfile" ( i.e. `.bashrc`, `$PROFILE`, etc). Maybe I might be missing the point though.

docs/docs/themes.md
```diff
@@ -16,6 +16,57 @@ Once you're ready to swap to a theme, follow the steps described in [🚀Install
 
 Themes with `minimal` in their names do not require a Nerd Font. Read about [🆎Fonts][fonts] for more information.
 
+To set a random theme everytime you open terminal, you can use the following:
+
+###### Windows PowerShell
```
JanDeDobbeleer: please rename the file to `.mdx` and add tabs (see any other `.mdx`) file for reference.

docs/docs/themes.md
```diff
@@ -16,6 +16,57 @@ Once you're ready to swap to a theme, follow the steps described in [🚀Install
...
+local theme = math.randomchoice(themes)
+local query =
+    string.format(
+    "oh-my-posh --config=%%userprofile%%\\AppData\\Local\\Programs\\oh-my-posh\\themes\\%s --init --shell cmd",
+    theme
+)
+print(theme) --This line can be removed if you don't want to know which theme is loaded.
+load(io.popen(query):read("*a"))()
+```
+[Source](https://gist.github.com/MRDGH2821/a05aa96724484e3bce55c3c66855b933)
```
JanDeDobbeleer: This needs to be a pointer to the URL rather than include it directly (see the ones below):

```markdown
[Source][source]

[source]: https://gist.github.com/MRDGH2821/a05aa96724484e3bce55c3c66855b933
```

JanDeDobbeleer (changes requested):
Two remarks, we can merge this once that's implemented.

docs/docs/themes.md
MRDGH2821: Changes done as requested

MRDGH2821:
Changes done as requested.

Also I was wondering if random theme loading capability itself could be added into oh-my-posh instead of these separate scripts, since oh-my-posh definitely knows the location of themes folder. But at the same time it also feels complicated procedure to do so.

One more concern, while I used proper syntax for admonitions (by referring Docusaurus guide) the info admonition didn't render in themes page, where as in other pages it did.

MRDGH2821:
Also, about the Lua script to load random theme in CMD, I'm not familiar with Lua scripting. All I did was copy paste resources and it worked flawlessly.

Hence there can be some scope of improvement in lua script (which I'm in no position to do, I'm C/C++/Python dev. Not Lua scripter)

I'm also fine with the PR being rejected since the scripts would increase load on Devs if they don't work reliably.

Finally Thank you for accepting my contribution!

JanDeDobbeleer:
Due to the amount of work I already have I decided to not include this (sorry it took so long).

PR #1520, opened 2021-12-30
Add package manager segment

paulhunttech:
### Prerequisites

### Description

**Help required before this can be merged!**

Added a package manager segment that shows the number of updates are available. Currently only works with [Chocolatey](https://community.chocolatey.org/) but designed so it can be extended for other package managers like WinGet.

The reason this can't be merged yet is because the performance is not good enough as it takes multiple seconds to run `choco outdated`. I have tried using a goroutine to run the command in a background thread but as soon as main exits then the goroutine is killed.


JanDeDobbeleer:
@paulhuntcode this is next for me after #1535

rfverbruggen:
What a coincidence, this weekend I created a winget segment for myself and was testing it to submit it, and now I discovered this PR. As soon as it is merged I'll alter my version to add it to this segment.

paulhunttech:
> What a coincidence, this weekend I created a winget segment for myself and was testing it to submit it, and now I discovered this PR. As soon as it is merged I'll alter my version to add it to this segment.

I've never used winget myself, did you have any issues with performance? For chocolately, the outdated command is just too slow to run synchronously.

rfverbruggen:
Yes it is slow as well because in the background it is refreshing the feeds and comparing against your installed software. I've added a cache to only check for updates once an hour (more than sufficient for me). But the downside is that if you then upgrade your software it is still telling you for one hour that you have 'x' amount of upgrades pending.

PR #1535, opened 2022-01-02
Added segment for Strava

larserikfinholt:
### Prerequisites

### Description
Added support for displaying activity data from Strava.  This will remind users (me) that it is time to get away from the laptop, and go for workout. It can show properties from your last run or ride, and indicate by color that it has been to long since you where out training.

image

[Description of the change]
Added segment that calls an api that should return your latest Strava activity. I calculate the hours since that activity, and you may use that to colorize the prompt. In the docs I'm showing how you could build that api. 

Question - I did not follow the git convention, but instead created a many minor commits. Could the PR still be merged as a squash commit, the the following message: `feat: added strava segment`


docs/docs/segment-strava.md
```diff
@@ -0,0 +1,123 @@
...
+  it yourself and get a single value. - defaults to ``
+- http_timeout: `int` - How long do you want to wait before you want to see
+  your prompt more than your strava data? I figure a half second is a good default -
+  defaults to 500ms
+- template: `string` - a go [text/template][go-text-template] template extended
+  with [sprig][sprig] utilizing the properties below.
+  See the example above.
+  Make sure your NerdFont has the glyph you want or search for one
+  at nerdfonts.com
+- TrainingDataCacheTimeout: `int` in minutes - How long do you want your numbers cached? -
```
JanDeDobbeleer: This needs to be renamed to `CacheTimeout` for conistency

docs/docs/segment-strava.md
```diff
@@ -0,0 +1,123 @@
...
+ To keep up with your training goals it is important to be constantly reminded about it.
+
+A Strava Oh My Posh segment would show your last activity,
+ and also indicate by a color if it is time to get away from your computer and do some workout.
+
+## Accessing your Strava data
+
+There are several ways you could get your last activity.
+
+### Build you own api, integrating with Strava
```
JanDeDobbeleer: This won't work for most people so we can't do it like this. We have to access the API directly to get the latest activity rather than forcing people to become full blown programmers (which a lot won't be able to do)

The following API endpoints lists all activities for a given athlete:
`https://www.strava.com/api/v3/athletes/id/activities`

The only challenge is then getting the actual token as it needs to be refreshed in a timely fashion. That's maybe something we can facilite by having that logic inside oh-my-posh by means of a link you can click so you can provide an `access_token` and `refresh_token` for us to store.

JanDeDobbeleer (changes requested):
We need to make this more accessible for people to consume. I believe we can. by creating a simple redirect on the oh-my-posh website that displays the credentials for the user to add to the settings.

docs/docs/segment-strava.md
larserikfinholt: Yes, that would absolute be a more user friendly approach.  I think that to access strava api we have two options:

1. Ask users to create an app in the strava developer portal
2. Someone else (you or me) creates an app in strava developer portal

This app must be linked to a url to allow for login (and then get the access and refresh tokens). The user will have to grant access to that url also.

Do you think adding a https://ohmyposh.dev/auth/strava is a good idea? That would make it much easier for future segments that need this type of data.  I'm not sure if I'm up for that job.

As an alternative I woke up my old project fartslek.no (at a new url https://fartslekreg.azurewebsites.net ) and provided a simple endpoint that returns the last activity  by a token. I did have plans for this, so just now, you may go and try out fartslekreg for a simple way of getting your last activity.

I've also updated the docs.

JanDeDobbeleer: Yes, I would create that path. We have the ability to add an [API endpoint](https://docs.microsoft.com/en-us/azure/static-web-apps/add-api?tabs=vanilla-javascript) to the docs as it's an Azure Static web site. This has got quite the information, maybe I'll give that a go but the site is part of this codebase under /docs so anyone could add it.

JanDeDobbeleer: @larserikfinholt I just pushed a dummy one at https://github.com/JanDeDobbeleer/oh-my-posh/pull/1536

larserikfinholt: @JanDeDobbeleer  Just to make sure I understand:

* The user opens https://ohmyposh.dev/auth/strava?login
* This page shows a "login" button
* User clicks this button, is redirected to strava login page
* Strava then ask for grant to access the `activity:read` scope (and is asked "do you grant access to OhMyPoshStravaSegment https://ohmyposh.dev")
* User accepts, and is redirected to https://ohmyposh.dev/auth/strava
* This page then display the access token, and refresh token, and ask user to copy to clipboard
* User should then maybe store that as a "initialTokens" property in the config file.
* Then the segment uses this tokens from "initialTokens" property the first time, and then stores the updated refreshtoken, along with access tokens and expiry dates in a TEMP file. Tokens from this file is used (and updated) for all future calls

The webpage part is implemented as you showed in the sample #1536 I assume we could store the Strava `ClientSecret` and `ClientId` in azure, and access it from the function. I don't think theres a need for a fancy gui, just return plain text, and a button. Agree?

Do you agree that its a good idea that you "own" the api? In case you need to do the following:
* Create a API application https://www.strava.com/settings/api
* Add `ClientId` and `ClientSecret` as AppSettings i azure portal (so I may retrive them from `process.env.ClientId` and `process.env.ClientSecret` in the function)
* Add `https://ohmyposh.dev` as a Authorization Callback Domain
* Add the OhMyPosh logo

I suggest I add the api part in a separate PR, so that you may enter `ClientId` and `ClientSecret` in Application settings, and we could verify the login process separatly?

The rest of the logic will be implemented in the `segment_strava.go` file. I'm unsure how to create and update a temp file in go (this is my first time using go) but assume that should work ok. Any problems related to files, access and different OS?

Other things that should be considered?

JanDeDobbeleer: I would indeed construct the authentication first in a separate PR. The way I see it we can redirect directly to Strava, no need for a button as we'll have to trigger the user to click the link from the prompt anyways (we can print it or make a clickable link, let's figure that out afterwards). Ideally the landing page had a copy-paste snippet at the ready, but also something we can make better once it works. Let's assume you're the first and probably only user so we can take our time 😬

I already have a Strava API client created, I'll build the function ;-)

larserikfinholt: Ok, good. Could you let me know when you have decided on the format the user should copy and paste into the config files "initialTokens" property? I suggest something like this:

```
'{"access_token":"ACCESS_TOKEN","refresh_token":"REFRESH_TOKEN","expires_at":123123132,"atheleteId":123123}'
```

JanDeDobbeleer: @larserikfinholt I would only create two properties, `access_token` and `refresh_token`. We can try to refresh, it it fails we'll display the link to authenticate again.

larserikfinholt: @JanDeDobbeleer agree. I'll assume the user will receive this from you, and save it in the config file `'{"access_token":"ACCESS_TOKEN","refresh_token":"REFRESH_TOKEN"}'`

Any guidance on the where I should store the updated tokens? Something like this:
```
tokensPath = fmt.Sprintf("%s/.strava/tokens", a.env.homeDir())
```

JanDeDobbeleer: We have a cache you can use in Environment. If you set the timeout to the expiration of the refresh token, we immediately have a way to detect if we no longer have credentials.

JanDeDobbeleer: @larserikfinholt I finally debugged the whole thing locally :-) I have a working api setup that prints the access and refresh token using the oh-my-posh client. Debugging a Static Web App API endpoint in Azure is in beta it seems and the Azure version still gives a 500 (and have no idea why just yet).

larserikfinholt: @JanDeDobbeleer  great! 

I've just started implementing the refresh logic, and relized that I'll be needing the ClientId and ClientSecret to do the refresh.  Do you have some built in logic where I could get those? Some global setting?

JanDeDobbeleer: Allow me to also implement an endpoint for that.

JanDeDobbeleer: I was going MENTAL: https://github.com/Azure/static-web-apps/issues/165

JanDeDobbeleer: @larserikfinholt [link](https://www.strava.com/oauth/authorize?client_id=76033&response_type=code&redirect_uri=https://ohmyposh.dev/api/auth&approval_prompt=force&scope=read,activity:read
)

larserikfinholt: @JanDeDobbeleer  ok. I just completed a (untested) implementation of the refresh logic, you may use it. 

The link works fine. I suggest using a format that is simple to paste into the property, maybe something like this:

```
access_token:11111111111111111,refresh_token:1111111111111111
```

JanDeDobbeleer: I would use two properties, just like they are there.

```json
{
  "type": "strava",
  "properties": {
    "access_token": "11111111111111111",
    "refresh_token": "11111111111111111"
  }
}
```

larserikfinholt: @JanDeDobbeleer just pushed a implementation, but it does not work - I get a "invalid token" error from Strava when using the access token I receive from your api. Any idea what could be wrong?

(as you can see, I've removed the "getNewAccessToken" and is always using the one from config - but that should be valid for 6 hours, so I don't think thats the problem)

About the token exchange - should we just reuse the existing api from `https://www.strava.com/api/v3/oauth/token` ? and place it at `https://ohmyposh.dev/api/strava/token` (of course without the need to post clientId and clientSecret)

lnu: @larserikfinholt  I just tested the token received from @JanDeDobbeleer  link in postman and it works(I get my stats). Works also with you pr:
image

larserikfinholt: @lnu Cool! Glad to see its working. Yes, I fixed a small typo in auth header :-)

larserikfinholt: Updated with tests - only thing missing is to enable the use tokenservice when @JanDeDobbeleer  is ready.

JanDeDobbeleer: I'm back so I'll have a look ;-)

src/segment_strava.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (ns *strava) getAccessToken() (string, error) {
+	var at, rt string
+	var hasAt, hasRt bool
+	var err error
+
+	// get directly from cache
+	at, hasAt = ns.env.cache().get(StravaAccessToken)
+	if !hasAt {
```
JanDeDobbeleer: If you invert this, you can return the statement at the bottom and continue inline rather than indented. It can even be a single statement:

```go
if access_token, OK = ns.env.cache().get(StravaAccessToken); ok {
    return access_token, nil
}
```

src/segment_strava.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+// segment struct, makes templating easier
+type strava struct {
+	props Properties
+	env   Environment
+
+	StravaData
+	ActivityIcon string
+	Ago          string
```
JanDeDobbeleer, line 19: We should use a template function for this. My plan was to extend the template functions to have these more generic, for example, the executiontime segment already has quite some "formatters" which can also be used elsewhere (I still need to do that though).

src/segment_strava.go
```diff
@@ -0,0 +1,258 @@
...
+func (ns *strava) getAccessToken() (string, error) {
+	var at, rt string
+	var hasAt, hasRt bool
+	var err error
+
+	// get directly from cache
+	at, hasAt = ns.env.cache().get(StravaAccessToken)
+	if !hasAt {
+		rt, hasRt = ns.env.cache().get(StravaRefreshToken)
+		if hasRt {
```
JanDeDobbeleer: do we want to do this, or try the `access_token` and if that call fails, use the `refresh_token` and try again?

src/segment_strava.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+		if err != nil {
+			return "", err
+		}
+	}
+	return at, err
+}
+
+// This method will be implemented in ohmyposh.dev api (I guess it works, just neet to set StravaClientSecret)
+func (ns *strava) getNewAccessToken(refreshToken string) (string, error) {
```
JanDeDobbeleer: would rename to refreshToken(), we can also get the `refresh_token` here rather than provide it

docs/docs/segment-strava.md
JanDeDobbeleer: refresh is live at `https://ohmyposh.dev/api/refresh?segment=strava&token=refresh_token`

JanDeDobbeleer: @larserikfinholt if you can mark this PR as "Allow edits from maintainers" (can't see if that's enabled or not). See [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) for more info.

larserikfinholt: @JanDeDobbeleer yes, I it's already enabled. Please feel free do do anything :-) (Hopefully I'll get the chance do some updates tonight after work - if you've not already fixed it)

JanDeDobbeleer: I'll add commits on top so you can see the changes 😉

JanDeDobbeleer:
@larserikfinholt you need to reset your branch to your origin as I just force pushed it for simplicity

src/segment_strava.go
```diff
@@ -0,0 +1,222 @@
...
+			return acccessToken, nil
+		}
+		return "", &AuthError{}
+	}
+	// use initial tokens from property
+	accessToken := s.props.getString(AccessToken, "")
+	refreshToken := s.props.getString(RefreshToken, "")
+	if len(accessToken) == 0 || len(refreshToken) == 0 {
+		return "", &AuthError{}
+	}
```
larserikfinholt, line 144: @JanDeDobbeleer I think we should take one extra attempt to call refreshToken() to get an accesstoken from the initial refresh token before we givue up due to the following scenario:

* Refreshtoken stored in cache is invalid for some reason
* User has entered a new valid refreshtoken in the props

JanDeDobbeleer: Fair point!

src/segment_strava.go
larserikfinholt: @JanDeDobbeleer I think we're ready.

JanDeDobbeleer (approved):

JanDeDobbeleer:
@larserikfinholt the only thing I still want to check is if I can add a page that nicely renders the tokens in a code black the way you need to add them (rather than a raw json). But if OK I would already merge this while I add that. It's not required.

JanDeDobbeleer:
Somehow the API works locally but gives an error (that comes from Strava) in production. My bet is that the proxy is causing it. I'll try to debug this tomorrow.

larserikfinholt:
> Somehow the API works locally

And the redirect domain in Strava settings is updated I assume?

JanDeDobbeleer:
@larserikfinholt there's still an issue when it tries to refresh in 1 situation. Saw it happen twice, it displayed the "couldn't authenticate" message and simply doing it again displayed the segment correctly. Haven't been able to debug. The authentication flow should work.

JanDeDobbeleer:
Found the issue, on the first request to refresh the token, the API takes longer than expected. Meaning we need to wait a bit longer :-) Once that first call is done, we get a token rather fast the second time.

larserikfinholt:
@JanDeDobbeleer Just wanted to say a big thank you for your support on this PR. Even if this was my first ever `go` code, it was a great experience, thanks to your help. Thanks for taking the time and implement this feature!  

Now, let's go out and do some real production testing with a ride or run  :-)

JanDeDobbeleer:
@All-contributors please add @larserikfinholt for code

PR #1629, opened 2022-01-22
refactor: deprecate string functions

JanDeDobbeleer:
### Prerequisites

### Description

This will allow templates for all segments, but will also break some existing functionality.
It is needed for more flexibility and remove some useless properties.

- [x] add templates to all segments
- [x] remove `enable_hyperlink`
- [x] remove deprecated functions
- [x] change segmentWriter's string() function to template()
- [x] hyperlink template function e.g. `{{ hyperlink .Path .Location }}`
- [x] remove `prefix` and `postfix`
- [x] move the segments to their own module (architectural change)
- [x] create migration capabilities
- [x] auto migrate configuration by adding a `version` to `Config`
- [x] do not display segments with an empty or only spaces segment text
- [x] store a backup of the old configuration on migration


cdonnellytx:
Noticed this preexisting typo when doing a quick lookover of the PR to have an idea of what would have to change on mine:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/2a640266c83cb237671243d99d7d308695e3a730/docs/docs/contributing-segment.md#L171

JanDeDobbeleer:
One more refactor to do, will land this week.

JanDeDobbeleer:
@lnu can you do me a favour and build this one to see if it migrates your config right using `oh-my-posh --config $POSH_THEME --migrate --write`? I did a lot of testing here, manually and automated so I'm pretty confident but some additional validation could be really useful 🙏🏻 

I tried to auto migrate but there are so many edge cases here that this will need to be a one time manual effort.

lnu:
I'm typing as I test:
- you should backup the old one before overwriting it
- before
image
- after
image
dotnet segment custom template broken => display_mode issue, always visible
broken color in git segment
exit code spacing
image
Battery percentage missing => migrated template: `"template": " {{ if not .Error }}{{ $stateList := list \"Discharging\" \"Charging\" }}{{ if has .State.String $stateList }}{{ .Icon }}{{ .Percentage }}{{ end }}{{ end }}{{ .Error }}% "`

original config:
[original.txt](https://github.com/JanDeDobbeleer/oh-my-posh/files/7980359/original.txt)

migrated one:
[migrated.txt](https://github.com/JanDeDobbeleer/oh-my-posh/files/7980361/migrated.txt)

JanDeDobbeleer:
@lnu thank you so much. Will adjust.

I do see an issue with the following:

- the git segment is broken as there's a mix of template and old string() switches. I'll see if we can not migrate inline overrides in case a template was already present
- the dotnet segment is weird, I'll look into it
- The battery segment is a PITA as you can't migrate a postfix into a segment which is then always enabled. So that's always going to a quick manual override

What I'll still do: migrate automatically, I believe that if we add a version to the config we can do incremental upgrades if needed. This one can't run twice for example, mostly due to the pre/postfix adjustments.

lnu:
also this one is not migrated correctly:
```
        {
          "background": "#0077c2",
          "foreground": "#ffffff",
          "powerline_symbol": "",
          "properties": {
            "prefix": " \uFCB5 ",
            "mapped_shell_names": {
              "fish":"\uf739"
            }
          },
          "style": "powerline",
          "type": "shell"
        },
```
to
```
        {
          "background": "#0077c2",
          "foreground": "#ffffff",
          "powerline_symbol": "\ue0b0",
          "properties": {
            "mapped_shell_names": {
              "fish": "\uf739"
            },
            "template": " \ufcb5 {{ .Name }} "
          },
          "style": "powerline",
          "type": "shell"
        },
```

lnu:
console title template:
`"{{if .Root}} ⚡ {{end}}{{.Folder | replace \"~\" \"🏚\" }}@{{.HostName}}",`
to
`"{{if .Root}} \u26a1 {{end}}{{.Folder | replace \"~\" \"\u1f3da\" }}@{{.HostName}}",`

JanDeDobbeleer:
@lnu console title due to the glyphs being Unicode? That's on purpose. Shouldn't impact functionality

I don't see what's wrong with shell though?

lnu:
I think the behaviour of the text segment has changed:
- when it was empty, the segment was not rendered
- now it's rendered all the time

lnu:
> @lnu console title due to the glyphs being Unicode? That's on purpose. Shouldn't impact functionality
> 
> I don't see what's wrong with shell though?

before:
image
after:
image

JanDeDobbeleer:
@lnu depends on the template as the pre and postfix and included now that requires an inline if switch. If that evaluates to an empty string, the segment isn't rendered.

lnu:
> @lnu depends on the template as the pre and postfix and included now that requires an inline if switch. If that evaluates to an empty string, the segment isn't rendered.

ok, a blank space is considered not empty, my bad

lnu:
> @lnu console title due to the glyphs being Unicode? That's on purpose. Shouldn't impact functionality
> 
> I don't see what's wrong with shell though?

It's ok

JanDeDobbeleer:
> ok, a blank space is considered not empty, my bad

@lnu I'll fix that, was thinking about that indeed.

JanDeDobbeleer:
@lnu OK, so I did the following:

- migration now also backs up your existing config (same name but `.bak` suffix
- the inline color overrides are now skipped if a template already exists (see your git segment issue)
- emoji's are no longer escaped
- when a template is empty or whitespace, the segment is disabled
- the exit segment is now correctly migrated for `"display_exit_code": false`

What I can't fix:

- battery percentage, that should be handled by the user, moving it between `{{ .Percentage }}%{{ end }}` does the trick.
- botnet (or any language segment) works as designed, in this case it does not fetch the dotnet version as it's disabled, but it does render as it should when I change the switches.

If you have a moment, a re-validation would be immensely appreciated 🙏🏻 

I notice the text segment has an issue still, will fix.

lnu:
For the backup I see two things:
- the positive one is that it cleans the original config and save it back
-  but the downside is that it's not an exact backup of the original but the original file loaded and saved. So you lose for example the comments.

JanDeDobbeleer:
@lnu I can also do a copy-paste of the file data rather than what we read in?

lnu:
@JanDeDobbeleer Probably it's better to keep an exact copy of the original file, don't you think? On the opposite I also like the "cleaning" effect of the current migration(reordering of the properties, comments removed, ...).

JanDeDobbeleer:
@lnu done. I also validated all themes, everything that was wrong turned out to be an original issue except for one minor spacing issue due to an existing template. But trying to fix that would be a lot more work with weird edge cases vs for that occasion having the person add a space in the right spot. I'm pretty confident in merging this now.

PR #1656, opened 2022-01-26
Add This new theme.

if-tea:
Please accept the request and add this newly beautiful theme to the repository. Thanks in advance. ✌

### DIFF
--- a/themes/if_tea.omp.json
+++ b/themes/if_tea.omp.json
@@ -1,0 +1,183 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "blocks": [
+    {
+      "alignment": "left",
+      "newline": true,
+      "segments": [
+        {
+          "foreground": "#1865f5",
+          "properties": {
+            "alpine": "\uf300",
+            "arch": "\uf303",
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@if-tea what's the difference with #1648?

if-tea:
@JanDeDobbeleer I mistakenly added the image to my previous commit. That is why I committed for the second time. 🥰

JanDeDobbeleer:
@if-tea can you migrate the theme to config version 1? You can update oh-my-posh and use `oh-my-posh --config if_tea.omp.json --migrate --write` and update this PR.

JanDeDobbeleer:
@if-tea I did it for you, it does seem there's either an issue with the last block or something that's missing (the white hook at the start feels off)

if-tea:
I am creating another theme. Will push very soon .

JanDeDobbeleer:
@if-tea when merging this I need you agree we do this under MIT rather than GPL. If you would be so kind to acknowledge that, I can merge before #1807 goes in main.

if-tea:
> @if-tea when merging this I need you agree we do this under MIT rather than GPL. If you would be so kind to acknowledge that, I can merge before #1807 goes in main.

Of course I agree.

PR #1725, opened 2022-02-08
feat(haskell): add haskell segment

jedwillick:
### Prerequisites

- [x] The commit message follows the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) guidelines

### Description

Adds a Haskell segment which fetches the active ghc version. Can also differentiate between ghc on path and the sandboxed ghc used by [stack](https://docs.haskellstack.org/), but using stack ghc causes reduced performance. 
withStack
withoutStack

### DIFF
--- a/docs/docs/segment-haskell.md
+++ b/docs/docs/segment-haskell.md
@@ -1,0 +1,63 @@
+---
+id: haskell
+title: Haskell
+sidebar_label: Haskell
+---
+
+## What
+
+Display the currently active Glasgow Haskell Compiler (GHC) version.
+
+## Sample Configuration
+
+```json
+{
+  "type": "haskell",
+  "style": "powerline",
+  "powerline_symbol": "\uE0B0",
+  "foreground": "#906cff",
+  "background": "#100e23",
+  "properties": {
+    "template": " \ue61f {{ .Full }}"
+  }
+}
+```
+
+## Properties
+
+- home_enabled: `boolean` - display the segment in the HOME folder or not - defaults to `false`
+- fetch_version: `boolean` - display the GHC version - defaults to `true`
+- display_error: `boolean` - show the error context when failing to retrieve the version information - defaults to `true`
+- missing_command_text: `string` - text to display when the command is missing - defaults to empty
+- display_mode: `string` - determines when the segment is displayed
+  - `always`: the segment is always displayed
+  - `files`: the segment is only displayed when `*.hs`, `*.lhs`, `stack.yaml`, `package.yaml`, `*.cabal`,
+or `cabal.project` files are present (default)
+- stack_ghc_mode: `string` - determines when to use stack ghc to retrieve the version information.
+Using stack ghc will decrease performance.
+  - `never`: never use stack ghc (default)
+  - `package`: only use stack ghc when `stack.yaml` is in the root of the package
+  - `always`: always use stack ghc
+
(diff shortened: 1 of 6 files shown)


docs/docs/segment-haskell.md
```diff
@@ -0,0 +1,61 @@
...
+## Properties
+
+- home_enabled: `boolean` - display the segment in the HOME folder or not - defaults to `false`
+- fetch_version: `boolean` - display the GHC version - defaults to `true`
+- display_error: `boolean` - show the error context when failing to retrieve the version information - defaults to `true`
+- missing_command_text: `string` - text to display when the command is missing - defaults to empty
+- display_mode: `string` - determines when the segment is displayed
+  - `always`: the segment is always displayed
+  - `files`: the segment is only displayed when `*.hs`, `*.lhs`, `stack.yaml`, `package.yaml`, `*.cabal`, or `cabal.project` files are present (default)
+- use_stack_ghc: `string` - determines when to use stack ghc to retrieve the version information. Using stack ghc will decrease performance.
```
JanDeDobbeleer: can we rename this to `stack_ghc_mode`? That way it's in line with `display_mode`.

src/segments/haskell.go
```diff
@@ -0,0 +1,63 @@
...
+
+	h.language = language{
+		env:                env,
+		props:              props,
+		extensions:         []string{"*.hs", "*.lhs", "stack.yaml", "package.yaml", "*.cabal", "cabal.project"},
+		versionURLTemplate: "https://www.haskell.org/ghc/download_ghc_{{ .Major }}_{{ .Minor }}_{{ .Patch }}.html",
+	}
+
+	useStackGhc := h.props.GetString(UseStackGhc, "never")
+	if useStackGhc == "always" {
```
JanDeDobbeleer: would be cleaner if we transform this to a switch statement on `useStackGhc`:

```golang
switch useStackGhc {
    case "always":
        h.language.commands = []*cmd{stackCmd}
 	    h.StackGhc = true
    case "package":
        _, err := h.language.env.HasParentFilePath("stack.yaml")
        if err == nil {
             h.language.commands = []*cmd{stackCmd}
 	     h.StackGhc = true
 	}
    default:
        h.language.commands = []*cmd{ghcCmd}
 	h.StackGhc = false
}
```

JanDeDobbeleer (changes requested):

docs/docs/segment-haskell.md
jedwillick: changed everything related to `use_stack_ghc` to fit with `stack_ghc_mode`

src/segments/haskell.go
jedwillick: problem is for `package` I needed the command to be `ghcCmd` if  `stack.yaml` was not found. So instead of using a default case I just initially set the cmd in the language block hope thats okay?

jedwillick:
Thank you for the feedback!
Made the requested changes and fixed the markdown format.
 First time using `go`  so hopefully it wasnt too much of a mess!

JanDeDobbeleer:
@jedwillick it was beautiful 👌🏻

JanDeDobbeleer (approved):

JanDeDobbeleer:
@All-contributors please add @jedwillick for code

PR #1742, opened 2022-02-10
feat: UI5 tooling segment

wozjac:
### Prerequisites

### Description

A new segment for UI5 tooling - https://sap.github.io/ui5-tooling/
The check is based on a configuration file, by default it is ui5.yaml, might be also ui5-dist.yaml or any other name, which you can provide via --config param; but most of the times it has the "ui5" text in it. 

By default the file resides in the UI5 project root, but I added additional check to detect it from subdirectories. Usually you have the structure like:
- myproject
  - src
    - views
    - controllers
    ...

The check from subdirectories is up to 4 levels up, which should be enough for the most projects. 
From my daily experience - you can work with a large number of UI5 apps even within one project and jump frequently between them. Having the information exposed whether the app has the tooling configured is useful (you can do things differently). 

Best Regards,
Jacek

### DIFF
--- a/docs/docs/segment-ui5tooling.md
+++ b/docs/docs/segment-ui5tooling.md
@@ -1,0 +1,59 @@
+---
+id: ui5tooling
+title: UI5 Tooling
+sidebar_label: UI5 Tooling
+---
+
+## What
+
+Display the active [UI5 tooling](https://sap.github.io/ui5-tooling/) version (global or local if present -
+see [the documentation](https://sap.github.io/ui5-tooling/pages/CLI/#ui5-versions)).
+
+## Sample Configuration
+
+```json
+{
+  "background": "#f5a834",
+  "foreground": "#100e23",
+  "powerline_symbol": "\ue0b0",
+  "properties": {
+    "template": " \ufab6ui5 {{ .Full }} "
+  },
+  "style": "powerline",
+  "type": "ui5tooling"
+}
+```
+
+## Properties
+
+- home_enabled: `boolean` - display the segment in the HOME folder or not - defaults to `false`
+- fetch_version: `boolean` - display the UI5 tooling version - defaults to `true`
+- display_error: `boolean` - show the error context when failing to retrieve the version information - defaults to `true`
+- missing_command_text: `string` - text to display when the java command is missing - defaults to empty
+- display_mode: `string` - determines when the segment is displayed
+  - `always`: the segment is always displayed
+  - `files`: the segment is only displayed when `*ui5*.y(a)ml` file is present in the current folder
+  - `context`: (default) the segment is only displayed when `*ui5*.y(a)ml` file is present in the current folder
+    or it has been found in the parent folders (check up to 4 levels)
+
+## Template ([info][templates])
+
+:::note default template
+
+```template
+{{ if .Error }}{{ .Error }}{{ else }}{{ .Full }}{{ end }}
+```
+
+:::
+
+## Template Properties
+
+- `.Full`: `string` - the full version
+- `.Major`: `string` - major number
+- `.Minor`: `string` - minor number
+- `.Patch`: `string` - patch number
+- `.Prerelease`: `string` - prerelease info text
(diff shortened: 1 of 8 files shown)


docs/docs/segment-ui5tooling.md
```diff
@@ -0,0 +1,59 @@
+---
+id: ui5tooling
+title: UI5 Tooling
+sidebar_label: UI5 Tooling
+---
+
+## What
+
+Display the active [UI5 tooling](https://sap.github.io/ui5-tooling/) version (global or local if present -
```
JanDeDobbeleer: please use bottom style links (see https://commonmark.org/help/tutorial/07-links.html)

src/environment/shell.go
```diff
@@ -319,6 +320,25 @@ func (env *ShellEnvironment) HasFilesInDir(dir, pattern string) bool {
 	return len(matches) > 0
 }
 
+func (env *ShellEnvironment) HasFileInParentDirs(pattern string, depth uint) bool {
+	defer env.trace(time.Now(), "HasFileInParent", pattern, fmt.Sprint(depth))
+	currentFolder := env.Pwd()
+
+	for c := 0; c < int(depth); c++ {
```
JanDeDobbeleer, line 327: why uint and then convert to int here?

src/segments/ui5tooling.go
```diff
@@ -0,0 +1,49 @@
...
+		inContext:   u.inContext,
+		displayMode: props.GetString(DisplayMode, DisplayModeContext),
+		commands: []*cmd{
+			{
+				executable: "ui5",
+				args:       []string{"--version"},
+				regex:      `(?:(?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+))))`,
+			},
+		},
+		versionURLTemplate: "[%[1]s](https://github.com/SAP/ui5-cli/blob/master/CHANGELOG.md_V%[2]s.md#%[1]s)",
```
JanDeDobbeleer: This only needs to be the URL now, not the syntax `[]()`

src/segments/ui5tooling_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"fmt"
+	"oh-my-posh/environment"
+	"oh-my-posh/mock"
+	"oh-my-posh/properties"
+	"path/filepath"
+	"testing"
+
+	"github.com/alecthomas/assert"
```
JanDeDobbeleer: qhy the additional assert? We use `github.com/stretchr/testify/assert` by default.

JanDeDobbeleer (changes requested):
Really good PR, few small pointers/questions

src/segments/ui5tooling_test.go
wozjac: I missed this, I'll correct it.

src/segments/ui5tooling.go
wozjac: Noted, I'll correct this.

src/environment/shell.go
wozjac: When used without type conversion there is an error:
_cannot compare c < depth (mismatched types int and uint)compiler[MismatchedTypes]_
uint in the parameter seemed to me more accurate, carrying an intention of passing only a non-negative number. 
But I'm a fresher in Go world, with habits from my long-used languages, please advice what is the best or correct from your point of view.

docs/docs/segment-ui5tooling.md
wozjac: Noted, I'll correct this.

src/environment/shell.go
wozjac: Should it be rather
`for c := uint(0); c < depth; c++ {`
?

JanDeDobbeleer: You're right, the reasoning is solid.

JanDeDobbeleer (approved):

PR #1785, opened 2022-02-19
feat: cloud foundry target segment

wozjac:
### Prerequisites

### Description

Cloud Foundry target segment. Displays the details of Cloud Foundry endpoint you are logged in (after running `cf login`).
Zrzut ekranu 2022-02-19 123321

The details are taken from `cf target` output, which gives info in such format:
Zrzut ekranu 2022-02-19 152650

Best Regards
Jacek

### DIFF
--- a/src/segments/cf_target.go
+++ b/src/segments/cf_target.go
@@ -1,0 +1,83 @@
+package segments
+
+import (
+	"oh-my-posh/environment"
+	"oh-my-posh/properties"
+	"regexp"
+)
+
+type CfTarget struct {
+	props properties.Properties
+	env   environment.Environment
+
+	CfTargetDetails
+}
+
+type CfTargetDetails struct {
+	APIURL string
+	User   string
+	Org    string
+	Space  string
+}
+
+func (c *CfTarget) Template() string {
+	return "{{if .Org }}{{ .Org }}{{ end }}{{if .Space }}/{{ .Space }}{{ end }}"
+}
+
+func (c *CfTarget) Init(props properties.Properties, env environment.Environment) {
+	c.props = props
+	c.env = env
+}
+
+func (c *CfTarget) Enabled() bool {
+	return c.setCFTargetStatus()
+}
+
+func (c *CfTarget) getCFTargetCommandOutput() string {
+	if !c.env.HasCommand("cf") {
+		return ""
+	}
+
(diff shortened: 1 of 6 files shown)


src/segments/cf_target.go
```diff
@@ -0,0 +1,79 @@
...
+	"oh-my-posh/environment"
+	"oh-my-posh/properties"
+	"regexp"
+)
+
+type CfTarget struct {
+	props properties.Properties
+	env   environment.Environment
+
+	Target CfTargetDetails
```
JanDeDobbeleer: If you embed it directly, you get inheritance and the template doesn't need to use the `.Target.XX` anymore. You can just use `.XX`:

```go
type CfTarget struct {
	props properties.Properties
	env   environment.Environment

	CfTargetDetails
}
```

src/segments/cf_target.go
```diff
@@ -0,0 +1,79 @@
...
+func (c *CfTarget) setCFTargetStatus() {
+	output := c.getCFTargetCommandOutput()
+
+	if output != "" {
+		regex := regexp.MustCompile(`API endpoint:\s*(?P<api_url>http[s].*)|user:\s*(?P<user>.*)|org:\s*(?P<org>.*)|space:\s*(?P<space>(.*))`)
+		match := regex.FindAllStringSubmatch(output, -1)
+		result := make(map[string]string)
+
+		for i, name := range regex.SubexpNames() {
+			if i != 0 && name != "" {
```
JanDeDobbeleer: we always use `len(name) != 0`

I would also invert the statement and continue rather than indent the entire block:

```go
if i == 0 && len(name) == 0 {
    continue
}
...
```

JanDeDobbeleer (changes requested):

src/segments/cf_target.go
wozjac: Noted, will correct.

src/segments/cf_target.go
wozjac: Noted, I'll adjust.

wozjac: Corrected.

src/segments/cf_target.go
wozjac: Corrected.

PR #1863, opened 2022-03-03
theme: add takuya

vedantmgoyal9:
### Prerequisites

### Description

Adding a new theme

### DIFF
--- a/takuya.omp.json
+++ b/takuya.omp.json
@@ -1,0 +1,127 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "final_space": false,
+  "osc99": true,
+  "blocks": [
+    {
+      "type": "prompt",
+      "alignment": "left",
+      "segments": [
+        {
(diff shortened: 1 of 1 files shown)


vedantmgoyal9:
@jandedobbeleer can you add an update function that can notify users when there's an update available (for major versions since minor versions are iterated frequently). Here's an example from GitHub's CLI:

image

Or maybe `oh-my-posh.exe -update` which could auto-update the binary itself (this would upgrade to minor and patch versions too)

JanDeDobbeleer:
>  can you add an update function that can notify users when there's an update available 

@vedantmgoyal2009 yes, been thinking about that before. What we could do is print this on a weekly basis, so that we do not annoy people x times a day. The autoupdate has 1 caveat and the main reason why I don't do it yet, it will make the PowerShell module code and included themes out of sync.

JanDeDobbeleer:
@vedantmgoyal2009 I removed this commit from main as I too late noticed the theme doesn't follow the version 1 spec. You need to update your oh-my-posh and use this theme, it will automatically create an updated version which you can then submit.

PR #1870, opened 2022-03-06
feat: cds segment

wozjac:
### Prerequisites

### Description

A segment for cds command from [SAP Cloud Application Programming Model - SAP CAP](https://cap.cloud.sap/docs/tools/#command-line-interface-cli). 
In the project root directory you can have .cdsrc.json (or .cdsrc-private.json) file, but it is not mandatory, so package.json dependency is also checked. *.cds files are located in `db', `srv` subdirectories (usually scaffolded with `cds init`).

Best Regards
Jacek

### DIFF
--- a/docs/docs/segment-cds.md
+++ b/docs/docs/segment-cds.md
@@ -1,0 +1,59 @@
+---
+id: cds
+title: CDS (SAP CAP)
+sidebar_label: CDS
+---
+
+## What
+
+Display the active [CDS CLI][sap-cap-cds] version.
+
+## Sample Configuration
+
+```json
+{
+  "background": "#a7cae1",
+  "foreground": "#100e23",
+  "powerline_symbol": "\ue0b0",
+  "properties": {
+    "template": " \ue311 cds {{ .Full }} "
+  },
+  "style": "powerline",
+  "type": "cds"
+}
+```
+
+## Properties
+
+- home_enabled: `boolean` - display the segment in the HOME folder or not - defaults to `false`
+- display_error: `boolean` - show the error context when failing to retrieve the version information - defaults to `true`
+- missing_command_text: `string` - text to display when the cds command is missing - defaults to empty
+- display_mode: `string` - determines when the segment is displayed
+  - `always`: the segment is always displayed
+  - `files`: the segment is displayed when `.cdsrc.json`, `.cdsrc-private` or `*.cds` file is present
+  - `context`: (default) the segment is displayed when conditions from `files` mode are fulfilled or
+    `package.json` file is present and `@sap/cds` is in `dependencies` section
+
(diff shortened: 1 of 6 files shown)


src/segments/cds.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"encoding/json"
+	"oh-my-posh/environment"
+	"oh-my-posh/properties"
+)
+
+type Cds struct {
+	language
+	HasCDSDependency bool
```
JanDeDobbeleer: As the context is already CDS, does it makes sense to rename this to `HasDependency`?

src/segments/cds.go
```diff
@@ -0,0 +1,68 @@
...
+	dependencies := map[string]json.RawMessage{}
+
+	if err := json.Unmarshal(objmap["dependencies"], &dependencies); err != nil {
+		return
+	}
+
+	for d := range dependencies {
+		if d == "@sap/cds" {
+			c.HasCDSDependency = true
+			return
```
JanDeDobbeleer: should this be `break`?

JanDeDobbeleer (changes requested):

src/segments/cds.go
wozjac: sounds ok

src/segments/cds.go
wozjac: will correct

src/segments/cds.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"encoding/json"
+	"oh-my-posh/environment"
+	"oh-my-posh/properties"
+)
+
+type Cds struct {
+	language
+	HasDependency bool
```
JanDeDobbeleer, line 11: This isn't present in the docs template properties it seems

wozjac: Could you point to similar case in existing segments? Not sure if I understand correctly.

docs/docs/segment-cds.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Template Properties
+
+- `.Full`: `string` - the full version
+- `.Major`: `string` - major number
+- `.Minor`: `string` - minor number
+- `.Patch`: `string` - patch number
+- `.Prerelease`: `string` - prerelease info text
+- `.BuildMetadata`: `string` - build metadata
+- `.Error`: `string` - when fetching the version string errors
```
JanDeDobbeleer, line 55: @wozjac I would expect to see `HasDependency` also listed here.

src/segments/cds.go
JanDeDobbeleer: Added a comment above.

JanDeDobbeleer (approved):

PR #1890, opened 2022-03-10
Update contributing-segment.md

estruyf:
Seems that `package main` needs to be updated to `package segments` otherwise it doesn't build.


JanDeDobbeleer:
@estruyf I'm helping out with the commit messages (and added your other question's solution too).

estruyf:
@JanDeDobbeleer 🤓 - I knew you were going to mention that, but used the `Edit this page` feature.

PR #1901, opened 2022-03-11
feat: package segment

estruyf:
### Prerequisites

### Description

New package segment to show the version of the package file. Created in a generic way so that other types can be added in the future.

image

### DIFF
--- a/themes/schema.json
+++ b/themes/schema.json
@@ -199,7 +199,8 @@
             "ui5tooling",
             "kotlin",
             "swift",
-            "npm"
+            "npm",
+            "project"
           ]
         },
         "style": {
@@ -265,6 +266,17 @@
         }
       },
       "allOf": [
+        {
+          "if": {
+            "properties": {
+              "type": { "const": "project" }
+            }
+          },
+          "then": {
+            "title": "Project Segment",
+            "description": "https://ohmyposh.dev/docs/project"
+          }
+        },
         {
           "if": {
             "properties": {
--- a/docs/docs/segment-project.md
+++ b/docs/docs/segment-project.md
@@ -1,0 +1,44 @@
+---
+id: project
+title: Project
+sidebar_label: Project
+---
+
+## What
+
+Display the current version of your project defined in the package file.
+
+Supports:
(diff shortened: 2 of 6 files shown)


src/segments/package.go
```diff
@@ -0,0 +1,81 @@
...
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Version     string `json:"version"`
+}
+
+type Package struct {
+	props properties.Properties
+	env   environment.Environment
+
+	packages []PackageItem
```
JanDeDobbeleer: you can make this a pointer array so you always pass by reference:

```go
packages []*PackageItem
```

themes/schema.json
```diff
@@ -265,6 +266,17 @@
...
       },
       "allOf": [
+        {
+          "if": {
+            "properties": {
+              "type": { "const": "package" }
+            }
+          },
+          "then": {
+            "title": "NPM Segment",
```
JanDeDobbeleer: Wrong title :-)

JanDeDobbeleer (changes requested):
I love the extensibility, but wonder if the name is correct in that case. In the end what it does is provide the version number of the project, right?

src/segments/package.go
estruyf: Need to get more familiar with Go it seems. 🤓

JanDeDobbeleer: That's why I'm here :-)

estruyf: Should have known, as, in the test, I got a warning to use the pointer notation `*` because the object reference was otherwise too big.

estruyf:
Yes, got the inspiration from Starship, there it's also called "package", but it can be anything else. 

On some cases, the project is the actual package version, but still,  it's a project. Shall we go for "project", leave it, or do you have another suggestion?

JanDeDobbeleer:
Project makes sense 👍🏻

estruyf:
Updated the files and code to use `project` instead.

JanDeDobbeleer (approved):

PR #1930, opened 2022-03-15
refactor(cli): isolate debug command

JanDeDobbeleer:
### Prerequisites

### Description


JanDeDobbeleer:
@TylerLeonhardt I have a question regarding vscode and the change in https://github.com/PowerShell/PowerShellEditorServices/pull/627. it seems this logic is still in there, although that code path no longer exists and I was wondering if there's a better to handle this by detecting if there's custom prompt present (and thus not do that).

When the custom prompt has `[DBG]`:

image

When it doesn't have `[DBG]`:

image

TylerLeonhardt:
It's been a while since I looked at this stuff. @andschwa or @SeeminglyScience would probably be better folks to chat with

PR #1939, opened 2022-03-18
add a new theme "almeida.omp.json"

codigoperfeito:
Another theme was added to the themes folder, with the name of "Almeida.omp.json" I hope it's within the compliant, it's my personal theme.

this is the img of theme

img

### DIFF
--- a/themes/almeida.omp.json
+++ b/themes/almeida.omp.json
@@ -1,0 +1,143 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "blocks": [
+    {
+      "alignment": "left",
+      "segments": [
+        {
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@codigoperfeito you will need to migrate this to version 2. Update oh-my-posh and use `oh-my-posh config migrate --config almeida.oh-my-posh.json --write`

JanDeDobbeleer:
@codigoperfeito this theme is very similar to others we already have and I'm trying to reduce the amount of identical themes in the stack. I might need to create a separate repo for this specifically, as it's getting a bit out of hand.

codigoperfeito:
> @codigoperfeito this theme is very similar to others we already have and I'm trying to reduce the amount of identical themes in the stack. I might need to create a separate repo for this specifically, as it's getting a bit out of hand.

Ok !

Issue #1766, opened 2022-02-16
OMP rendering generic linux icon instead of ubuntu icon in WSL Ubuntu 20.04


devhawk:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

Screenshot: 
![image](https://user-images.githubusercontent.com/8965/154342828-c2e984f2-2c69-402e-a411-7c93c0b6398b.png)

### Theme

Relevant Segments: 

```json
        {
          "background": "#4082ce",
          "background_templates": [
            "{{if contains \"Ubuntu\" .Env.WSL_DISTRO_NAME}}#dd4814{{end}}",
            "{{if contains \"Debian\" .Env.WSL_DISTRO_NAME}}#D70A53{{end}}",
          ],
          "foreground": "#ffffff",
          "leading_diamond": "\ue0b6",
          "properties": {
            "template": "{{.Icon}}",
            "windows": "\uf17a"
          },
          "style": "diamond",
          "type": "os"
        },
        {
          "background": "#4082ce",
          "background_templates": [
            "{{if contains \"Ubuntu\" .Env.WSL_DISTRO_NAME}}#dd4814{{end}}",
            "{{if contains \"Debian\" .Env.WSL_DISTRO_NAME}}#D70A53{{end}}",
          ],
          "foreground": "#ffffff",
          "properties": {
            "template": " {{ .Env.WSL_DISTRO_NAME }}"
          },
          "style": "diamond",
          "trailing_diamond": "\ue0b4",
          "type": "text"
        },
```

### What OS are you seeing the problem on?

Linux

### Which shell are you using?

bash

### Log output

```shell
[1mVersion:[0m 7.18.0

[1mSegments:[0m

ConsoleTitle(true) -   0 ms - ]0;~ / Ubuntu-20.04
path(true)     -   1 ms - [38;2;49;57;127m[0m[48;2;49;57;127m[38;2;255;255;255m ~ [0m[38;2;49;57;127m[0m
battery(false) -   0 ms - 
git(false)     -   0 ms - 
node(false)    -   0 ms - 
dotnet(false)  -   0 ms - 
rust(false)    -   0 ms - 
python(false)  -   0 ms - 
go(false)      -   0 ms - 
os(true)       -   0 ms - [38;2;221;72;20m[0m[48;2;221;72;20m[38;2;255;255;255m[0m
[...]

PR #1944, opened 2022-03-18
fix(wsl): get correct distro name

lewis-yeung:
### Prerequisites

### Description

Store the correct distro name of WSL into `omp.cache` so that in the "OS" segment, an icon of WSL distro will be rendered instead of a generic one. Related issue: #1766.

### DIFF
--- a/src/environment/unix.go
+++ b/src/environment/unix.go
@@ -65,8 +65,8 @@ func (env *ShellEnvironment) Platform() string {
 		env.Cache().Set(key, platform, -1)
 	}()
 	if wsl := env.Getenv("WSL_DISTRO_NAME"); len(wsl) != 0 {
-		platform = strings.ToLower(wsl)
-		return strings.Split(platform, "-")[0]
+		platform = strings.Split(strings.ToLower(wsl), "-")[0]
+		return platform
 	}
 	platform, _, _, _ = host.PlatformInformation()
 	if platform == "arch" {


src/environment/unix.go
```diff
@@ -65,15 +65,15 @@ func (env *ShellEnvironment) Platform() string {
...
-		platform = strings.ToLower(wsl)
-		return strings.Split(platform, "-")[0]
-	}
-	platform, _, _, _ = host.PlatformInformation()
-	if platform == "arch" {
-		// validate for Manjaro
-		lsbInfo := env.FileContent("/etc/lsb-release")
-		if strings.Contains(strings.ToLower(lsbInfo), "manjaro") {
-			platform = "manjaro"
+		platform = strings.Split(strings.ToLower(wsl), "-")[0]
```
JanDeDobbeleer, line 68: Can you do an early return here rather than use an else statement?

JanDeDobbeleer (changes requested):

src/environment/unix.go
lewis-yeung: @JanDeDobbeleer Updated.

JanDeDobbeleer:
@lewis-yeung your commit message paragraph is too long ;-)

> Store the correct distro name of WSL into `omp.cache` so that in the "OS" segment, an icon of WSL distro will be rendered instead of a generic one. Related issue: #1766.
✖   footer's lines must not be longer than 100 characters [footer-max-line-length]

lewis-yeung:
@JanDeDobbeleer Changed.

PR #1995, opened 2022-03-27
Bugfix for themes schema.json definition reference

atakiya:
### Prerequisites

### Description

As the commit message states, ee3b1127dec3abd9edbc4c8aceea38000e7fcd10 moved `#/definitions/templates` to `#/definitions/segment/properties/template`, but seemingly one reference has been forgotten.
This PR simply updates the definition path for the reference under `#/definitions/color_templates/items`

Still a bit new to JSON schemas & I'm not sure if this templating schema is wanted for that property, as the URLs for `foreground_templates` and `background_templates` are non-existing, so I hope this is correct.

### DIFF
--- a/themes/schema.json
+++ b/themes/schema.json
@@ -30,7 +30,7 @@
       "description": "https://ohmyposh.dev/docs/config-overview#foreground-templates",
       "default": [],
       "items": {
-        "$ref": "#/definitions/template"
+        "$ref": "#/definitions/segment/properties/template"
       }
     },
     "fetch_version": {


JanDeDobbeleer:
@atakiya commitlint will fail as it doesn't accept upper case sentences. But you are correct with the change.

atakiya:
> @atakiya commitlint will fail as it doesn't accept upper case sentences. But you are correct with the change.

Updated, removed commit body as it was apparently also above the limit.

JanDeDobbeleer:
@all-contributors please add @atakiya for code

PR #2013, opened 2022-03-31
feat: enable posh-git on $GitPromptSettings.EnablePromptStatus

JanDeDobbeleer:
### Prerequisites

### Description

relates to https://github.com/dahlbyk/posh-git/issues/883

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -57,11 +57,9 @@ function global:Get-PoshContext {
 }
 
 function global:Initialize-ModuleSupport {
-    if ($env:POSH_GIT_ENABLED -eq $true -and (Get-Module -Name "posh-git")) {
-        [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification = 'Variable used later (not in this scope)')]
+    if ($GitPromptSettings.EnablePromptStatus) {
         $global:GitStatus = Get-GitStatus
-        [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification = 'Variable used by posh-git (not in this script)')]
-        $GitPromptSettings = $null
+        $GitPromptSettings.EnablePromptStatus = $false
         $env:POSH_GIT_STATUS = Write-GitStatus -Status $global:GitStatus
     }
 }
--- a/docs/docs/segment-posh-git.mdx
+++ b/docs/docs/segment-posh-git.mdx
@@ -10,7 +10,6 @@ Display the [posh-git][posh-git] prompt.
 
 :::caution
 This segment only works within Powershell and requires the posh-git module to be installed and imported.
-To enable the `posh-git` module, set `$env:POSH_GIT_ENABLED = $true` in your `$PROFILE`.
 :::
 
 ## Sample Configuration


JanDeDobbeleer:
@dahlbyk this would work. BTW, I force the prompt to disabled as otherwise it still displays, meaning we can be in full control (and still use the [posh-git](https://ohmyposh.dev/docs/poshgit) segment in oh-my-posh)

src/shell/scripts/omp.ps1
```diff
@@ -57,11 +57,9 @@ function global:Get-PoshContext {
 }
 
 function global:Initialize-ModuleSupport {
-    if ($env:POSH_GIT_ENABLED -eq $true -and (Get-Module -Name "posh-git")) {
-        [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification = 'Variable used later (not in this scope)')]
+    if ($GitPromptSettings.EnablePromptStatus) {
         $global:GitStatus = Get-GitStatus
-        [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSProvideCommentHelp', '', Justification = 'Variable used by posh-git (not in this script)')]
-        $GitPromptSettings = $null
+        $GitPromptSettings.EnablePromptStatus = $false
```
dahlbyk, line 62: Won't disabling this prevent all but the first `Get-GitStatus` call from returning anything?
https://github.com/dahlbyk/posh-git/blob/dc2b6367324a9db79db55186375462a454f29c2e/src/GitUtils.ps1#L247-L248

Clearing `$GitPromptSettings` will have worked differently because `Get-GitStatus` initializes default `$settings` (which would have wiped out the ability to customize `Write-GitStatus` via `$GitPromptSettings`):
https://github.com/dahlbyk/posh-git/blob/dc2b6367324a9db79db55186375462a454f29c2e/src/GitUtils.ps1#L245

JanDeDobbeleer: oooooooooow, I made a mistake doing this in between things. I'll fix that.

JanDeDobbeleer: @dahlbyk the intent was to disable the posh-git prompt from showing up. It still did that. How do I disable only that?

dahlbyk: What do you mean when you say the posh-git prompt is showing up? The `$GitPromptScriptBlock` we replace the default `prompt` with does a lot of things, including calling `Write-VcsStatus` which calls `Write-GitStatus` via `$global:VcsPromptStatuses`.

So you could set `$global:VcsPromptStatuses = @()` and the Git status should go away, but that leaves the rest of the `prompt` running. I guess I'm unclear on the relationship between OMP and the actual PS `prompt` function.

JanDeDobbeleer: We [override the prompt](https://github.com/JanDeDobbeleer/oh-my-posh/blob/749eb6cf263fd960644b9d099d973c7c94af4958/src/shell/scripts/omp.ps1#L67) just like anyone does in PowerShell. I noticed that on certain conditions, posh-git prepends itself to it.

JanDeDobbeleer: @dahlbyk basically, what I would love to achieve is the [following](https://github.com/dahlbyk/posh-git/blob/cb070b777363ab5d51c6dff1ead22828e22122b2/src/posh-git.psm1#L141) to not resolve to true. Would it imply I need to advise people to initialise oh-my-posh before `Import-Module posh-git`?

dahlbyk: > Would it imply I need to advise people to initialise oh-my-posh before `Import-Module posh-git`?

That should always work. We're supposed to do nothing if `prompt` has been changed. Conversely, we shouldn't object to OMP overwriting the posh-git `prompt` afterwards, either.

> I noticed that on certain conditions, posh-git prepends itself to it.

If you see that again, please check `$GitPromptSettings.AnsiConsole`. If we don't detect an ANSI-capable host, we use `Write-Host` instead of returning a string of the prompt.

Issue #2014, opened 2022-03-31
Add support for other project files


hadrizi:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

It would be nice if [`Project` segment](https://ohmyposh.dev/docs/project) supported other project files(like `pyproject.toml` or `Cargo.toml`)

JanDeDobbeleer:
@hadrizi can you give some example files, because yes, that's the intention.

PR #2015, opened 2022-03-31
feat: cargo and poetry project packages support

hadrizi:
### Prerequisites

### Description
see #2014 

cc @JanDeDobbeleer

### DIFF
--- a/src/segments/project.go
+++ b/src/segments/project.go
@@ -14,20 +14,14 @@ type ProjectItem struct {
 	Fetcher func(item ProjectItem) (string, string)
 }
 
-//Node.js Package
-type NodePackageJSON struct {
-	Name    string `json:"name"`
-	Version string `json:"version"`
+type ProjectData struct {
+	Version string
+	Name    string
 }
 
 // Rust Cargo package
 type CargoTOML struct {
-	Package CargoPackageTOML
-}
-
-type CargoPackageTOML struct {
-	Name    string
-	Version string
+	Package ProjectData
 }
 
 // Python Poetrey package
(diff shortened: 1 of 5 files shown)


src/segments/project.go
```diff
@@ -24,6 +50,7 @@ type Project struct {
 
 	projects []*ProjectItem
 	Version  string
+	Name     string
```
JanDeDobbeleer: I would create 1 type called ProjectData that contains these fields and is added using struct inheritance:

```go
type ProjectData struct {
    Name    string
    Version string
}

type Project struct {
    ...
 	projects []*ProjectItem
    ProjectData
```

You can then drop all project specific types, they don't really mean anything outside

JanDeDobbeleer (changes requested):

src/segments/project.go
hadrizi: done

src/segments/project.go
```diff
@@ -4,27 +4,43 @@ import (
...
+	Version string
+	Name    string
+}
+
+// Rust Cargo package
+type CargoTOML struct {
+	Package ProjectData
+}
+
+// Python Poetrey package
```
JanDeDobbeleer: Typo -> `Poetry`

src/segments/project.go
```diff
@@ -4,27 +4,43 @@ import (
...
+type PyprojectTOML struct {
```
JanDeDobbeleer: I would respect the Pascal Casing: `PyProjectTOML`

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@hadrizi I will address the commitlint issue ;-)

src/segments/project.go
hadrizi: done

src/segments/project.go
hadrizi: done

JanDeDobbeleer (approved):

Issue #2033, opened 2022-04-05
Segment Project Error


neradp:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

Segment throws unexpected end of JSON input  on windows and pwsh or powershell.

Write-PoshDebug freezes powershell prompt.

### Theme

```json
{
  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
  "blocks": [
	
    {
      "alignment": "left",
      "segments": [
    {
          "background": "#2f2f2f",
          "foreground": "#26C6DA",
          
          "properties": {
            "alpine": "\uf300",
            "arch": "\uf303",
            "centos": "\uf304",
            "debian": "\uf306",
            "elementary": "\uf309",
            "fedora": "\uf30a",
            "gentoo": "\uf30d",
            "linux": "\ue712",
            "macos": "\ue711",
            "manjaro": "\uf312",
            "mint": "\uf30f",
            "opensuse": "\uf314",
            "raspbian": "\uf315",
            "ubuntu": "\uf31c",
            "windows": "\ue70f"
          },
          "style": "diamond",
		  "leading_diamond": "\ue0c7",
          "template": " {{ if .WSL }}WSL at {{ end }}{{.Icon}}<#7a7a7a> \ue0b1</>",
          "type": "os"
        },
        {
          "background": "#2f2f2f",
          "foreground": "#fafafa",
          "properties": {
            "display_default": false
          },
          "style": "diamond",
          "template": " {{ if .SSHSession }}\uf817 {{ end }}{{ if .Root }}\uf0e7 {{ end }}<#77f5d6>{{ .UserName }}</><#7a7a7a>\uf1fa</><#2EEFBF>{{ .HostName }}</><#7a7a7a> \ue0b1</>",
          "type": "session"
        },
		
        
        {
          "background": "#2f2f2f",
          "foreground": "#fafafa",
          "properties": {
            "folder_icon": "<#B5B2C2>\ue5fe </>",
            "folder_separator_icon": "<#B5B2C2>\ue0bf </>",
            "home_icon": "\uf015",
            "style": "agnoster_short"
          },
          "style": "powerline",
		  "powerline_symbol": "\uE0B0",
		  
          "template": "<#f2f200> \ue5ff </>{{ .Path }} ",
          "type": "path"
        },
        {
          "background": "#2f2f2f",
[...]

PR #2043, opened 2022-04-06
fix(project): fully qualified project path

lnu:
fix an issue with pwsh reading the file from users home instead of cwd

### Prerequisites

### Description

https://github.com/JanDeDobbeleer/oh-my-posh/issues/2033

### DIFF
--- a/src/environment/shell.go
+++ b/src/environment/shell.go
@@ -397,6 +397,9 @@ func (env *ShellEnvironment) HasFolder(folder string) bool {
 
 func (env *ShellEnvironment) FileContent(file string) string {
 	defer env.trace(time.Now(), "FileContent", file)
+	if !filepath.IsAbs(file) {
+		file = filepath.Join(env.Pwd(), file)
+	}
 	content, err := ioutil.ReadFile(file)
 	if err != nil {
 		env.log(Error, "FileContent", err.Error())
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -12,6 +12,7 @@ function global:Start-Utf8Process
 	$StartInfo.FileName = $filename
 	$StartInfo.Arguments = $Arguments
     $StartInfo.UseShellExecute = $false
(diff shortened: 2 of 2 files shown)


JanDeDobbeleer (approved):
Just a test to be fixed by adding the mock and we're good to go. Thanks, I didn't understand why this was happening tbh.

JanDeDobbeleer:
@lnu ~~should we add that logic (prepending the path) to the `FileContent` function instead? As I don't think we ever use it to fetch content other than the current folder.~~

We actually do, so it becomes the responsibility of the caller. Unless this could work:

```diff
diff --git a/src/environment/shell.go b/src/environment/shell.go
index 6e895421..a6b1ffe7 100644
--- a/src/environment/shell.go
+++ b/src/environment/shell.go
@@ -397,6 +397,9 @@ func (env *ShellEnvironment) HasFolder(folder string) bool {
 
 func (env *ShellEnvironment) FileContent(file string) string {
 	defer env.trace(time.Now(), "FileContent", file)
+	if !filepath.IsAbs(file) {
+		file = filepath.Join(env.Pwd(), file)
+	}
 	content, err := ioutil.ReadFile(file)
 	if err != nil {
 		env.log(Error, "FileContent", err.Error())
```

UPDATE: the above works like a charm

JanDeDobbeleer (changes requested):
Let's try with the proposal I added, can validate later today.

lnu:
Did the proposed change. Works on my side too.

lnu:
It comes from the utf8 process startup refactoring. Since the refactoring, the startup folder is set to the home of the user and not the current path. Tried with an older version of the code and now it works.

lnu:
should be fixed with the second commit

lnu:
We had other issues in the past with some segments relying on the PWD parameter and others relying on the current working directory. There are probably other cases like this one hidden somewhere.

JanDeDobbeleer:
@lnu indeed. This has been a recurring topic. And PowerShell is even more difficult due to PSWD and potential issues with drives.

JanDeDobbeleer (approved):

PR #2063, opened 2022-04-09
feat(python): support pyenv .python-version files

bewing:
Pyenv will use .python-version files up the file heirarchy when
determining the active Python version or virtual environment to use
based on which folder your shell is in.  This change will read the file,
and if the first line does not look like a Python version, set the Venv
to that string for display.

This feature is gated behind the `use_python_version_file` Property

### Prerequisites

### Description
Read .python-version files to determine virtualenv if environment variable search turns up empty.
Note:  Does not read $PYENV_ROOT/version -- should we?
Question:  Would it be better to just parse `pyenv version` directly to determine the first environment?
Finally:  I'm using regex to try to not display a global python version as the python virtualenv -- my first thought was to compare the returned string to `python.language.version.Full`, but apparently that's not populated when we're detecting the Venv.

### DIFF
--- a/src/segments/python.go
+++ b/src/segments/python.go
@@ -3,6 +3,8 @@ package segments
 import (
 	"oh-my-posh/environment"
 	"oh-my-posh/properties"
+	"oh-my-posh/regex"
+	"strings"
 )
 
 type Python struct {
@@ -13,7 +15,8 @@ type Python struct {
 
 const (
 	// FetchVirtualEnv fetches the virtual env
-	FetchVirtualEnv properties.Property = "fetch_virtual_env"
+	FetchVirtualEnv      properties.Property = "fetch_virtual_env"
+	UsePythonVersionFile properties.Property = "use_python_version_file"
 )
 
 func (p *Python) Template() string {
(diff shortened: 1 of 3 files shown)


src/segments/python.go
```diff
@@ -68,6 +70,13 @@ func (p *Python) loadContext() {
 			break
 		}
 	}
+	if f, err := p.language.env.HasParentFilePath(".python-version"); err == nil {
+		re := regexp.MustCompile("[0-9]+.[0-9]+.[0-9]+") // Don't display version twice
```
JanDeDobbeleer: Have a look at our regex functionality, it caches everything for you.

JanDeDobbeleer:
@bewing I can do the changes if you don't have time, would love to merge this one.

bewing:
> @bewing I can do the changes if you don't have time, would love to merge this one.

I think this change is useful enough on its own -- I will add a config gate and documentation today.

JanDeDobbeleer (changes requested):

src/segments/python.go
JanDeDobbeleer: @bewing I had this pending request I did not submit 🤦🏻

JanDeDobbeleer (approved):

PR #2071, opened 2022-04-11
docs(homebrew): adjust details about installation

lewis-yeung:
### Prerequisites

### Description

This may help to prevent other Homebrew users from getting the same confusion as I got in #2023. ~~The Way **#B** to install OMP could be more friendly to non-Gophers.~~

**P.S.** I had ever thought that _**homebrew-oh-my-posh**_ could provide a way of installing from binary releases (e.g., [go-task](https://github.com/go-task/homebrew-tap/blob/master/Formula/go-task.rb), which uses [GoReleaser](https://github.com/goreleaser/goreleaser)) since this could be more friendly to non-Gophers. However, _**homebrew-core**_ already provides the way of installing from bottles, so now I think it appears to be non-essential.

### DIFF
--- a/docs/docs/install-homebrew.mdx
+++ b/docs/docs/install-homebrew.mdx
@@ -2,12 +2,12 @@ A [Homebrew][brew] formula is available for easy installation. When installing H
 
 ```bash
 brew tap jandedobbeleer/oh-my-posh
-brew install oh-my-posh
+brew install jandedobbeleer/oh-my-posh/oh-my-posh
 ```
 
 This installs two things:
 
-- `oh-my-posh` - Executable, added to `$(brew --prefix)`
+- `oh-my-posh` - Executable, added to `$(brew --prefix)/bin`
 - `themes` - The latest Oh My Posh [themes][themes]
 
 If you want to use a standard theme, you can find them in `$(brew --prefix)/oh-my-posh/themes`, referencing them as such


docs/docs/install-homebrew.mdx
```diff
@@ -1,17 +1,25 @@
 A [Homebrew][brew] formula is available for easy installation. When installing Homebrew for Linux, be sure to follow *[Next steps][nextsteps]* instructions to add Homebrew to your PATH and to your bash shell profile script.
 
+There are two different ways to install Oh My Posh via Homebrew:
+
+- **#A**
+
 ```bash
-brew tap jandedobbeleer/oh-my-posh
-brew install oh-my-posh
+brew install jandedobbeleer/oh-my-posh/oh-my-posh
```
JanDeDobbeleer, line 5: I would only add this line, 2 ways of installing where one is one that I don't maintain is not something I wish to support ;-). The fact that it also installs golang is irrelevant in my book as that's how brew works. That is completely unrelated to oh-my-posh and goes for any installation.

JanDeDobbeleer (changes requested):
see in-line comment

docs/docs/install-homebrew.mdx
lewis-yeung: @JanDeDobbeleer Got your thought. Removed others while the executable path should be `$(brew --prefix)/bin` indeed.

lewis-yeung:
@all-contributors Please add @lewis-yeung for code and doc.

PR #2122, opened 2022-04-21
feat(python): real pyenv support

bewing:
If `python` points at a pyenv shim, use `pyenv version-name` to
determine Python version and Venv information.

### Prerequisites

### DIFF
--- a/src/environment/shell.go
+++ b/src/environment/shell.go
@@ -139,7 +139,9 @@ type Environment interface {
 	HasFolder(folder string) bool
 	HasParentFilePath(path string) (fileInfo *FileInfo, err error)
 	HasFileInParentDirs(pattern string, depth uint) bool
+	ResolveSymlink(path string) (string, error)
 	DirMatchesOneOf(dir string, regexes []string) bool
+	CommandPath(command string) string
 	HasCommand(command string) bool
 	FileContent(file string) string
 	LsDir(path string) []fs.DirEntry
@@ -401,6 +403,10 @@ func (env *ShellEnvironment) HasFolder(folder string) bool {
 	return hasFolder
 }
 
+func (env *ShellEnvironment) ResolveSymlink(path string) (string, error) {
+	return filepath.EvalSymlinks(path)
+}
+
 func (env *ShellEnvironment) FileContent(file string) string {
 	defer env.trace(time.Now(), "FileContent", file)
 	if !filepath.IsAbs(file) {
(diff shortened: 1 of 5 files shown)


bewing:
This approach shaves about 100msec off python prompt generation on my WSL environment (since we never run `python` directly, and `pyenv` is 100% bash)

debug log for the segment:
```
python(true)         -  51 ms -   virt 3.9.9
2022/04/21 14:05:05 debug: Pwd
/home/bewing/python
2022/04/21 14:05:05 Pwd duration: 924ns, args:
2022/04/21 14:05:05 debug: Getenv

2022/04/21 14:05:05 Getenv duration: 1.322µs, args: VIRTUAL_ENV
2022/04/21 14:05:05 debug: Getenv

2022/04/21 14:05:05 Getenv duration: 1.161µs, args: CONDA_ENV_PATH
2022/04/21 14:05:05 debug: Getenv

2022/04/21 14:05:05 Getenv duration: 1.096µs, args: CONDA_DEFAULT_ENV
2022/04/21 14:05:05 debug: Pwd
/home/bewing/python
2022/04/21 14:05:05 Pwd duration: 915ns, args:
2022/04/21 14:05:05 PathSeparator duration: 44ns, args:
2022/04/21 14:05:05 debug: HasFiles
true
2022/04/21 14:05:05 HasFiles duration: 16.68µs, args: *.py
2022/04/21 14:05:05 debug: RunCommand
virt
2022/04/21 14:05:05 RunCommand duration: 31.667596ms, args: pyenv version-name
2022/04/21 14:05:05 TemplateCache duration: 763ns, args:
2022/04/21 14:05:05 TemplateCache duration: 481ns, args:

**Comment by bewing** [comment]
Need to refactor this code to make it mockable/testable.  Changes to `environment` probably needed

File: src/environment/shell.go [hunk 1]
```diff
@@ -500,22 +501,29 @@ func (env *ShellEnvironment) RunShellCommand(shell, command string) string {
 	return ""
 }
 
-func (env *ShellEnvironment) HasCommand(command string) bool {
+func (env *ShellEnvironment) HasCommandPath(command string) string {
```
JanDeDobbeleer: This implies a boolean return but isn't, can we rename this to `CommandPath`?

src/segments/python.go
```diff
@@ -101,3 +95,41 @@ func (p *Python) canUseVenvName(name string) bool {
...
 	return true
 }
+
+func (p *Python) pyenvVersion() (string, error) {
+	// Use `pyenv root` instead of $PYENV_ROOT?
+	// Is our Python executable at $PYENV_ROOT/bin/python ?
+	// Should p.env expose command paths?
+	path := p.env.HasCommandPath("python")
+	// TODO:  pyenv-win has this at $PYENV_ROOT/pyenv-win/shims
+	if path == filepath.Join(p.env.Getenv("PYENV_ROOT"), "shims", "python") {
```
JanDeDobbeleer: Reverse the statement so we can return early rather than indent this entirely?

JanDeDobbeleer (changes requested):

src/segments/python.go
bewing: inverted.

src/environment/shell.go
bewing: renamed

bewing:
Think code and tests are good now.  Will look at docs next.  Do we want to feature gate this, or no?

JanDeDobbeleer:
@bewing I wouldn't gate it, just ship. If you tested locally, we'll find out just how representative it was 😃

bewing:
Bah, I thought I was testing locally, but I wasn't.  Not ready for review yet, but I'm not sure how to mark it as draft again

bewing:
Found the error in my test case, fixed it.  Really really think it's ready now.

src/environment/shell.go
```diff
@@ -401,6 +403,10 @@ func (env *ShellEnvironment) HasFolder(folder string) bool {
 	return hasFolder
 }
 
+func (env *ShellEnvironment) RealPath(path string) (string, error) {
```
JanDeDobbeleer: I would rename this to `ResolveSymlink` as that what this does.

JanDeDobbeleer (changes requested):

src/environment/shell.go
bewing: renamed

JanDeDobbeleer:
Thanks for the changes! 🚀

PR #2129, opened 2022-04-22
Align tooltip behaviors

lewis-yeung:
### Prerequisites

### Description

Hi, Jan. Currently, tooltips behave differently in different shells. We should give a consideration on the cross-shell consistency.

<details><summary>ℹ️ Info</summary>

AFAIK:

- **In zsh:**
	- Tooltips are implemented by user-defined widgets in ZLE.
	- **WHATEVER** single character input can trigger tooltip rendering.
	- A tip-check uses the **ENTIRE** command line buffer.

- **In PowerShell:**
	- Tooltips are implemented by `Set-PSReadLineKeyHandler`.
	- Only a **SPACE** character input can trigger tooltip rendering.
	- A tip-check uses only the **FIRST WORD** of command line buffer.

I have not tested fish or cmd yet and I know little about fish or Lua scripts, but via source code, I got these keys:

- **In fish:**
	- Only a **SPACE** character input can trigger tooltip rendering.
	- A tip-check uses the **ENTIRE** command line buffer.

- **In cmd:**
	- Only a **SPACE** character input can trigger tooltip rendering.
	- A tip-check uses only the **FIRST WORD** of command line buffer.

</details>

I have some ideas about it.

<details><summary>💡 Idea 1</summary>

I noticed the description in the [documentation](https://ohmyposh.dev/docs/configuration/tooltips#configuration):

> This configuration will render a right-aligned git segment when you type `git` or `g` followed by a **space**.

However, in zsh whatever single character input will trigger a tip-check on the buffer. This is unreasonable in certain cases. E.g., someone uses `g` as the tip of a `git` tooltip and `go` for a `go` one, when the user wants to get a `go` tooltip rendered in a Git repo, a `g` must be typed first, and then `o`, while at this point the `git` tooltip gets rendered, which is obviously superfluous for the user, especially in a large-size repo this may lead to an unnecessary slowdown. So it's better to make it the same as that in other shells.

I have got it done in this draft. ✅

</details>

<details><summary>💡 Idea 2</summary>

If a tip-check uses the entire command line buffer, it supports a tip including interpolated spaces, e.g., `g s` for a `git` tooltip. I think this can be applied to all these 4 shells to improve flexibility and ensure consistency.

I have got it done in PowerShell. Lacking knowledge about Lua scripts, I'm not able to adjust that in cmd for the time being but it may be done later on. Or, would be so grateful if you adopt this and get the rest work done. 🙏

</details>

OMP is an awesome project. Thanks for your hard work!

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -20,21 +20,40 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
     function Start-Utf8Process {
         param(
             [string] $FileName,
-            [string] $Arguments
+            [string[]] $Arguments = @()
         )
 
         $Process = New-Object System.Diagnostics.Process
         $StartInfo = $Process.StartInfo
+        $StartInfo.FileName = $FileName
+        if ($StartInfo.ArgumentList.Add) {
+            # ArgumentList is supported in PowerShell 6.1 and later (built on .NET Core 2.1+)
+            # ref-1: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-6.0
+            # ref-2: https://docs.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.2#net-framework-vs-net-core
+            $Arguments | ForEach-Object -Process { $StartInfo.ArgumentList.Add($_) }
+        } else {
+            # escape arguments manually in lower versions, refer to https://docs.microsoft.com/en-us/previous-versions/17w5ykft(v=vs.85)
+            $escapedArgs = $Arguments | ForEach-Object {
+                # escape N consecutive backslash(es), which are followed by a double quote, to 2N consecutive ones
+                $s = $_ -replace '(\\+)"', '$1$1"'
+                # escape N consecutive backslash(es), which are at the end of the string, to 2N consecutive ones
+                $s = $s -replace '(\\+)$', '$1$1'
+                # escape double quotes
+                $s = $s -replace '"', '\"'
+                # quote the argument
+                "`"$s`""
+            }
+            $StartInfo.Arguments = $escapedArgs -join ' '
+        }
         $StartInfo.StandardErrorEncoding = $StartInfo.StandardOutputEncoding = [System.Text.Encoding]::UTF8
         $StartInfo.RedirectStandardError = $StartInfo.RedirectStandardInput = $StartInfo.RedirectStandardOutput = $true
-        $StartInfo.FileName = $Filename
-        $StartInfo.Arguments = $Arguments
         $StartInfo.UseShellExecute = $false
         if ($PWD.Provider.Name -eq 'FileSystem') {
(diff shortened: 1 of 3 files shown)


JanDeDobbeleer:
@lewis-yeung sounds like the right change to me. I can check for CMD if you want.

src/shell/scripts/omp.ps1
```diff
@@ -26,7 +26,7 @@ function global:Start-Utf8Process {
 
 $env:POWERLINE_COMMAND = "oh-my-posh"
 $env:CONDA_PROMPT_MODIFIER = $false
-$env:SHELL_VERSION = $PSVersionTable.PSVersion.ToString()
+$global:OMP_PSVERSION = $PSVersionTable.PSVersion.ToString()
```
JanDeDobbeleer: OK, fair point. I'm the env var person 😆

lewis-yeung:
@JanDeDobbeleer Much appreciated!

src/shell/scripts/omp.ps1
lewis-yeung: Aha, this is just in case the value of env `SHELL_VERSION` is inherited by a child process that coincidentally depends on it. So leaving it in PowerShell can be a better choice. 😆

lewis-yeung:
@JanDeDobbeleer I've put up another PR (#2152) and if you accept it, I will rebase this commit on it. If so, could you please do the rest work of `omp.lua` for CMD?

JanDeDobbeleer:
@lewis-yeung I would mark this is complete. I'll do the additional work later as a follow-up.

src/shell/scripts/omp.ps1
```diff
@@ -69,7 +69,6 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
             $command = $null
             $cursor = $null
             [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$command, [ref]$cursor)
-            $command = ($command -split " ")[0]
```
JanDeDobbeleer, line 72: this was here because using tooltips in a continuation prompt failed. Can you validate that use-case?

lewis-yeung: Yes, it's broken. Even without a continuation prompt context, it will fail when I have a double quote typed and then a space. 😲 But a single quote doesn't lead to an error. This is due to:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/07e0d0e0c60c528e034de815e90222dfff32544f/src/shell/scripts/omp.ps1#L73

where it messes up the double quote nesting for arguments. We can fix this, just let me try. 😆

lewis-yeung: We may need to refactor the `Start-Utf8Process` function. Yesterday I have seen the [implementation](https://github.com/starship/starship/blob/1736fe3f871e7a97e97a3ad58da10b7c8c065d6d/src/init/starship.ps1#L26-L65) in the init script for PowerShell of Starship, which uses an argument list. I'm not sure whether this helps for our case but it's worth trying.

lewis-yeung: @JanDeDobbeleer I have fixed this and re-validated it.

PR #2131, opened 2022-04-23
docs: add time zone example using Sprig

theperiscope:
### Prerequisites

- [x ] I have read and understood the `CONTRIBUTING` guide
- [x ] The commit message follows the [conventional commits][cc] guidelines
- [n/a ] Tests for the changes have been added (for bug fixes/features)
- [ x] Docs have been added / updated (for bug fixes/features)

### Description

Coming from pilot, sysadmin, and distributed team experiences I thought it would be useful to include an example in the documentation to display time in more than one time zone (esp. UTC since sometimes servers are configured in different time zones than dev machines).

Generally, Sprig provides excellent additions to templates for oh-my-posh and other places may benefits from similar function cross-references as well.

### DIFF
--- a/website/docs/segments/time.md
+++ b/website/docs/segments/time.md
@@ -76,5 +76,14 @@ Show the current timestamp.
 - StampMicro = "Jan _2 15:04:05.000000"
 - StampNano  = "Jan _2 15:04:05.000000000"
 
+## Examples
+
+To display the time in multiple time zones, using [Sprig's Date Functions][sprig-date]:
+
+```text
+{{ .CurrentDate | date .Format }} {{ dateInZone "15:04Z" .CurrentDate "UTC" }}
+```
+
 [templates]: /docs/configuration/templates
 [format]: https://yourbasic.org/golang/format-parse-string-time-date-example/
+[sprig-date]: https://masterminds.github.io/sprig/date.html


docs/docs/segments/time.md
```diff
@@ -21,6 +21,19 @@ Show the current timestamp.
 }
 ```
 
+### Display time in multiple time zones with [Sprig](https://masterminds.github.io/sprig/date.html)
```
JanDeDobbeleer: - Would it be an idea to add an `## Examples` section below Properties, where this is one of the examples? I feel like we want to have people read through to the bottom and then see how to use certain functionality.
- Links always need to be at the bottom, see other link examples in this file.

docs/docs/segments/time.md
```diff
@@ -21,6 +21,19 @@ Show the current timestamp.
...
+
+```json
+{
+  "type": "time",
+  "properties": {
+    "time_format": "1/2 15:04"
+  },
+  "style": "plain",
+  "template": "\ue389 {{ .CurrentDate | date .Format }} {{ dateInZone \"15:04Z\" .CurrentDate \"UTC\" }}"
```
JanDeDobbeleer: I would only show the `template`, not the entire block.

JanDeDobbeleer (changes requested):
Just a few remarks, great addition!

docs/docs/segments/time.md
theperiscope: The PR definitely was meant to spark a discussion, I appreciate the feedback. `Examples` is a great section to add across the documentation- Git's documentation [has it](https://git-scm.com/docs/git-restore#_examples) and a good sense of familiarity always enhances usability. Note, the headings there are also linked via named anchors -- makes sections easier to crosss-reference and share - I like that feature as well. What do you think?

docs/docs/segments/time.md
theperiscope: Would you show

```
"template": "\ue389 {{ .CurrentDate | date .Format }} {{ dateInZone \"15:04Z\" .CurrentDate \"UTC\" }}"
```
or skip icon and focus on template value (leave to user to Json-encode it correctly)? I think I prefer the latter slightly better, seems cleaner.
```
{{ .CurrentDate | date .Format }} {{ dateInZone "15:04Z" .CurrentDate "UTC" }}
```

theperiscope:
@JanDeDobbeleer I see that Markdownlint is failing due to the named anchor in the Examples header - shall I remove it? I do think it helps with public links it can be removed it if it is a requirement for everything to pass and consistency.

JanDeDobbeleer:
@theperiscope I'm going to have a look at this during the week. Been rather hectic lately.

Issue #2151, opened 2022-04-25
Deadlock in PowerShell function `Start-Utf8Process`


lewis-yeung:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

When I run the following command:

```powershell
Export-PoshTheme -Format json
```

I didn't get the expected output:

```
Theme copied to clipboard
```

The OMP child process is blocked indefinitely and a <kbd>Ctrl</kbd>+<kbd>C</kbd> input cannot kill it. While a direct call of `oh-my-posh config export --format=json` works well. Besides, this doesn't seem to happen on Linux. (I have tested WSL: Ubuntu-20.04 on my end, it works.)

The problem is located in `omp.ps1`:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/623fabcef31853255fa39afcf43cf7138d69079a/src/shell/scripts/omp.ps1#L152

https://github.com/JanDeDobbeleer/oh-my-posh/blob/623fabcef31853255fa39afcf43cf7138d69079a/src/shell/scripts/omp.ps1#L19-L24

It seems there is a deadlock condition. See [MS docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=net-6.0#remarks):

> The parent process would wait indefinitely for the child process to close its **StandardOutput** stream. The child process would wait indefinitely for the parent to read from the full **StandardError** stream.

> You can use asynchronous read operations to avoid these dependencies and their deadlock potential.

I got a solution from [Stack Overflow](https://stackoverflow.com/a/36539226) and made a fix for this within #2152.

### Theme

(This issue has nothing to do with a specific theme config.)

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

(N/A)

PR #2119, opened 2022-04-21
refactor(pwsh): purify global scope and environment

lewis-yeung:
### Prerequisites

- [x] I have read and understood the `CONTRIBUTING` guide
- [x] The commit message follows the [conventional commits][cc] guidelines

### Description

**Changes in init script `omp.ps1` for PowerShell:**

1. Do not leave temporary variables (`$Prompt` and `$secondaryPrompt`) in the current scope.

2. Use `$global:POSH_TRANSIENT` instead of `$env:POSH_TRANSIENT`. Its value is always automatically changed and there is no need to make it an environment variable.

3. Align all `OMP` prefixed global variable names using uppercase.

---

**Changes in `.vscode/launch.json`:**

Remove a redundant equal sign in `"--config==${workspaceRoot}/themes/cinnamon.omp.json"`.

---

BTW, I have another idea of moving the init stuff into a script block and making it a [**dynamic module**](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/new-module?view=powershell-7.2) imported in `omp.ps1`. This helps to utilize the encapsulation feature of the module, reduce unnecessary exports of global functions and global/environment variables to provide a tidy scope and make the script easily-maintained. I'm interested in completing further work and making it practical in the future.

[cc]: https://www.conventionalcommits.org/en/v1.0.0/#summary

PR #2152, opened 2022-04-25
Use dynamic module for init in PowerShell

lewis-yeung:
### Prerequisites

### Description

This has been mentioned for the first time in #2119. I haven't made that many changes in the body of the original init script `omp.ps1` (though it shows a large amount of differences, and this should go after any other work on `omp.ps1` just in case we should resolve a bunch of merge conflicts 😆). Changes as follows:

1. Reduce unnecessary exports of global functions (e.g., `Start-Utf8Process`, this function has a deadlock potential, see #2151) and global/environment variables to provide a tidy scope.

	<details><summary>List of necessary exported functions:</summary>

	- `Set-PoshContext`
	- `Enable-PoshTooltips`
	- `Enable-PoshTransientPrompt`
	- `Enable-PoshLineError`
	- `Export-PoshTheme`
	- `Get-PoshThemes`

	</details>

2. The error code consistency is now guaranteed.

	<details><summary>About this:</summary>

	I used to be confused about a weird behavior of the exit code in PowerShell. If I make a clear-screen by <kbd>Ctrl</kbd>+<kbd>L</kbd> (default key combination for this), the error code (`.Code`) in a newly spawned OMP prompt will be always 0 whatever the real last return status is! Actually, it happens even without the use of OMP.

	Is it a bug or by design? And if it is by design, how to get the real last return status? I searched for a solution everywhere but got nothing good. After I found this is done by a key handler which can be bound by `Set-PSReadLineKeyHandler`, I got the keypoint. The `$?` is (actually, have to be) always `$true` in any key handler scope, and if I make a command execution failure in a custom key handler, it will show an error like this:

	```
	An exception occurred in custom key handler, see $error for more information: ...
	```

	As the default [`ClearScreen`](https://github.com/PowerShell/PSReadLine/blob/77e256abcbd4f6e2ce47cacb691d730967969e0b/PSReadLine/Movement.cs#L506-L511) handler function calls `InvokePrompt` (corresponding to `[Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt` in PowerShell):

	```cs
	public static void ClearScreen(ConsoleKeyInfo? key = null, object arg = null)
	{
	    var console = _singleton._console;
	    console.Write("\x1b[2J");
	    InvokePrompt(null, console.WindowTop);
	}
	```

	the `$?` will be always `$true` in `prompt` function which is invoked internally.

	Since a key handler doesn't change the "history id", we make a simple workaround: store the current error code and next time try to change it only when a non-empty command is called (the "history id" doesn't change for an empty hit).

	On the case of a check for the recent error, I followed the solution in [starship/src/init/starship.ps1](https://github.com/starship/starship/blob/d4d84a592c2be43a9c8fa398884066ecc98a9f80/src/init/starship.ps1#L94-L100). **Special thanks to contributors of [Starship](https://github.com/starship/starship).** ❤️

	</details>

3. Rename some variables in the `script:` scope.

**🆕 Update: The fix for #2151 has been added.**


src/shell/scripts/omp.ps1
```diff
@@ -1,249 +1,183 @@
...
+    }
 
-function global:Get-PoshContext {
-    $cleanPWD = $PWD.ProviderPath
-    $cleanPSWD = $PWD.ToString()
-    $cleanPWD = $cleanPWD.TrimEnd('\')
-    $cleanPSWD = $cleanPSWD.TrimEnd('\')
-    return $cleanPWD, $cleanPSWD
-}
+    function Set-PoshContext {}
```
JanDeDobbeleer, line 48: can we still override this one in this case?

JanDeDobbeleer: Never mind, I missed the export at the bottom

src/shell/scripts/omp.ps1
```diff
@@ -1,249 +1,183 @@
-function global:Start-Utf8Process {
-    param(
-        [string] $FileName,
-        [string] $Arguments
-    )
+New-Module -Name "OhMyPosh" -ScriptBlock {
```
JanDeDobbeleer: Can we rename this to `oh-my-posh-init`? The standard module is called `oh-my-posh` so this aligns the names a bit. Other than that, I'm impressed. Amazing work (also validated it myself).

JanDeDobbeleer (changes requested):
1 minor remark

src/shell/scripts/omp.ps1
lewis-yeung: 💡 Just a suggestion: how about name it "oh-my-posh-core"? The dynamic module not only does the init work, but also persists in the PowerShell session and plays a "core" role in wrapping OMP calls.

JanDeDobbeleer:
@lewis-yeung something I didn't validate is check if this also works from the module. Can we import a module from within a module?

src/shell/scripts/omp.ps1
JanDeDobbeleer: Perfect!

JanDeDobbeleer: Perfect!

lewis-yeung:
@JanDeDobbeleer Tested just now, and... Oops.

```
PS> Enable-PoshTooltips
Enable-PoshTooltips: The term 'Enable-PoshTooltips' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

The functions seem to be unexported because of a nested import. But don't worry, we can find a way out. 😆

JanDeDobbeleer:
@lewis-yeung I was planning on removing the module anyways. Meaning I'd need to prepare that first, then merge this one. I'd say the time is now.

lewis-yeung:
@JanDeDobbeleer The simplest workaround I can find is to use a `global:` prefix for those function definitions instead of making them exported. This can be needed if the PowerShell module is not totally deprecated in coming releases.

JanDeDobbeleer:
@lewis-yeung OK, but that goed against the purpose we're trying to achieve here. I'll deprecate the module this week.

lewis-yeung:
> The functions seem to be unexported because of a nested import.

I got an easier fix! Just add `-Global` to `Import-Module`. Ref: [Powershell loading modules inside of a module [Scope]](https://stackoverflow.com/a/50876222).

I'll make this draft ready once I finish some other checks and fixes.

JanDeDobbeleer:
@lewis-yeung this can go live as this version will never be downloaded in the module (removed).

lewis-yeung:
@JanDeDobbeleer I'm on it. Soon to be live. 👌

lewis-yeung:
@JanDeDobbeleer This is now ready.

JanDeDobbeleer:
@lewis-yeung did you re-validate?

lewis-yeung:
@JanDeDobbeleer Yes, I did.

PR #2195, opened 2022-04-29
feat: iTerm2 integration for MacOS & Theme

ehawman-rosenberg:
### Prerequisites

- [x] [ ! ] Tests for the changes have been added (for bug fixes/features)

### Description

iTerm2 offers a shell integration which puts a mark in the user's prompt. You may remember this from JanDeDobbeleer/oh-my-posh2#261.

I've tried to create a segment that will inject this mark only if the user 
A) is using iTerm
B) has properly set env ITERM2_SQUELCH_MARK=1
C) is using zsh, bash, or fish

My issue is that I don't know Go and I need just a little leg-up on completing my tests. It fails properly, but returns empty strings when I'm expecting responses. Can somebody take a really quick look at my segment/test and tell me which is to blame? I suspect that my test is malformed...

### DIFF
--- a/docs/docs/contributing/segment.md
+++ b/docs/docs/contributing/segment.md
@@ -69,7 +69,7 @@ NEW SegmentType = "new"
 
 ## Add the SegmentType mapping
 
-Map your `SegmentType` to your Segment in the `mapSegmentWithWriter` function.
+Map your `SegmentType` to your Segment in the `mapSegmentWithWriter` function (respect the alphabetical order).
 
 ```go
 NEW: &New{},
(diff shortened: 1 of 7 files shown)


ehawman-rosenberg:
I have tested this in a shell on Windows and MacOS and it works as expected. I'm further convinced that there's a problem with my test code.

JanDeDobbeleer:
@ehawman-rosenberg I'll have a look at it and will fix. Hopefully this evening!

JanDeDobbeleer:
@All-contributors please add @ehawman-rosenberg for code

PR #2259, opened 2022-05-12
theme: add kali-wsl

playday3008:
### Prerequisites

### Description

Same as `kali.omp.json`, but colours identical to original Kali bash terminal in WSL, also changed `@` to `㉿` in non-root user
Also, (I don't know why, but) I change hard-coded `root` username to dynamic

### DIFF
--- a/themes/kali-wsl.omp.json
+++ b/themes/kali-wsl.omp.json
@@ -1,0 +1,91 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "blocks": [
+    {
+      "alignment": "left",
+      "segments": [
+        {
+          "foreground": "lightBlue",
+          "foreground_templates": [
+            "{{ if .Root }}lightRed{{ end }}"
+          ],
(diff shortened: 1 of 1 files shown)


playday3008:
Root:
image
Non-root:
image
(IDK why symbols is not displaying, so):
image

lewis-yeung:
> IDK why symbols is not displaying

@playday3008 From [docs](https://ohmyposh.dev/docs/share):

> Some glyphs aren't rendered correctly, that's not you but the limitations of the renderer.

BTW, I like the cute skull 💀. 😆

JanDeDobbeleer:
@playday3008 any reason why we're not adjusting the kali theme itself?

playday3008:
> @playday3008 any reason why we're not adjusting the kali theme itself?

There is no important reason

JanDeDobbeleer:
@playday3008 in that case I would adjust the current theme to reflect the real prompt. Makes it easier for people.

playday3008:
> @playday3008 in that case I would adjust the current theme to reflect the real prompt. Makes it easier for people.

Ok, close PR after that)

Issue #2289, opened 2022-05-19
Conflict with zsh-autosuggestions and history-incremental-search-backward


AlexJF:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

https://github.com/JanDeDobbeleer/oh-my-posh/blame/main/src/shell/scripts/omp.zsh#L62 introduced by https://github.com/JanDeDobbeleer/oh-my-posh/pull/2129 causes (at least) the following symptoms in ZSH if `enable_poshtooltips` is called:

* With zsh-autosuggestions, suggested autocompletion will be accepted but with a space inserted in the middle:
   
    <img width="175" alt="image" src="https://user-images.githubusercontent.com/373323/169356531-c74d1105-6c32-4f3f-8669-b2f20d4745c9.png">
    <img width="180" alt="image" src="https://user-images.githubusercontent.com/373323/169356692-979ad913-9def-4ecc-94de-b0e7a6deb71f.png">

* With the history-incremental-search-backward function (`bindkey '^R' history-incremental-search-backward`), search will be exited as soon as space is pressed

    <img width="181" alt="image" src="https://user-images.githubusercontent.com/373323/169356855-8c655402-b9fe-4c26-87b0-5fbf06c8dd57.png">
    <img width="184" alt="image" src="https://user-images.githubusercontent.com/373323/169356899-7f43797c-1179-4064-aab9-71cb284272af.png">

As soon as I overwrite the keybinding on " " to one of the traditional behaviours (e.g. `bindkey " " magic-space` or `bindkey " " self-insert`), things work again but tooltips seem to stop working:

* zsh-autosuggestions (not entirely clear but the suggestion is not accepted here, it's still just a suggestion for the command that I ended up executing before due to this bug)

    <img width="175" alt="image" src="https://user-images.githubusercontent.com/373323/169356531-c74d1105-6c32-4f3f-8669-b2f20d4745c9.png">
    <img width="174" alt="image" src="https://user-images.githubusercontent.com/373323/169357168-18def0a4-4e85-48b6-bc09-d4fb85d92b6b.png">

* history-incremental-search-backward function

    <img width="181" alt="image" src="https://user-images.githubusercontent.com/373323/169356855-8c655402-b9fe-4c26-87b0-5fbf06c8dd57.png">
[...]

PR #2296, opened 2022-05-20
feat(zsh): trigger a tip check only on a space key input

lewis-yeung:
### Prerequisites

### Description

Related: #2129, #2289. A reimplementation for aligning tooltip behaviors in zsh. This doesn't overwrite the keybinding on the space key, which won't affect zsh plugins that work before.

Ref: [Zle Widgets](https://zsh.sourceforge.io/Doc/Release/Zsh-Line-Editor.html#Zle-Widgets).

### DIFF
--- a/src/shell/scripts/omp.zsh
+++ b/src/shell/scripts/omp.zsh
@@ -48,7 +48,10 @@ function self-insert() {
     zle .self-insert
     return
   fi
-  tooltip=$(::OMP:: print tooltip --config="$POSH_THEME" --shell=zsh --command="$BUFFER" --shell-version="$ZSH_VERSION")
+  # trigger a tip check only if the input is a space character
+  if [[ "$KEYS" = " " ]]; then
+    tooltip=$(::OMP:: print tooltip --config="$POSH_THEME" --shell=zsh --command="$BUFFER" --shell-version="$ZSH_VERSION")
+  fi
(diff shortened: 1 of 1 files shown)


src/shell/scripts/omp.zsh
```diff
@@ -48,7 +48,10 @@ function self-insert() {
     zle .self-insert
     return
   fi
-  tooltip=$(::OMP:: print tooltip --config="$POSH_THEME" --shell=zsh --command="$BUFFER" --shell-version="$ZSH_VERSION")
+  # trigger a tip check only if the input is a space character
+  if [[ "$KEYS" = " " ]]; then
```
JanDeDobbeleer, line 52: Can't we do this check in `self_insert`? That way we don't reload unnecessarily.

lewis-yeung: Sorry, I don't quite understand but where is `self_insert`? There is only `self-insert` and this check is done in it.

JanDeDobbeleer: I'm sorry, should've looked at this on a bigger display 🤦🏻‍♂️

lewis-yeung: Fine. So is there anything I should change? 😄

JanDeDobbeleer (approved):

PR #2333, opened 2022-05-26
feat: Nx segment

evanfuture:
### Prerequisites

### Description

Add a new segment to show the Nx (monorepo tooling) version.


src/segments/nx.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (a *Nx) Init(props properties.Properties, env environment.Environment) {
+	a.language = language{
+		env:        env,
+		props:      props,
+		extensions: []string{"workspace.json"},
+		commands: []*cmd{
+			{
+				regex: `(?:(?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+))))`,
+				getVersion: func() (string, error) {
```
JanDeDobbeleer: Can you define this function below and reference it here?

JanDeDobbeleer (changes requested):

src/segments/nx.go
evanfuture: Done.  I'm not really proficient with Go, but it seems to work.

I basically just copy/pasted from the Angular segment. 😅

JanDeDobbeleer:
@evanfuture there's some other changes needed but I might just do them as the logic isn't flawed, it just needs to be a little more cross platform proof (and I'm not at my laptop right now).

src/segments/nx.go
JanDeDobbeleer: I noticed that only after the facts. And truth be told, that one was actually not OK. I must've been a bit under the weather when I merged that one in. But, fixed it.

PR #2345, opened 2022-05-27
Remove old fix, unhide last character from the line

nunu:
### Prerequisites

### Description

At the right segment you should add "horizontal_offset": 1,  

### DIFF
--- a/src/engine/engine.go
+++ b/src/engine/engine.go
@@ -150,13 +150,6 @@ func (e *Engine) renderBlock(block *Block) {
 		}
 		e.rprompt = text
 	}
-	// Due to a bug in Powershell, the end of the line needs to be cleared.
-	// If this doesn't happen, the portion after the prompt gets colored in the background
-	// color of the line above the new input line. Clearing the line fixes this,
-	// but can hopefully one day be removed when this is resolved natively.
-	if e.Env.Shell() == shell.PWSH || e.Env.Shell() == shell.PWSH5 {
-		e.writeANSI(e.Ansi.ClearAfter())
-	}
 }
 
 // debug will loop through your config file and output the timings for each segments


nunu:
image

JanDeDobbeleer:
@nunu what's your motivation for this change? As changing the line start isn't the same as clearing after the cursor. And additionally it puts an action in the hands of the end user.

nunu:
> @nunu what's your motivation for this change? As changing the line start isn't the same as clearing after the cursor. And additionally it puts an action in the hands of the end user.

for me it looks better it works just as well. 
Maybe I don't understand what it affects, do you have a link or some more information about?

JanDeDobbeleer:
@nunu what do you mean with `it looks better`? As this has no influence on the way it looks but rather prevents it from looking bad.

nunu:
By default one character is added to the end of line

image

I try with "horizontal_offset": 1, at right segment

And it hide last character and add empty character

image
image

If i remove those line and add "horizontal_offset": 1, it fix this

image


image

JanDeDobbeleer:
@nunu that's weird as I don't have that? can I get your config?

nunu:
> @nunu that's weird as I don't have that? can I get your config?

https://gist.github.com/nunu/7bac3696ae3e4b26c68f5d02aa2224c6

maybe is my wrong theme

image

JanDeDobbeleer:
@nunu I'll check!

JanDeDobbeleer:
@nunu I can't reproduce that block with your config. It's not there. What are your Windows Terminal settings?

nunu:
> @nunu I can't reproduce that block with your config. It's not there. What are your Windows Terminal settings?

https://gist.github.com/nunu/3149dc2d1774baa341fb91db7f2f47f3


"initialCols": 180,
"initialRows": 36,

georgetzlp:
This also happens to me, even happens in normal Command Prompt.

JanDeDobbeleer:
@wathhr do you have a screenshot for "normal command prompt"? Somehow this feels like a Windows Terminal bug, not an oh-my-posh one. Just like this fix is also there to solve a PowerShell bug with colorization when at the end of the terminal screen. So removing this isn't an option.

georgetzlp:
image
with
```json
    {
      "alignment": "right",
      "type": "prompt",
      "segments": [
        {
          "foreground": "darkGray",
          "style": "plain",
          "type": "text",
          "template": "test",
          "interactive": true
        }
      ]
    }
```

JanDeDobbeleer:
@wathhr @nunu if this also happens on CMD, the issue is different and the proposed fix isn't the solution as that only addresses PWSH.

PR #2398, opened 2022-06-09
fix(bash): contract consecutive quoted ANSI sequences \[...\]

akinomyoga:
## Prerequisites

> - Docs have been added/updated (for bug fixes/features)

I have a question: Where to put the docs for the change of the pull request? I found some "Fixes" sections e.g. in [website/blog/2022-05-19-whats-new-3.mdx](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/website/blog/2022-05-19-whats-new-3.mdx). but the recent changes mentioned in [the release page of v8.3.2](https://github.com/JanDeDobbeleer/oh-my-posh/releases/tag/v8.3.2) does not seem to be included in the docs in the repository. [The corresponding commit for the change in v8.3.2](https://github.com/JanDeDobbeleer/oh-my-posh/commit/265790659c5bdfd8ef28be3bcb6b57e68c35b522) does not either seem to include the doc changes. I would appreciate it if you could point me where to put the docs on the change. I also think it would be nice to put the same information in the pull-request template.

### Summary

There has been a report about the interference between oh-my-posh settings and my Bash configuration [`akinomyoga/ble.sh`](https://github.com/akinomyoga/ble.sh). The problem is that the right prompt by Oh My Posh (i.e., I mean `{ "alignment": "right", "type": "prompt" }` by the "right prompt") is not correctly shown with ble.sh loaded. The reason for the problem is that Oh My Posh is breaking the assumption that the ANSI sequences quoted by `\[...\]` do not affect the cursor position.

### Repeat-By

For example, the problem arises with the following simple configuration:

```js
{
  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
  "blocks": [
    {
      "alignment": "right",
      "segments": [
        {
          "template": " {{ .CurrentDate | date .Format }} ",
          "properties": { "time_format": "15:04:05" },
          "style": "plain",
          "type": "time"
        }
      ],
      "type": "prompt",
    },
    {
      "alignment": "left",
      "newline": true,
      "segments": [
        {
          "template": " {{ .Path }} ",
          "properties": { "style": "folder" },
          "style": "plain",
          "type": "path"
        }
      ],
      "type": "prompt"
    }
  ],
  "final_space": true,
  "version": 2
}
```

with bashrc being

```bash
eval "$("/path/to/oh-my-posh" init bash --config "/path/to/config.json")"
source /path/to/ble.sh --norc
```

An expected result is something like this:

image

However, the actual result becomes

image

### What is happening

Oh My Posh of the present `main` branch produces the `PS1` content equivalent to

```bash
PS1='\[\e[1000C\]\[\e[10D\]\[\e[37m\] 11:28:32 \[\e[0m\]\n\[\e[37m\] oh-my-posh \[\e[0m\]\[\e[0m\] '
```

Then, `ble.sh` parses the content of `PS1` and tries to reconstruct the rendered result in its *internal virtual terminal*. Since `\[\e[1000C\]` (*go right by 1000 columns in the terminal*) is enclosed within `\[\]`, ble.sh assumes that the cursor position is still at the original position (i.e., the beginning of the line) when it parses the next sequence `\[\e[10D\]` (*go left by 10 columns in the terminal*). However, the cursor cannot go left at the beginning of the line, so `\e[10D` does not actually have any effect in the virtual terminal and is ignored. Finally, the time <code>&nbsp;11:28:32 </code> is just written from the last column of the real terminal.

The reason why Bash requests us to enclose the escape sequences with `\[...\]` is that they otherwise disturb the cursor-position calculation of the line editor. `\[...\]` is used to mark that the enclosed string does not affect the cursor position. For this reason, the escape sequences enclosed within `\[...\]` should not alter the cursor position or come back to the original position. At least, the column position of the cursor should be preserved, or otherwise, Bash (or more specifically Readline library) fails to correctly render its contents (especially for a long command that does not fit in a single terminal line). However, the current implementation of OMP breaks this assumption by putting each cursor movement sequence in a single `\[\]` separately.

Actually, the present implementation does not cause the problem in Bash because Bash does not use the cursor position of the intermediate point of `PS1`, so even if the cursor position is moved within some `\[\]` ranges, no problem arises as far as the column position finally matches with Bash's prediction. On the other hand, ble.sh assumes that each `\[...\]` is correctly used to satisfy the original intent of Bash to calculate the cursor position of an arbitrary point in `PS1`.

### Ideal solution

Ideally, if we respect the original idea of `\[\]`, we should produce the following `PS1`:

```bash
PS1='\[\e[1000C\e[10D\e[37m 11:28:32 \e[0m\n\]\[\e[37m\] oh-my-posh \[\e[0m\]\[\e[0m\] '
```
where the entire *excursion* of the cursor for the right prompt (`\e[1000C\e[10D\e[37m 11:28:32 \e[0m\n` which finally recovers the original column position) is enclosed within a single `\[...\]` range.

However, by looking at the implementation, I would think we need to rewrite the entire codebase to properly handle this context-dependent quoting by `\[...\]`. The most non-trivial point here is that the right prompt in Oh My Posh is currently not designed to recover the original cursor position. This also involves the existing configuration user interface, which I think should not be easily changed.

### Alternative solution (Workaround, this PR)

Here, I would suggest a workaround: We combine the consecutive `\[\]`s to a single one, such as `\[xxx\]\[yyy\]\[zzz\]` to `\[xxxyyyzzz\]` at the final stage. This does not perfectly solve the present conceptual problem but works well with the current version of `ble.sh` (although it might still cause a problem in the future).

[ *Even if we forget about the present problem, the proposed change alone might also be good for reducing the length and improving the legibility of the raw `PS1`, though the difference might be negligible... /2c* ]

### DIFF
--- a/src/engine/engine.go
+++ b/src/engine/engine.go
@@ -5,6 +5,7 @@ import (
 	"oh-my-posh/color"
 	"oh-my-posh/console"
 	"oh-my-posh/environment"
+	"oh-my-posh/regex"
 	"oh-my-posh/shell"
 	"oh-my-posh/template"
 	"strings"
@@ -37,7 +38,16 @@ func (e *Engine) writeANSI(text string) {
 }
 
 func (e *Engine) string() string {
-	return e.console.String()
+	ansiSequence := e.console.String()
+	// Contract consecutive quoted ANSI sequences as
+	// %{...%}%{...%} -> %{......%} (zsh) or \[...\]\[...\] -> \[......\] (bash)
+	switch e.Env.Shell() {
+	case shell.ZSH:
+		ansiSequence = regex.ReplaceAllString(`((?:^|[^%])(?:%%)*)%\}%\{`, ansiSequence, "$1")
+	case shell.BASH:
+		ansiSequence = regex.ReplaceAllString(`((?:^|[^\\])(?:\\\\)*)\\\]\\\[`, ansiSequence, "$1")
+	}
+	return ansiSequence
 }
 
 func (e *Engine) canWriteRPrompt() bool {
--- a/src/engine/engine_test.go
+++ b/src/engine/engine_test.go
@@ -43,6 +43,57 @@ func TestCanWriteRPrompt(t *testing.T) {
 	}
 }
 
+func createEngineForShell(shellName string) *Engine {
+	env := &environment.ShellEnvironment{
+		CmdFlags: &environment.Flags{
+			Shell:  shellName,
+		},
+	}
+	env.Init()
+	defer env.Close()
+
+	cfg := LoadConfig(env)
+	defer testClearDefaultConfig()
+
+	ansi := &color.Ansi{}
+	ansi.Init(env.Shell())
+	writerColors := cfg.MakeColors(env)
+	writer := &color.AnsiWriter{
+		Ansi:               ansi,
+		TerminalBackground: shell.ConsoleBackgroundColor(env, cfg.TerminalBackground),
+		AnsiColors:         writerColors,
+	}
+	consoleTitle := &console.Title{
+		Env:      env,
+		Ansi:     ansi,
+		Template: cfg.ConsoleTitleTemplate,
+	}
+	engine := &Engine{
+		Config:       cfg,
+		Env:          env,
+		Writer:       writer,
+		ConsoleTitle: consoleTitle,
+		Ansi:         ansi,
+	}
+	return engine
+}
+
+func TestContractionOfQuotedAnsiSequences(t *testing.T) {
+	e := createEngineForShell(shell.BASH)
+	e.write(e.Ansi.CarriageForward())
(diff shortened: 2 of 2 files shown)


JanDeDobbeleer:
@akinomyoga I was looking at this, if I get it correctly only changing the way we print the rprompt should be fine too, yes? I'll give this a go too later this week as I wouldn't want magic regexes ideally (there's already a few 😅).

akinomyoga:
Thank you for your response! Great! If it will be fixed at the stage of constructing the sequences, that should be better than this workaround!

JanDeDobbeleer:
@akinomyoga can you have a look at #2407 and see if that also fixes the issue? This should print the entire rprompt as "plain" (not escaped) and enclose it in the escape characters.

akinomyoga:
Thank you! I have made a comment in #2407!

PR #2408, opened 2022-06-13
docs:Added source flag to winget installation

RFCreate:
### Prerequisites

### Description

I decided to add the source flag (-s winget) to the installation, because maybe you don't want to be prompted with a sort of error message that asks you to sign an agreement with Microsoft to let them know your device location. I propose to remove this annoying prompt by adding the source flag to winget.

### DIFF
--- a/website/docs/installation/windows.mdx
+++ b/website/docs/installation/windows.mdx
@@ -43,7 +43,7 @@ When using oh-my-posh inside the WSL, make sure to follow the [Linux][linux] ins
 Open a PowerShell prompt and run the following command:
 
 ```powershell
-winget install JanDeDobbeleer.OhMyPosh
+winget install JanDeDobbeleer.OhMyPosh -s winget
 ```
 
 </TabItem>
@@ -101,7 +101,7 @@ Exclusions should be added with the full path to the executable, you can get it 
 Open a PowerShell prompt and run the following command:
 
 ```powershell
-winget upgrade oh-my-posh
+winget upgrade JanDeDobbeleer.OhMyPosh -s winget
 ```
 
 </TabItem>


JanDeDobbeleer:
@shyguyCreate I never got this popup, BUT I wouldn't silently remove it for end users. I feel everyone should be in the know.

RFCreate:
@JanDeDobbeleer I agree with you, but shouldn't the command install the program just by running it, the prompt just breaks the flow making you read and sign things.

And most importantly, if you don't sign, the command will FAIL.

JanDeDobbeleer:
@shyguyCreate yes, but that's a winget thing. As such it has nothing to do with oh-my-posh.

RFCreate:
Internet photo of the agreement:
internet msstore agreement

@JanDeDobbeleer Of course it is a winget thing, but that behavior can be avoided. Personally, I always use the --source flag because I don't want to sign the agreement.

JanDeDobbeleer:
@shyguyCreate still, this is something for the winget docs, not something we should be working around or documenting imho.

RFCreate:
@JanDeDobbeleer this is the same thing you were referencing in this [discussion](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/2402).

You were referencing the winget source anyways in a discrete manner. What I say is that purposely point to the source that you want to install from, and based on your [answer](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/2402#discussioncomment-2924356) you want the source to be winget.

You could even revert back to the previous name, like this:

>winget install oh-my-posh -s winget

And it will work and no fail.  (personally I prefer Id names: `JanDeDobbeleer.OhMyPosh`)

JanDeDobbeleer:
@shyguyCreate you're right, but I needed to revert to JanDeDobbeleer.OhMyPosh which is better imho. So that would be perfect.

RFCreate:
> @shyguyCreate you're right, but I needed to revert to JanDeDobbeleer.OhMyPosh which is better imho. So that would be perfect.

@JanDeDobbeleer Yes, I said that I prefer Id names.

Maybe I did not understand, but what would be perfect?

JanDeDobbeleer:
@shyguyCreate change it to:

```
winget install JanDeDobbeleer.OhMyPosh
```

There's no overlap using that id.

RFCreate:
@JanDeDobbeleer the installation phase is already as you marked in the code.
>winget install JanDeDobbeleer.OhMyPosh

A change in the upgrade behavior will really suit for a commit.
```
winget upgrade JanDeDobbeleer.OhMyPosh
```
---
But I still recommend to use the source flag.
```
winget install JanDeDobbeleer.OhMyPosh -s winget
winget upgrade JanDeDobbeleer.OhMyPosh -s winget
```

JanDeDobbeleer (approved):

PR #2421, opened 2022-06-20
fix(pwsh): do not print transient prompt on a ReadLine parsing error

### LINKED ISSUE #2417: Transient prompt and secondary prompt (continuation prompt) cannot coexist in PowerShell
### Code of Conduct

### What happened?

This happens when the transient prompt is enabled. In a continuation prompt context, the cursor will not move to the next line when I press <kbd>Enter</kbd>, instead it sticks to the current line. Pressing <kbd>Enter</kbd> repeatedly, the transient prompt will move up and up until being out of the upper bound of the terminal screen buffer, which raises an exception: (the following is from `$error[0]`)

```
MethodInvocationException:
Line |
  85 |  …              [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
     |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "InvokePrompt" with "0" argument(s): "The value must be greater than or equal to zero and less than the console's buffer size in that dimension. (Parameter 'top')
Actual value was -1."
```

bug

This never happens in a PowerShell session without OMP. Having no solution for the time being. Or can we change the implementation of the transient prompt to fix this?

### Theme

(This issue has nothing to do with a specific theme config.)

(This issue has nothing to do with a specific OS.)

(This issue has nothing to do with a specific shell.)

### Log output

(N/A)

lewis-yeung:
### Prerequisites

### Description

Fix #2417.

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -97,13 +97,12 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
     }
 
     function Enable-PoshTooltips {
-        Set-PSReadlineKeyHandler -Key SpaceBar -ScriptBlock {
+        Set-PSReadLineKeyHandler -Key SpaceBar -ScriptBlock {
             [Microsoft.PowerShell.PSConsoleReadLine]::Insert(' ')
             $position = $host.UI.RawUI.CursorPosition
             $cleanPWD, $cleanPSWD = Get-PoshContext
             $command = $null
-            $cursor = $null
-            [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$command, [ref]$cursor)
+            [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$command, [ref]$null)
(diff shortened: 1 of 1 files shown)


src/shell/scripts/omp.ps1
```diff
@@ -111,12 +111,19 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
...
-        Set-PSReadlineKeyHandler -Key Enter -ScriptBlock {
-            $script:TransientPrompt = $true
+        Set-PSReadLineKeyHandler -Key Enter -ScriptBlock {
             $previousOutputEncoding = [Console]::OutputEncoding
-            [Console]::OutputEncoding = [Text.Encoding]::UTF8
             try {
-                [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
+                $ast = $null
+                $tokens = $null
+                $parseErrors = $null
```
JanDeDobbeleer, line 116: shouldn't we only need this variable and add the other `$null` inline?

JanDeDobbeleer (changes requested):

src/shell/scripts/omp.ps1
lewis-yeung: 🆗 Changed.

Issue #607, opened 2021-04-01
Conditional Newline


sdsalyer:
### Prerequisites

- [X] I have read and understand the `CONTRIBUTING` guide
- [X] I looked for duplicate issues before submitting this one

### Description

Is it possible to set a conditional newline? For example, if I'm in a directory under Git version control, display the git / poshgit info on the next line. Otherwise, don't display the extra (blank) line.

### Environment

- Oh my Posh version: 3.130.1
- Theme: (customized)
- Operating System: Windows
- Shell: PowerShell
- Terminal: Windows Terminal

### Steps to Reproduce
n/a

**Expected behavior:** [What you expected to happen]
```
 14:23:23 █C:    Bitbucket█
█develop ≢█
PS❯ cd ..
 14:23:30 █C:    PowerShell█
PS❯ pwd
```

**Actual behavior:** [What actually happened]
```
 14:23:23 █C:    Bitbucket█
█develop ≢█
PS❯ cd ..
 14:23:30 █C:    PowerShell█

PS❯ pwd
```

Theme file:
```json
{
    "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
    "blocks": [
        {
            "type": "prompt",
            "alignment": "left",
            "segments": [
                {
                    "type": "root",
                    "style": "plain",
                    "background": "#ac0101",
                    "foreground": "#fce93c",
                    "properties": {
                        "root_icon": "[ADMIN]"
                    }
                },
                {
                    "type": "time",
                    "style": "plain",
                    "properties": {
                        "time_format": "15:04:05"
                    }
                },
                {
                    "type": "path",
                    "style": "powerline",
                    "background": "#91ddff",
                    "foreground": "#100e23",
                    "powerline_symbol": "",
                    "properties": {
                        "folder_icon": "",
                        "folder_separator_icon": "  ",
                        "style": "agnoster_short"
                    }
                }
            ]
        },
        {
            "type": "newline"
        },
        {
            "type": "prompt",
            "alignment": "left",
            "segments": [
                {
                    "type": "git",
                    "style": "powerline",
                    "powerline_symbol": "\uE0B0",
                    "foreground": "#193549",
                    "background": "#95ffa4"
                }
            ]
        },
        {
            "type": "newline"
        },
        {
            "type": "prompt",
            "alignment": "left",
            "segments": [
                {
                    "foreground": "#007ACC",
                    "properties": {
[...]

PR #2435, opened 2022-06-22
docs: add `"type": "newline"` to block.mdx

Aierbote:
### Prerequisites

### Description

I'm pretty noob, so I hope the adjustment I made are fine

I was looking for a way to add a new line `\n`to get the cursor on a separate line and found no mention about this third kind of argument of `type` property.

Found the solution after a long while in a [comment of yours](https://github.com/JanDeDobbeleer/oh-my-posh/issues/607#issuecomment-812333646)

### DIFF
--- a/website/docs/configuration/block.mdx
+++ b/website/docs/configuration/block.mdx
@@ -80,10 +80,29 @@ Tells the engine what to do with the block. There are two options:
 - `prompt` renders one or more segments
 - `rprompt` renders one or more segments aligned to the right of the cursor. Only one `rprompt` block is permitted.
 Supported on zsh, bash, PowerShell, cmd and fish.
+- `newline` renders your cursor in a new line, useful if your custom prompt is full of segments and you aim at readabilty. 
+
+:::info
+Example given shows a segment with just this property as the last segment in the `blocks` list:
+
+```json
+        {
+            "type": "newline"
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@Aierbote this is actually deprecated. To start a block on a newline, you should use `"newline": true` in the block that should start on a new line. That's documented [here](https://ohmyposh.dev/docs/configuration/block).

Aierbote:
@JanDeDobbeleer Oh. That's odd, that was the simplest and only way I found to add a new line after the prompt...  🥲

Apologize for my noobness 😅

---

Since it's been deprecated, could it be useful to add a TIP/INFO in the block.mdx to do show an explicit alternative? Some kind of example along these lines:

"""
DEPRECATED: instead of  
```json
        {
            "type": "newline"
        },
```
use something like this to add a newline at the end
```json
        {
          "alignment": "left",
          "newline": true,
          "segments": [
            {
              "foreground": "#007ACC",
              "style": "plain",
              "template": "\ue602 ",
              "type": "text"
            }
          ],
          "type": "prompt"
        }
```
"""

JanDeDobbeleer:
@Aierbote that functionality isn't even documented anymore so I wonder what we can do to improve finding the actual way to do it.

PR #2453, opened 2022-06-28
docs:Added completion info

RFCreate:
### Prerequisites

### Description

I have not seen in the docs any part that talks about the 'completion' command. So I decided to add it myself.

I was just able to test for pwsh and bash, but there is this [discussion](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/1960) for fish, and the command worked for bash so I expect to work the same in zsh.

The problem comes with pwsh that always marks an error when running, so can you point me out where the completion scripts are in the repo because I was not able to find them.

### DIFF
--- a/website/docs/installation/prompt.mdx
+++ b/website/docs/installation/prompt.mdx
@@ -66,6 +66,14 @@ Once added, reload your profile for the changes to take effect.
 . $PROFILE
 ```
 
+:::tip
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@shyguyCreate this is something Cobra offers out of the box, and isn't a core feature worth documenting as you hardly ever need to use this.

PR #2477, opened 2022-07-02
docs(share theme): add info about offsets

### LINKED ISSUE #2476: Vertical-Offset not respected in export image
### Code of Conduct

### What happened?

The problem is that my prompt is render this way:
bash omp

So the image should look like this because I have `"vertical_offset": -1` in my config :
omp gist image

But instead it looks like this, btw I use ` --rprompt-offset 40`: (difference in color no matter now)
ohmyposhCustome

I understand that considering "vertical-offset" in the first prompt will be for nothing, but in the next blocks they change the look of the final prompt. I even tested using a positive number instead of a negative, (meaning `"vertical-offset": 1`), and not even then it adds a extra line in the image. 

### Theme

https://gist.githubusercontent.com/shyguyCreate/387ff25579b25bff63a6bc1a7635be27/raw/2df499d1f7441254ba5456396692a216855fa12c/ohmyposhCustome.omp.json

In summary, my config has this structure:

```
"blocks": [
    {
      "alignment": "right",
      "segments": [ ...
      ],
      "type": "prompt"
    },
    {
      "alignment": "left",
      "newline": true,
      "vertical_offset": -1,
      "segments": [ ...
      ],
      "type": "prompt"
    },
    {
      "alignment": "left",
      "newline": true,
      "segments": [ ...
      ],
      "type": "prompt"
    }
  ]
```

_No response_

_No response_

### Log output

```Shell
No debug flag for `oh-my-posh config export image`
```

RFCreate:
### Prerequisites

### Description

Closes #2476

### DIFF
--- a/website/docs/share-theme.md
+++ b/website/docs/share-theme.md
@@ -19,6 +19,9 @@ to an image file (.png).
 oh-my-posh config export image --cursor-padding 50
 ```
 
+Be aware that "vertical-offset" and "horizontal-offset" block properties in your config,
+will **not** be taken into consideration when exporting the image.
+
 There are a couple of additional flags you can use to tweak the image rendering:
 
 - `--cursor-padding`: spaces to add after the cursor indication (`_`)


JanDeDobbeleer:
@shyguyCreate what I tried to explain was that these properties will no longer be supported (except for backwards compatibility purposes). They're redundant and the cause of a lot of unwanted side-effects. The only reason I added this is because I had no proper solution for there issues I was facing at the time.

PR #2481, opened 2022-07-03
fix(text): improve effective text width measuring

lewis-yeung:
### Prerequisites

### Description

Related: #2083.

### DIFF
--- a/src/color/text.go
+++ b/src/color/text.go
@@ -3,14 +3,19 @@ package color
 import (
 	"oh-my-posh/regex"
 	"strings"
-	"unicode/utf8"
+
+	"github.com/mattn/go-runewidth"
 )
+
+func init() { // nolint:gochecknoinits
+	runewidth.DefaultCondition.EastAsianWidth = false
+}
 
 func (ansi *Ansi) MeasureText(text string) int {
 	// skip strings with ANSI
 	if !strings.Contains(text, "\x1b") {
 		text = ansi.TrimEscapeSequences(text)
-		length := utf8.RuneCountInString(text)
+		length := runewidth.StringWidth(text)
 		return length
 	}
 	if strings.Contains(text, "\x1b]8;;") {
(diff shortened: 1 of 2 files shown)


src/engine/image.go
```diff
@@ -274,7 +274,7 @@ func (ir *ImageRenderer) lenWithoutANSI(text string) int {
 		text = strings.ReplaceAll(text, match[str], "")
 	}
 	stripped := regex.ReplaceAllString(color.AnsiRegex, text, "")
-	length := utf8.RuneCountInString(stripped)
+	length := runewidth.StringWidth(stripped)
 	for _, rune := range stripped {
 		length += ir.runeAdditionalWidth(rune)
```
JanDeDobbeleer: Doesn't this become obsolete in that case?

lewis-yeung: Intuition tells me so, but...

If we keep that logic, there is no difference between the new output image and that of the previous version.

normal

However, if I remove that, it turns out to be something like the following:

abnormal

(Tested using the theme `atomic.omp.json` with `rprompt-offset` as `10`.)

No clue yet. 😅

JanDeDobbeleer: @lewis-yeung this part is magical at times. A real struggle. So we keep this as is and merge?

lewis-yeung: @JanDeDobbeleer Magical indeed. Also, I think I have understood a little bit.

https://github.com/JanDeDobbeleer/oh-my-posh/blob/9f9b91e56688ed10014666a4ca815a50e794bfb9/src/engine/image.go#L420-L425

Actually, most of Nerd Font icon characters should be measured as 2-width glyphs in this case, while on a string including these characters, `runewidth.StringWidth` may return a width just the same as `utf8.RuneCountInString` does. Thus we need to keep that logic.

I'll revert Line 277 and then it is probably ready to be merged.

PR #2535, opened 2022-07-18
Add themes sim web

yauyy:
### Prerequisites

### Description
sim-web


themes/sim-web.omp.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "blocks": [
+    {
+      "type": "prompt",
+      "alignment": "right",
+      "filler": " ",
```
JanDeDobbeleer: why is there an empty filler?

JanDeDobbeleer (changes requested):
One question, please have a look.

themes/sim-web.omp.json
yauyy: Sorry, did not run the correct command to generate the image, resulting in a style exception

themes/sim-web.omp.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
+  "blocks": [
+    {
+      "type": "prompt",
+      "alignment": "right",
+      "filler": "",
```
JanDeDobbeleer: this is the default and can be removed ;-)

themes/sim-web.omp.json
```diff
@@ -0,0 +1,88 @@
...
+      "newline": true,
```
JanDeDobbeleer: is there a specific reason to start with a newline?

JanDeDobbeleer (changes requested):
I would do the following, move the second block to the first position and make the first block a right aligned one so the prompt spans 2 lines and not three.

themes/sim-web.omp.json
yauyy: ok~

themes/sim-web.omp.json
yauyy: Wanted a little distance from the last command~

yauyy:
> 我将执行以下操作，将第二个块移动到第一个位置并使第一个块右对齐，以便提示跨越 2 行而不是 3 行。

Want it to be unaffected by the path

yauyy:
> I would do the following, move the second block to the first position and make the first block a right aligned one so the prompt spans 2 lines and not three.

Want it to be unaffected by the path

bend-n:
I think it needs to be in the `themes` folder, not `主题`

JanDeDobbeleer:
@bend-n ooh, good catch. I missed that.

bend-n (changes requested):
Needs to be in the `themes` folder.

yauyy:
> Needs to be in the `themes` folder.

Thanks for the reminder that a plugin causes

yauyy (commented):
update

PR #2552, opened 2022-07-20
fix(git): further fixes for submodule worktrees

pitkali:
filepath.Join() just concatenates paths. We have to check manually for
an absolute path and discard the first component to end up with the
correct path.

Also, fix the tests to demonstrate actual behaviour of git, not what
filepath.Join() was doing.

Additionally, note that on Windows git paths in gitdir use unix
separators, so search for `/worktrees/` instead of using system
separator, just like when processing all the other cases.

### Prerequisites

### Description

I have updated the code to actually handle absolute paths correctly. The tests seemed quite confused in particular, and needed some adjustments.

Interestingly, when testing on Windows, I discovered that prior attempt to use path separator when searching for `worktrees` in the path from Git was misguided as even my official Windows distribution of Git used `/` instead of the usual separator on Windows. This is probably why all the other searches ran on working forlder path also use `/`.

Tests pass on my Linux box. I also tested it in the wild both on Linux and Windows.

### DIFF
--- a/src/segments/git.go
+++ b/src/segments/git.go
@@ -196,11 +196,15 @@ func (g *Git) hasWorktree(gitdir *environment.FileInfo) bool {
 	// we need the parent folder to detect where the real .git folder is
 	ind = strings.LastIndex(g.workingFolder, "/.git/modules")
 	if ind > -1 {
-		g.rootFolder = filepath.Join(gitdir.ParentFolder, g.workingFolder)
+		if !filepath.IsAbs(g.workingFolder) {
+			g.rootFolder = filepath.Join(gitdir.ParentFolder, g.workingFolder)
+		} else {
+			g.rootFolder = g.workingFolder
+		}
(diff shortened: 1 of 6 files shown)


pitkali:
Right. The unit test fails on Windows, because the testcase uses only unix paths. The absolute paths would need `C:` prepended on Windows, so that they are absolute. Not quite sure how to do it in Go without "hacks."

JanDeDobbeleer:
@pitkali if the paths are always unix styled, I guess the check needs to be more explicit and validate for a leading `/`?

pitkali:
@JanDeDobbeleer I have finally sat down and fixed the cross-platform concerns, including the tests. It works for me on both Windows and Linux, and the tests pass on both as well.

JanDeDobbeleer:
@pitkali the error is a new linter check that fails. I'll fix that tomorrow morning, and will rebase your branch after.

PR #2531, opened 2022-07-16
Add Theme "wholespace"

wholespace:
### Prerequisites

- [x] I have read and understood the [contributing guide][CONTRIBUTING.md]
- [x] The commit message follows the [conventional commits][cc] guidelines
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added/updated (for bug fixes/features)

### Description

[CONTRIBUTING.md]: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/CONTRIBUTING.md
[cc]: https://www.conventionalcommits.org/en/v1.0.0/#summary

wholespace

PR #2564, opened 2022-07-24
Revert "Add Theme "wholespace""

wholespace:
Reverts JanDeDobbeleer/oh-my-posh#2531


JanDeDobbeleer:
@wholespace do you want to rename it? I didn't validate that indeed, my bad. Just rename the theme 😉

PR #2566, opened 2022-07-25
Disabling-feature support and PowerShell-specific fix for the transient prompt

lewis-yeung:
**[Superseded]** Part of this PR has been superseded by #2567.

### Prerequisites

### Description

1. Add functions for disabling the transient prompt feature in a PowerShell/zsh/fish session:

	- PowerShell: `Disable-PoshTransientPrompt`
	- zsh: `disable_poshtransientprompt`
	- fish: `disable_poshtransientprompt`

2. Fix a bug in PowerShell: if transient prompt is enabled and `PredictionViewStyle` for PSReadLine is set to `ListView`, once the command line is accepted with the suggestion list active, the text buffer below the current prompt line will persist, causing cluttered output of the command. Special thanks to @rashil2000 for providing this workaround in starship/starship#4143.

### DIFF
--- a/website/docs/configuration/transient.mdx
+++ b/website/docs/configuration/transient.mdx
@@ -76,11 +76,10 @@ properties below - defaults to `{{ .Shell }}> `
 }>
 <TabItem value="pwsh">
 
-
 Invoke Oh My Posh in your `$PROFILE` and add the following line below.
 
 ```powershell
-oh-my-posh init pwsh --config $env:POSH_THEMES_PATH/jandedobbeleer.omp.json | Invoke-Expression
+oh-my-posh init pwsh --config "$env:POSH_THEMES_PATH/jandedobbeleer.omp.json" | Invoke-Expression
 // highlight-start
 Enable-PoshTransientPrompt
 // highlight-end
@@ -95,7 +94,7 @@ Restart your shell or reload your `$PROFILE` using `. $PROFILE` for the changes 
 </TabItem>
 <TabItem value="cmd">
 
-Set the transient prompt in [clink][clink] to `always` in the `oh-my-posh.lua` script to enable the feature:
+You can run the command below to enable the feature permanently:
 
 ```shell
 clink set prompt.transient always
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer:
@lewis-yeung wait, I actually want to change how this works altogether. Rather than having a function, work with environment variables that are injected into the init script based on the presence of one of the prompts. For example.

```powershell
$env:POSH_TRANSIENT=::POSH_TRANSIENT::

Set-PSReadLineKeyHandler -Key Enter -ScriptBlock {
    if ($env:POSH_TRANSIENT -ne 1) {
        [Microsoft.PowerShell.PSConsoleReadLine]::AcceptLine()
        return
    }
    $previousOutputEncoding = [Console]::OutputEncoding
    try {
        $parseErrors = $null
        [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$null, [ref]$null, [ref]$parseErrors, [ref]$null)
        if ($parseErrors.Count -eq 0) {
            $script:TransientPrompt = $true
            [Console]::OutputEncoding = [Text.Encoding]::UTF8
            [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
        }
    } finally {
        [Microsoft.PowerShell.PSConsoleReadLine]::AcceptLine()
        [Console]::OutputEncoding = $previousOutputEncoding
    }
}
```

And inject it

website/docs/configuration/transient.mdx
```diff
@@ -123,11 +122,62 @@ Invoke Oh My Posh in `~/.config/fish/config.fish` and add the following line bel
...
+Disable-PoshTransientPrompt
+```
+
+</TabItem>
+<TabItem value="cmd">
+
+You cannot disable the feature when you are in a CMD session. You can **remove** the line below in the `oh-my-posh.lua` script to disable it permanently:
+
+```shell
+clink set prompt.transient always
```
rashil2000: Just a note: this is a not a Lua statement (and hence cannot be put in oh-my-posh.lua file). This command needs to be run in the shell session - it's a CMD command. And `clink set prompt.transient off` will turn transience off permanently. I think this needs to be changed in the docs too - https://ohmyposh.dev/docs/configuration/transient

lewis-yeung: 😅 Oops, I missed this. Thank you for the good catch.

lewis-yeung:
@JanDeDobbeleer I can imagine the approximate logic. So will you deprecate functions like `Enable-PoshTransientPrompt`? Will this be a breaking change? On the other hand, if we'd like to utilize an environment variable, we should do the same for other shells. This can be closed if you have made the final decision and I'll put up another PR for the bug fix in PowerShell.

JanDeDobbeleer:
> I can imagine the approximate logic. So will you deprecate functions like Enable-PoshTransientPrompt? Will this be a breaking change? On the other hand, if we'd like to utilize an environment variable, we should do the same for other shells. This can be closed if you have made the final decision and I'll put up another PR for the bug fix in PowerShell.

@lewis-yeung yes, all those functions can be deprecated in that way. It should not be breaking as I expect people who have a config with such a prompt to use it, and if not they can disable it by tweaking the environment variable. And yes, this needs to be consistent for all supported shells (where possible, as some shells require setting things explicitly).

lewis-yeung:
> I actually want to change how this works altogether. Rather than having a function, work with environment variables that are injected into the init script based on the presence of one of the prompts.

@JanDeDobbeleer Is this still on your list? Or are you working on it?

JanDeDobbeleer:
@lewis-yeung I was planning on doing this, yes. Was finishing something else, will move to that next.

JanDeDobbeleer:
@lewis-yeung I have a working version I want to test a bit more. Only wondering if I should go for a breaking change or not. We can either remove the functions (which breaks people's shell config scripts), or keep dummies in there which do not.

lewis-yeung:
@JanDeDobbeleer I think we can do it just like how we deprecated PowerShell module before. The functions now do nothing but inform users to tweak their configs. BTW, if we're gonna have a breaking change (new major version number), we can also remove the old CLI (e.g., the `prompt` subcommand and the `--init` flag) which I think is certainly outdated nowadays.

PR #2584, opened 2022-07-28
PowerShell-specific refactoring and tweak

lewis-yeung:
### Prerequisites

### Description

1. Refactor the logic of `prompt` function in the init script `omp.ps1` for PowerShell, and with this we now support the global template property `.Code` for the debug prompt in PowerShell, the same to any global properties in the future. This refactoring also handles the count of lines for three types of prompts properly, which fixes bugs of line overwriting, no line wrap, etc. for the debug prompt in a transient-prompt-enabled context, if the count of lines is different between the debug and the primary prompts.

2. A tweak in the workaround for the transient prompt in PowerShell.

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -125,12 +125,16 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
                     [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
                 }
             } finally {
+                # If PSReadline is set to display suggestion list, this workaround is needed to clear the buffer below
+                # before accepting the current commandline. The max amount of items in the list is 10, so 12 lines
+                # are cleared (10 + 1 more for the prompt + 1 more for current commandline).
                 if ((Get-PSReadLineOption).PredictionViewStyle -eq 'ListView') {
-                    # If PSReadline is set to display suggestion list, this workaround is needed to clear the buffer below
-                    # before accepting the current commandline. The max amount of items in the list is 10, so 12 lines
-                    # are cleared (10 + 1 more for the prompt + 1 more for current commandline).
-                    [Microsoft.PowerShell.PSConsoleReadLine]::Insert("`n" * [System.Math]::Min($Host.UI.RawUI.WindowSize.Height - $Host.UI.RawUI.CursorPosition.Y - 1, 12))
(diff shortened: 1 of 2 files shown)


src/shell/scripts/omp.ps1
```diff
@@ -265,57 +269,56 @@ Example:
...
-                    # retrieve info of the most recent error
-                    $global:Error[0] | Where-Object { $_ -ne $null } | Select-Object -ExpandProperty InvocationInfo
-                } catch { $null }
-                # check if the last command caused the last error
-                if ($null -ne $invocationInfo -and $lastHistory.CommandLine -eq $invocationInfo.Line) {
-                    $script:ErrorCode = 1
-                } elseif ($realLASTEXITCODE -is [int] -and $realLASTEXITCODE -ne 0) {
-                    # native app exit code
-                    $script:ErrorCode = $realLASTEXITCODE
+        } else {
```
JanDeDobbeleer: I don't like to use if/else, makes everything rather difficult to read. So please refactor this to early returns. the whole block is now indented. We could also create an internal function which returns the prompt type.

website/docs/configuration/debug-prompt.mdx
```diff
@@ -49,6 +49,7 @@ properties below - defaults to `[DBG]: `
 - `.Shell`: `string` - the current shell name
 - `.UserName`: `string` - the current user name
 - `.HostName`: `string` - the host name
+- `.Code`: `int` - the last exit code
```
JanDeDobbeleer, line 52: this will only be valid for PowerShell now, right?

JanDeDobbeleer (changes requested):

website/docs/configuration/debug-prompt.mdx
lewis-yeung: Yes, since the debug prompt feature is only available for PowerShell as is described at the beginning of the page.

> This feature only works in `powershell` for the time being.

src/shell/scripts/omp.ps1
lewis-yeung: I have updated the refactoring. Also, I wrote the logic of getting the error code into a function and used early returns instead of "else" statements that you dislike. 😄

JanDeDobbeleer: You are the 👑

PR #2588, opened 2022-07-28
fix(pwsh): prevent text from disappearing when a tooltip is rendered

lewis-yeung:
### Prerequisites

### Description

This workaround has resolved a problem in PowerShell. The following are steps to reproduce it:

1. Add a tooltip in the theme config (e.g., I use `git` as the tip in this case).
2. Enable tooltips by `Enable-PoshTooltips` in PowerShell.
3. **Enter a few spaces at the beginning of a command line,** then the tip `git`, and move the cursor back to front of the first character `g`.
4. Insert a space by pressing <kbd>Space</kbd>, which makes the tooltip **actually rendered**. Now the text `git` after the cursor disappears.
5. Now if we continue to insert a **non-space** character, the disappeared content can be reproduced.

A similar bug happens if the suggestion list is enabled (`Set-PSReadLineOption -PredictionViewStyle ListView`) and has been showed before we press <kbd>Space</kbd>.

A simple demo:

https://user-images.githubusercontent.com/83903009/181591180-0069c502-de8a-45a4-a9e3-382010404d3f.mp4

In this demo, the suggestion list should persist when the tooltip is rendered, however it didn't.


JanDeDobbeleer:
@lewis-yeung always such good PR's, love it!

PR #2610, opened 2022-08-04
feat(ipify): added offline response

the-eduardo:
### Prerequisites

### Description
Added offline response for IPify when the HTTPRequest returns an error (httpTimeout)
### Why do that?
Because the  IPify segment can mainly be used for check online connection, personally it seems more useful if I can see when I'm offline than just returning 'null'/nothing to the shell.
image

If approved I'll update the docs and include the description of this feature.


src/segments/ipify.go
```diff
@@ -47,7 +47,7 @@ func (i *IPify) getResult() (string, error) {
 
 	body, err := i.env.HTTPRequest(url, nil, httpTimeout)
 	if err != nil {
-		return "", err
+		return "OFFLINE", err
```
JanDeDobbeleer, line 50: if we make this a const we can also reuse it in the tests.

src/segments/ipify.go
```diff
@@ -20,12 +20,12 @@ func (i *IPify) Template() string {
 }
 
 func (i *IPify) Enabled() bool {
-	ip, err := i.getResult()
+	i.IP, _ = i.getResult()
+	/*i.IP = ip
```
JanDeDobbeleer, line 24: this needs to be deleted ;-)

JanDeDobbeleer (changes requested):

PR #2630, opened 2022-08-07
fix(shell): quote paths properly in init scripts

lewis-yeung:
### Prerequisites

### Description

Inspired by [Starship](https://github.com/starship/starship/blob/9220df70d965ff43f7baa75be4c622c6963a7652/src/init/mod.rs#L38-L87), I implemented functions to quote the OMP executable path and the config file path (if necessary) and escape any special characters (if necessary) in the syntax of specific shell/script so that we can safely replace placeholders like `::OMP::` in a script with the quoted strings. This resolves a problem that happens especially when simply using single quotes to quote a path including any single quotes.

BTW, I also fixed the path separator inconsistency (Windows only) in the output of PowerShell function `Get-PoshThemes`.

### DIFF
--- a/src/shell/init.go
+++ b/src/shell/init.go
@@ -3,6 +3,7 @@ package shell
 import (
 	_ "embed"
 	"path/filepath"
+	"runtime"
 	"strconv"
 
 	"fmt"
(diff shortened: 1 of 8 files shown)


src/shell/init.go
```diff
@@ -51,25 +52,111 @@ func getExecutablePath(env environment.Environment) (string, error) {
 	// On Windows, it fails when the excutable is called in MSYS2 for example
 	// which uses unix style paths to resolve the executable's location.
 	// PowerShell knows how to resolve both, so we can swap this without any issue.
-	executable = strings.ReplaceAll(executable, "\\", "/")
-	switch env.Flags().Shell {
-	case BASH, ZSH:
-		executable = strings.ReplaceAll(executable, " ", "\\ ")
-		executable = strings.ReplaceAll(executable, "(", "\\(")
-		executable = strings.ReplaceAll(executable, ")", "\\)")
+	if runtime.GOOS == "windows" {
```
JanDeDobbeleer: you can reuse the `WINDOWS` const we have rather than the string.

src/shell/init.go
```diff
@@ -51,25 +52,111 @@ func getExecutablePath(env environment.Environment) (string, error) {
...
+			b.WriteRune(r)
+		case '\\', '\'':
+			b.WriteByte('\\')
+			b.WriteRune(r)
+		default:
+			b.WriteRune(r)
+			normal = true
+		}
+		if !normal {
+			needQuoting = true
```
JanDeDobbeleer, line 127: Can't this be moved inside the default case and thus not need the `normal` variable? If it reaches the default once, `needQuoting` will always be true for the remaining of the loop, we only set it to true.

JanDeDobbeleer (changes requested):

src/shell/init.go
lewis-yeung: I use the `normal` variable because we don't have to quote a string that doesn't contain any special characters. In this logic, only if all runes match the `default` case can we confirm the string needs no quoting. I have another way where we can count the non-special characters and check if the quantity is equal to the total. Not so graceful, though. Do you have any ideas on how to simplify this logic?

lewis-yeung: @JanDeDobbeleer So we just keep the original, right?

src/shell/scripts/omp.ps1
```diff
@@ -286,27 +286,34 @@ Example:
     }
 
     function Update-PoshErrorCode {
-        $script:ExecutionTime = -1
         $lastHistory = Get-History -ErrorAction Ignore -Count 1
+
+        # default values
+        $script:ErrorCode = 0
```
lewis-yeung, line 292: @JanDeDobbeleer If you move this line here, a non-zero error code cannot persist after empty hits, which breaks the original behavior.

JanDeDobbeleer: Ah damn it. I did validate it, but not for that use-case. Might be in line with the other shells now though. The problem is, I'm not available for the next 3 weeks. Finally taking a holiday.

lewis-yeung: @JanDeDobbeleer We should revert this and find another way to resolve #2633 later.

Issue #2640, opened 2022-08-09
pwsh: validate execution time type still happen in 8.29.2


vunhatchuong:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

I update oh-my-posh from Scoop to 8.29.2 after seeing the debug. But it still happen
![image](https://user-images.githubusercontent.com/20581504/183614824-6d3502dc-1056-470f-9808-2edef7899492.png)

### Theme

Theme: [bubblesline.omp.json](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/bubblesline.omp.json)

Config:```
[console]::InputEncoding = [console]::OutputEncoding = New-Object System.Text.UTF8Encoding
oh-my-posh init pwsh | Invoke-Expression
$omp_config = Join-Path $PSScriptRoot ".\bubblesline.omp.json"
oh-my-posh init pwsh --config $omp_config | Invoke-Expression
```

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

```Shell
Version: 8.29.2

Segments:

ConsoleTitle(false) -   0 ms -
path(true)     -   0 ms -
git(false)     -   5 ms -
python(false)  -   0 ms -
go(false)      -   0 ms -
node(false)    -   0 ms -
ruby(false)    -   0 ms -
java(false)    -   0 ms -
julia(false)   -   0 ms -
battery(true)  -  14 ms -
session(true)  -   0 ms -

Run duration: 32.6579ms

Cache path: C:\Users\ronny\AppData\Local\oh-my-posh

Config path: C:\Users\ronny\.config\powershell\bubblesline.omp.json

Logs:

2022/08/09 16:25:01 debug: Getenv
C:\Users\ronny\AppData\Local
2022/08/09 16:25:01 Getenv duration: 0s, args: LOCALAPPDATA
2022/08/09 16:25:01 CachePath duration: 514.6µs, args:
2022/08/09 16:25:01 debug: Getenv
C:\Users\ronny\.config\powershell\bubblesline.omp.json
2022/08/09 16:25:01 Getenv duration: 0s, args: POSH_THEME
2022/08/09 16:25:01 Shell duration: 10.3613ms, args:
2022/08/09 16:25:01 resolveConfigPath duration: 10.3613ms, args:
2022/08/09 16:25:01 Init duration: 11.3887ms, args:
[...]

Issue #2633, opened 2022-08-08
Error: invalid argument "" for "--execution-time" flag: strconv.ParseFloat: parsing "": invalid syntax


muchisx:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

This error appears entering any command with the integrated vscode powershell 7

Error: invalid argument "" for "--execution-time" flag: strconv.ParseFloat: parsing "": invalid syntax                                                                                                                                                                               

Flags:
      --command string         tooltip command
  -e, --error int              last exit code
      --eval                   output the prompt for eval
      --execution-time float   timing of the last command
  -h, --help                   help for print
  -p, --plain                  plain text output (no ANSI)
      --pswd string            current working directory (according to pwsh)
      --pwd string             current working directory
      --shell string           the shell to print for
      --shell-version string   the shell version
  -s, --stack-count int        number of locations on the stack
  -w, --terminal-width int     width of the terminal

Global Flags:
  -c, --config string   config (required)

I have not modified anythinc in the theme, just installed it normally as following the guides.

This error also is reproduced in Windows Terminal 

![image](https://user-images.githubusercontent.com/34131801/183523902-fcd53daa-d576-40e4-937e-b21471013ff4.png)

![image](https://user-images.githubusercontent.com/34131801/183523941-16e29828-5242-4571-9ecd-cb6e8b590478.png)

### Theme

Atomic theme
https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/atomic.omp.json

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

```Shell
Version: 8.29.0

Segments:

ConsoleTitle(false)  -   0 ms -
shell(true)          -   0 ms - ╭─ pwsh
root(false)          -   0 ms -
path(true)           -   0 ms -  project-2
git(true)            -  76 ms -    master ≡  ~2
executiontime(true)  -   0 ms -  羽0ms⠀
node(true)           -  30 ms -     17.9.0
python(false)        -   0 ms -
java(false)          -   0 ms -
dotnet(false)        -   0 ms -
go(false)            -   0 ms -
rust(false)          -   0 ms -
dart(false)          -   0 ms -
angular(false)       -   0 ms -
nx(false)            -   0 ms -
julia(false)         -   0 ms -
ruby(false)          -   1 ms -
azfunc(false)        -   0 ms -
aws(false)           -   0 ms -
[...]

Issue #2639, opened 2022-08-09
Prompt history broken since 8.27.0


jorisroovers:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

Hi Jan, big fan, thanks for creating and maintaining oh-my-posh!

Since a few versions, my prompt history breaks. That is, when hitting return, previous prompt lines no longer contain my full prompt, only the current working directory.

I've been able to pinpoint this to oh-my-posh 8.27, I believe this screenshot shows it well:

<img width="875" alt="CleanShot 2022-08-09 at 10 03 57@2x" src="https://user-images.githubusercontent.com/575086/183599317-964f9ccd-48af-4f03-bc3c-d5a401054096.png">

Note that on 8.26 (first 3 lines) the full commandline prompt shows for each command that's been executing, but after switching to 8.27 only `foo` shows, i.e. the current working directory. The rest of the prompt only shows on the active prompt line, but then changes to only show `foo` right after executing a command.

Apologies if this is known change that requires updating on my side, I wasn't able to tell easily from the 8.27 code changes.
This does sound somewhat similar to #2636 but not sure.

I'm on `zsh 5.8.1 (x86_64-apple-darwin21.0)`.

### Theme

{
    "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
    "blocks": [
        {
            "type": "prompt",
            "alignment": "left",
            "segments": [
                {
                    "background": "p:yellow",
                    "foreground": "p:black",
                    "leading_diamond": "\ue0b6",
                    "properties": {
                        "display_host": false
                    },
                    "style": "diamond",
                    "template": " {{ if .SSHSession }}\uf817 {{ .UserName }}@{{ .HostName }}{{ end }}",
                    "trailing_diamond": "\ue0b0",
                    "type": "session"
                },
                {
                    "type": "python",
                    "style": "diamond",
                    "trailing_diamond": "\ue0b0",
                    // "powerline_symbol": "\uE0B0",
                    // "trailing_diamond": "<transparent,background>\ue0b0</>",
                    "foreground": "#100e23",
                    "background": "p:orange",
                    "template": " \uE235 ({{ .Venv }})"
                },
                {
                    "type": "path",
                    "background": "#0087D8",
                    "foreground": "p:white",
                    "powerline_symbol": "\ue0b0",
                    "properties": {
                        "folder_separator_icon": "/",
[...]

PR #2643, opened 2022-08-09
fix(pwsh): set initial value for `$script:ExecutionTime`

lewis-yeung:
### Prerequisites

### Description

@muchisx @ander-Sgr @vunhatchuong123 I cannot validate this fix because I've never been able to reproduce the bug mentioned in #2633 and #2640 on my end. 😓 Can you do a test when this goes live?

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -15,6 +15,7 @@ function global:Get-PoshStackCount {
 
 New-Module -Name "oh-my-posh-core" -ScriptBlock {
     $script:ErrorCode = 0
+    $script:ExecutionTime = 0
     $script:OMPExecutable = ::OMP::
     $script:ShellName = "::SHELL::"
     $script:PSVersion = $PSVersionTable.PSVersion.ToString()
@@ -286,17 +287,14 @@ Example:
     }
 
     function Update-PoshErrorCode {
-        $script:ExecutionTime = -1
         $lastHistory = Get-History -ErrorAction Ignore -Count 1
         # error code should be updated only when a non-empty command is run
         if (($null -eq $lastHistory) -or ($script:LastHistoryId -eq $lastHistory.Id)) {
+            $script:ExecutionTime = 0
             return
         }
         $script:LastHistoryId = $lastHistory.Id
         $script:ExecutionTime = ($lastHistory.EndExecutionTime - $lastHistory.StartExecutionTime).TotalMilliseconds
-        if (-not ($script:ExecutionTime -is [int])) {
-            $script:ExecutionTime = 0
-        }
         if ($script:OriginalLastExecutionStatus) {
             $script:ErrorCode = 0
             return


muchisx:
@lewis-yeung Yes, will do - Will be using PowerShell 7.2.5 which is the one im getting the error from initially.

muchisx:
@lewis-yeung @JanDeDobbeleer 

image

I didn't know where to comment this but yeah this PR fixed it for me! Thank you !

ander-Sgr:
It's fixed 😄  thx..!

vunhatchuong:
It did not completely fix the bug for me. It doesn't show the `--execution-time` error anymore. 

But the terminal output is slow every time I write a command and the UI showing `pwsh` instead of `ronny` in my case

image

muchisx:
Well, I'm a very fresh user (literally installed ohmyposh a like an hour before posting the issue) so I don't really know what is the expected execution speed, but when I press enter for a command it does have a liiiiiiiittle delay.

also previously, when the error was being printed, the execution time module worked, it showed the execution time of the command in the prompt - but still outputed the error, so maybe that execution time evaluation function is running more than one time somehow?

lewis-yeung:
> But the terminal output is slow every time I write a command and the UI showing `pwsh` instead of `ronny` in my case

@vunhatchuong123 That's not a bug. See [my comment in another issue](https://github.com/JanDeDobbeleer/oh-my-posh/issues/2639#issuecomment-1209085547). However, if you are using a preset "bubblesline" theme without any modifications, it should not behave this way since "bubblesline" does not contain a `transient_prompt` config.

vunhatchuong:
> > But the terminal output is slow every time I write a command and the UI showing `pwsh` instead of `ronny` in my case
> 
> @vunhatchuong123 That's not a bug. See [my comment in another issue](https://github.com/JanDeDobbeleer/oh-my-posh/issues/2639#issuecomment-1209085547). However, if you are using a preset "bubblesline" theme without any modifications, it should not behave this way since "bubblesline" does not contain a `transient_prompt` config.

Yeah I use the default one.

lewis-yeung:
@vunhatchuong123 Now I know why the bug appeared on your end but not mine. With a previous version of 8.27.0, I tried writting two `oh-my-posh init ...` lines like what you did (also mentioned in [another comment](https://github.com/JanDeDobbeleer/oh-my-posh/issues/2640#issuecomment-1209157882)) in the PowerShell profile, and finally I can reproduce the bug! In other words, this is actually the root cause. For the first `oh-my-posh init ...` command that has no config file specified, by default, Oh My Posh uses [a "default" theme that contains a `transient_prompt` config](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/default.omp.json#L184-L188) so that the initialization script enables the transient prompt feature in the current PowerShell session. Once enabled, even though a second `oh-my-posh init ...` command uses a config without a `transient_prompt` section, the behaviors cannot be reverted.

The following I'm gonna explain may be a little more complicated, but it's the logic of how the bug is introduced:

1. The first initialization uses a "default" config **with the transient prompt feature enabled**, so it will bind a custom function to <kbd>Enter</kbd> using `Set-PSReadLineKeyHandler`:

	https://github.com/JanDeDobbeleer/oh-my-posh/blob/35efd3cf008e306ff9d58e9414d778010bd906c7/src/shell/scripts/omp.ps1#L119-L146

2. Then there comes the second initialization that uses a config **with the transient prompt feature disabled**. However, it does not bind a new function to <kbd>Enter</kbd> since the placeholder of `::TRANSIENT::` is replaced with `false` before the initialization starts, and this prevents the above code from being executed.

3. The `prompt` function defined in the first dynamic module **doesn't actually output a prompt yet**. The `Update-PoshErrorCode` function never runs so **the `$script:ExecutionTime` variable doesn't get an initial value**:

	https://github.com/JanDeDobbeleer/oh-my-posh/blob/35efd3cf008e306ff9d58e9414d778010bd906c7/src/shell/scripts/omp.ps1#L288-L289

	that means it's always **undefined**.

4. Now the key point: due to the encapsulation and isolation features of PowerShell modules, the two dynamic modules don't share the same `$script:ExecutionTime` variable. Notice that the custom function bound to <kbd>Enter</kbd> is defined in the first dynamic module, it cannot reach any functions or variables defined in the second one. Even though the second initialization has "removed" the first dynamic module, the previously defined key handler function can still reach functions and variables defined in the first module (the logic is similar to closures), so **the command below invokes the `prompt` function defined in the first module instead of the second one**:

	https://github.com/JanDeDobbeleer/oh-my-posh/blob/35efd3cf008e306ff9d58e9414d778010bd906c7/src/shell/scripts/omp.ps1#L128

	since each press on <kbd>Enter</kbd> renders a **transient** prompt, the `Update-PoshErrorCode` function will still never run in that `prompt`:

	https://github.com/JanDeDobbeleer/oh-my-posh/blob/35efd3cf008e306ff9d58e9414d778010bd906c7/src/shell/scripts/omp.ps1#L323-L325

	thus **the `$script:ExecutionTime` variable of the first module will never be populated**, and the **undefined** variable causes OMP to receive an empty value for the `--execution-time` flag:

	https://github.com/JanDeDobbeleer/oh-my-posh/blob/35efd3cf008e306ff9d58e9414d778010bd906c7/src/shell/scripts/omp.ps1#L334

	so, a bug appears and you'll see the error message. 🤦🏻‍♂️

**EDITED:**

A similar situation can occur on `$script:ErrorCode`.

**Conclusion:** Starting from v8.27.0, it cannot work properly if we do a second initialization in the same PowerShell session to change the theme. Anyway, this is at least not a common use case so it doesn't really hurt. 😅

vunhatchuong:
Yep remove one `oh-my-posh init ...` fix it

JanDeDobbeleer:
@lewis-yeung to change the theme it's sufficient to adjust the `POSH_THEME` variable. But still edit the profile 😉

lewis-yeung:
> to change the theme it's sufficient to adjust the `POSH_THEME` variable. But still edit the profile

@JanDeDobbeleer Yeah, I know this. But if we have initialized with a theme that does not contain a transient prompt config, and then change the `POSH_THEME` environment variable to use one that does, the transient prompt feature won't work due to the current logic of initialization. So, restarting PowerShell is always required. 😄

JanDeDobbeleer:
@lewis-yeung true, but that's not a major issue in my opinion.

lewis-yeung:
> true, but that's not a major issue in my opinion.

@JanDeDobbeleer Agree. After all, it's not a common use case. 😄

PR #2663, opened 2022-08-15
feat(wakatime): parse url as template

### LINKED ISSUE #2662: Restructure Wakatime Properties and Allow Environment Variable Parsing on Properties
### Code of Conduct

Hello!

This is a two feature request that revolves around the Wakatime segment. Assuming that one modifies/implements the other.

Currently, I was building my own custom theme and I stumbled upon the Wakatime segment. Upon evaluation, I feel like there's a need for adjustments.

## Wakatime Properties Refactor Sugggestion

Consume the following below: (From https://ohmyposh.dev/docs/segments/wakatime#sample-configuration)

```json
{
    "type": "wakatime",
    "style": "powerline",
    "powerline_symbol": "\uE0B0",
    "foreground": "#ffffff",
    "background": "#007acc",
    "properties": {
        "url": "https://wakatime.com/api/v1/users/current/summaries?start=today&end=today&api_key=API_KEY",
        "cache_timeout": 10,
        "http_timeout": 500
    }
},
```
From this JSON structure, we can see that the `URL` property is exposed, shouldn't it be embedded inside the engine? Also, I don't think there are other endpoints that send the summary context but other than that, I could think of this segment as an HTTP Segment that displays custom data.

Properties such as `cache_timeout` and `http_timeout` seem to be inconsistent in terms of time metric. According to documentation, `http_timeout` defaults to 20**ms**, while the `cache_timeout` implicates its default (_10m_) under the time metric of **m** which goes by minutes. I think they should have a consistent time metric such as milliseconds or seconds to avoid confusion.

With those insights, I think we can re-structure the wakatime segment under the following:
```json
{
    "type": "wakatime",
    "style": "powerline",
    "powerline_symbol": "\uE0B0",
    "foreground": "#ffffff",
    "background": "#007acc",
    "properties": {
        "api_key": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX",  // Also allow Env properties to be rendered here.
        "cache_timeout_seconds": 10,
        "http_timeout_seconds": 500,
        "range": "" // string, can be daily, weekly, or monthly. Achievable by evaluating `start` and `end` URL parameters and with ISO dateformat.
    }
},
```

The new JSON structure now properly shows [1] a clean way to insert an `API_KEY`, [2] time metric now consistent between timeout properties, and [3] additional attributes (highlighting start and end datetime) correlating to the summaries API URL parameters were added, allowing for more customization or query of data. (See more information here about other parameters to include in the properties: https://wakatime.com/developers#summaries)

Also, if timeout property proposal change is implausible; another alternative suggestion for [2] would be indicating the time metric as its suffix (for instance, `cache_timeout` -> `cache_timeout_min` or `http_timeout` -> `http_timeout_ms`), so that the implication of the time metric is there.

## Environment Variable Parsing at Properties

As `API_KEY` was highlighted in the new JSON structure, another suggestion or feature that I want to see is the Environment Variable Parsing under the _Properties_ of a segment. So far, I have never seen any string parsing example under the properties of any other segment examples and I think the introduction of env parsing would allow other themes to extend other services to display as segments such as Brewfather, OpenWeatherMap(OWM), Withings, etc, other than the WakaTime.

We need this (speaking for myself) this feature as per my statement in the first section to allow services to be rendered as segments and to be extended in themes without worrying about accidental committing changes with a personal token.

Pictures below are the attempts to fetch data with Wakatime with the Environment Variable appended in the URL string.

image

Upon checking the `debug`, it shows HTTP 401 `Unauthorized`.
image

To verify that the env var reference has a value, the figure below shows its preceding values.
image

To verify that the (transient) prompt was able to parse its value, the following figure/s show its declaration (as highlighted) and output.
image

Output:
image

Output concludes that properties cannot parse environment values. Other than that, I cannot suspect if sprig cannot be parsed under properties, but I think it's likely.

In the end, all of these are a suggestion and I may be asking too much at this point, feel free to be against it.

JanDeDobbeleer:
resolves #2662

### Prerequisites

### Description


JanDeDobbeleer:
I'll need my laptop for this one. Blocked until my return.

PR #2670, opened 2022-08-16
Theme: fully remove session segment from slim/slimfat for default user

manningli:
### Prerequisites

### Description

Prior to this PR, the Slim and Slimfat themes leave artifacts in the Session segment when the default user is active which should remove this segment. 

Slimfat:
image

Slim:
image

After this change, the segment is properly removed:
image


themes/slim.omp.json
```diff
@@ -178,10 +178,10 @@
...
   "console_title_template": "{{if .Root}}root :: {{end}}{{.Shell}} :: {{.Folder}}",
   "final_space": true,
-    "transient_prompt": {
+  "transient_prompt": {
     "background": "transparent",
     "foreground": "#7a7a7a",
     "template": " ~#@\u276f "
   },
   "version": 2
-}
```
lewis-yeung, line 187: We should keep this line feed. The same to the other. :)

themes/slim.omp.json
```diff
@@ -178,10 +178,10 @@
...
+}
```
lewis-yeung: We should keep this line feed. The same to the other. :)

themes/slim.omp.json
```diff
@@ -33,7 +33,7 @@
           "background": "#2f2f2f",
           "foreground": "#fafafa",
           "style": "diamond",
-          "template": " {{ if ne .Env.POSH_SESSION_DEFAULT_USER .UserName }}<#77f5d6>{{ .UserName }}</><#7a7a7a>\uf1fa</><#2EEFBF>{{ end }}{{ .HostName }}</><#7a7a7a> \ue0b1</>",
+          "template": " {{ if ne .Env.POSH_SESSION_DEFAULT_USER .UserName }}<#77f5d6>{{ .UserName }}</><#7a7a7a>\uf1fa</><#2EEFBF>{{ .HostName }}</><#7a7a7a> \ue0b1</>{{ end }}",
```
lewis-yeung: The host name should be displayed even for the default user, shouldn't it? The same to the other. :)

```suggestion
          "template": " {{ if ne .Env.POSH_SESSION_DEFAULT_USER .UserName }}<#77f5d6>{{ .UserName }}</><#7a7a7a>\uf1fa</>{{ end }}<#2EEFBF>{{ .HostName }}</><#7a7a7a> \ue0b1</>",
```

JanDeDobbeleer: @manningli any feedback?

manningli: Sorry for the slow reply, currently the host name doesn't display for default user in the slimfat theme so I assumed the intent was to remove the segment completely. 

If it's preferred to still keep the host name, moving the `{{ end }}` statement to before the opening color tag would allow the host name to be the correct color and would remove the closing tag artifact.

I don't have a strong preference either way, but after going through a few new machines at work recently, having to modify the theme to remove the `<\>` from the slimfat theme each time felt like some change is needed. Regardless of which direction is preferred, I'll be available to make the update quicker now.

manningli: Just noticed I missed @lewis-yeung suggested change in the comment mirroring what I just said... I'll commit that so host name remains and make the same change to slimfat for consistency.

themes/slim.omp.json
manningli: Done.

PR #2709, opened 2022-08-26
feat(gitversion): add gitversion segment

Lemorz56:
### Prerequisites

### Description
This PR adds a new segment to display the SemVer version from GitVersion if the directory/repo you're in has that available.

### DIFF
--- a/src/segments/gitversion.go
+++ b/src/segments/gitversion.go
@@ -1,0 +1,112 @@
+package segments
+
+import (
+	"encoding/json"
+	"errors"
+	"oh-my-posh/platform"
+	"oh-my-posh/properties"
+)
+
+type gitVersion struct {
+	Major                           int    `json:"Major"`
+	Minor                           int    `json:"Minor"`
+	Patch                           int    `json:"Patch"`
+	PreReleaseTag                   string `json:"PreReleaseTag"`
+	PreReleaseTagWithDash           string `json:"PreReleaseTagWithDash"`
+	PreReleaseLabel                 string `json:"PreReleaseLabel"`
+	PreReleaseLabelWithDash         string `json:"PreReleaseLabelWithDash"`
+	PreReleaseNumber                int    `json:"PreReleaseNumber"`
+	WeightedPreReleaseNumber        int    `json:"WeightedPreReleaseNumber"`
+	BuildMetaData                   int    `json:"BuildMetaData"`
+	BuildMetaDataPadded             string `json:"BuildMetaDataPadded"`
+	FullBuildMetaData               string `json:"FullBuildMetaData"`
+	MajorMinorPatch                 string `json:"MajorMinorPatch"`
+	SemVer                          string `json:"SemVer"`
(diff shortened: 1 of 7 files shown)


website/docs/segments/gitversion.mdx
```diff
@@ -0,0 +1,50 @@
...
+sidebar_label: GitVersion
+---
+
+## What
+
+Display the [GitVersion][gitversion] version.
+I _strongly_ recommend using [GitVersion Portable][gitversion-portable] for this.
+
+:::caution
+The GitVersion CLI can be a bit slow causing the prompt to feel slow.
```
JanDeDobbeleer: I'm wondering if there's a way to cache the information. Maybe by folder and interval?

Lemorz56: Good thought, do you have any example of how this is done in another segment? I tried to implement it a bit like NBGV is implemented since it's quite similar.

JanDeDobbeleer: @Lemorz56 there's a good pointer [here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/c93d637128b47bd235776a4833abb59da28d44b2/src/http/request.go#L28).

Lemorz56: I've tried a lot with testing both via the tests but also trying to install and test it manually, but I can't really figure out what is wrong with the way I implemented the cache. Ive pushed it and I will continue to debug and see where my misstake is, but if you have the time and can, i'd love if you can take a look at the code too 😄

src/segments/gitversion.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (n *Gitversion) Enabled() bool {
+	gitversion := "gitversion"
+	if !n.env.HasCommand(gitversion) {
+		return false
+	}
+	dir := n.env.Pwd()
+	result, err := getCacheValue(n, dir)
+
+	if result.FullSemVer == "" {
```
JanDeDobbeleer: we can do an early return here instead of indenting the full block.

```suggestion
	if len(result.FullSemVer) != 0 {
	    return err == nil
	}
```

JanDeDobbeleer:
@Lemorz56 I'm currently refactoring this so it can be merged. Stay tuned.

Lemorz56:
@JanDeDobbeleer Thanks. Sorry for not replying to your earlier comment, I must have forgotten about fixing that. 

I'm guessing when you are refactoring you will also touch the tests which might be good. I thought they were kinda tricky to validate.

JanDeDobbeleer:
@Lemorz56 yes, I'm almost done. Will finish up tomorrow and push the changes so you can also have a look.

JanDeDobbeleer:
@Lemorz56 I'm done, tests work and I expect this all to work correctly but didn't do an integration test end to end.

JanDeDobbeleer:
Tested, we can go live.

JanDeDobbeleer:
@all-contributors please add @Lemorz56 for code,design,doc

PR #2788, opened 2022-09-14
Add 'Networks' segment (Windows only)

LNKLEO:
### Prerequisites

### Description


JanDeDobbeleer:
@LNKLEO what's the benefit of this one over the wifi segment?

LNKLEO:
wifi segment only provide wifi network, my segment provide all networks, including wifi, cellular, wired network, even if a network over bluetooth.

image

besides above, it also able to identify FireWire 1394, DSL, etc. I just grouped them as 'other', for a low usage in personal area.

JanDeDobbeleer:
@LNKLEO exactly, so the wifi segment is redundant in that case. I would remove the wifi segment in favor of a network segment.

LNKLEO:
yes, you are right, I have removed wifi, and will maintenance network segment when need

JanDeDobbeleer:
@LNKLEO the review is on my list when I get the time (as it larger, I will probably rework myself, so no need to keep rebasing).

JanDeDobbeleer:
@LNKLEO I was trying to look at the actual code here to see how we could improve it as it's very inflexible (and complicated). I'm going to solve that challenge by creating multiple segments per connection type. This also allows for extensibility per platform.

JanDeDobbeleer:
See #2857

PR #2856, opened 2022-09-30
fix(path): parse path correctly

lewis-yeung:
### Prerequisites

### Description

Refactor most of the logic for rendering a `path` segment. For a full path, parse it to a root and a path, replace mapped locations and join results with separators to get a clean path.

With these changes, we can handle various types of paths (Unix, Windows, PSDrive and UNC). I've considered lots of annoying edge cases, especially a Windows drive root and a UNC path. The following bugs in a `path` segment will be fixed:

1. In PowerShell, when the CWD is a Windows root path, e.g., `C:\`, the `.Parent` should be empty as expected, however it is actually `C:.\`.
2. In PowerShell, mapped locations don't work for some PSDrive paths like `HKLM:\SOFTWARE` (Windows only) and `Function:/`.
3. Using `\` as the path separator for a mapped location can cause a path misidentification on Unix-like systems. In fact, `\` is a valid character in a file/directory name on Linux/macOS, which means it cannot be used to separate path elements.
4. The custom folder separator (`folder_separator_icon`) does not work in the `.Parent`.
5. The `.Parent` is not properly set based on the actual value of the `.Path`. E.g., when the CWD is the Home path, it is always set to empty even if the Home path is not replaced with a single tilde (`~`). To make sense, a mapped location replacement, if any, should be done before we start to extract the parent path.

### DIFF
--- a/src/environment/shell.go
+++ b/src/environment/shell.go
@@ -337,11 +337,16 @@ func (env *ShellEnvironment) Pwd() string {
 	}
 	correctPath := func(pwd string) string {
 		// on Windows, and being case sensitive and not consistent and all, this gives silly issues
-		driveLetter := regex.GetCompiledRegex(`^[a-z]:`)
-		return driveLetter.ReplaceAllStringFunc(pwd, strings.ToUpper)
+		if env.GOOS() == WINDOWS {
+			driveLetter := regex.GetCompiledRegex(`^[a-z]:`)
+			return driveLetter.ReplaceAllStringFunc(pwd, strings.ToUpper)
+		}
+		return pwd
 	}
 	if env.CmdFlags != nil && env.CmdFlags.PWD != "" {
-		env.cwd = correctPath(env.CmdFlags.PWD)
+		// ensure a clean path
+		root, path := ParsePath(env, correctPath(env.CmdFlags.PWD))
+		env.cwd = root + path
 		return env.cwd
 	}
 	dir, err := os.Getwd()
@@ -735,6 +740,9 @@ func (env *ShellEnvironment) TemplateCache() *TemplateCache {
 	pwd := env.Pwd()
 	tmplCache.PWD = ReplaceHomeDirPrefixWithTilde(env, pwd)
 	tmplCache.Folder = Base(env, pwd)
+	if env.GOOS() == WINDOWS && strings.HasSuffix(tmplCache.Folder, ":") {
+		tmplCache.Folder += `\`
+	}
 	tmplCache.UserName = env.User()
 	if host, err := env.Host(); err == nil {
 		tmplCache.HostName = host
@@ -766,19 +774,21 @@ func (env *ShellEnvironment) DirMatchesOneOf(dir string, regexes []string) (matc
 }
 
 func dirMatchesOneOf(dir, home, goos string, regexes []string) bool {
-	normalizedCwd := strings.ReplaceAll(dir, "\\", "/")
-	normalizedHomeDir := strings.ReplaceAll(home, "\\", "/")
+	if goos == WINDOWS {
+		dir = strings.ReplaceAll(dir, "\\", "/")
+		home = strings.ReplaceAll(home, "\\", "/")
+	}
 
 	for _, element := range regexes {
 		normalizedElement := strings.ReplaceAll(element, "\\\\", "/")
 		if strings.HasPrefix(normalizedElement, "~") {
-			normalizedElement = strings.Replace(normalizedElement, "~", normalizedHomeDir, 1)
+			normalizedElement = strings.Replace(normalizedElement, "~", home, 1)
 		}
 		pattern := fmt.Sprintf("^%s$", normalizedElement)
 		if goos == WINDOWS || goos == DARWIN {
 			pattern = "(?i)" + pattern
 		}
-		matched := regex.MatchString(pattern, normalizedCwd)
+		matched := regex.MatchString(pattern, dir)
 		if matched {
 			return true
 		}
@@ -786,7 +796,7 @@ func dirMatchesOneOf(dir, home, goos string, regexes []string) bool {
 	return false
 }
 
-func isPathSeparator(env Environment, c uint8) bool {
+func IsPathSeparator(env Environment, c uint8) bool {
 	if c == '/' {
 		return true
 	}
@@ -800,13 +810,13 @@ func isPathSeparator(env Environment, c uint8) bool {
 // Trailing path separators are removed before extracting the last element.
 // If the path consists entirely of separators, Base returns a single separator.
 func Base(env Environment, path string) string {
-	if path == "/" {
-		return path
-	}
 	volumeName := filepath.VolumeName(path)
 	// Strip trailing slashes.
-	for len(path) > 0 && isPathSeparator(env, path[len(path)-1]) {
+	for len(path) > 0 && IsPathSeparator(env, path[len(path)-1]) {
 		path = path[0 : len(path)-1]
+	}
+	if len(path) == 0 {
+		return env.PathSeparator()
 	}
 	if volumeName == path {
 		return path
@@ -815,22 +825,69 @@ func Base(env Environment, path string) string {
 	path = path[len(filepath.VolumeName(path)):]
 	// Find the last element
 	i := len(path) - 1
-	for i >= 0 && !isPathSeparator(env, path[i]) {
+	for i >= 0 && !IsPathSeparator(env, path[i]) {
 		i--
 	}
 	if i >= 0 {
 		path = path[i+1:]
 	}
 	// If empty now, it had only slashes.
-	if path == "" {
+	if len(path) == 0 {
 		return env.PathSeparator()
 	}
 	return path
 }
 
+// ParsePath parses an input path and returns a clean root and a clean path.
+func ParsePath(env Environment, inputPath string) (root, path string) {
+	if len(inputPath) == 0 {
+		return
+	}
+	separator := env.PathSeparator()
+	clean := func(path string) string {
+		matches := regex.FindAllNamedRegexMatch(fmt.Sprintf(`(?P<element>[^\%s]+)`, separator), path)
+		n := len(matches) - 1
+		s := new(strings.Builder)
+		for i, m := range matches {
+			s.WriteString(m["element"])
+			if i != n {
+				s.WriteString(separator)
+			}
+		}
+		return s.String()
+	}
+
+	if env.GOOS() == WINDOWS {
+		inputPath = strings.ReplaceAll(inputPath, "/", `\`)
+		// for a UNC path, extract \\hostname\sharename as the root
+		matches := regex.FindNamedRegexMatch(`^\\\\(?P<hostname>[^\\]+)\\+(?P<sharename>[^\\]+)\\*(?P<path>[\s\S]*)$`, inputPath)
+		if len(matches) > 0 {
+			root = `\\` + matches["hostname"] + `\` + matches["sharename"] + `\`
+			path = clean(matches["path"])
+			return
+		}
+	}
+	s := strings.SplitAfterN(inputPath, separator, 2)
+	root = s[0]
+	if !strings.HasSuffix(root, separator) {
+		// a root should end with a separator
+		root += separator
+	}
+	if len(s) == 2 {
+		path = clean(s[1])
+	}
+	return root, path
+}
+
 func ReplaceHomeDirPrefixWithTilde(env Environment, path string) string {
-	if strings.HasPrefix(path, env.Home()) {
-		return strings.Replace(path, env.Home(), "~", 1)
+	home := env.Home()
+	// match Home directory exactly
+	if !strings.HasPrefix(path, home) {
+		return path
+	}
+	rem := path[len(home):]
+	if len(rem) == 0 || IsPathSeparator(env, rem[0]) {
+		return "~" + rem
 	}
 	return path
 }
--- a/src/segments/path.go
+++ b/src/segments/path.go
@@ -5,8 +5,8 @@ import (
 	"oh-my-posh/environment"
 	"oh-my-posh/properties"
 	"oh-my-posh/regex"
+	"oh-my-posh/shell"
 	"oh-my-posh/template"
-	"path/filepath"
 	"sort"
 	"strings"
 )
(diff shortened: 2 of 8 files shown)


JanDeDobbeleer:
This is going to take a while to go through. Kudos on the bravery though, I can already credit that. This is one of the oldest and most annoying segments.

src/environment/shell.go
```diff
@@ -799,22 +809,22 @@ func isPathSeparator(env Environment, c uint8) bool {
...
-	if path == "/" {
-		return path
-	}
 	volumeName := filepath.VolumeName(path)
 	// Strip trailing slashes.
-	for len(path) > 0 && isPathSeparator(env, path[len(path)-1]) {
+	for len(path) > 0 && IsPathSeparator(env, path[len(path)-1]) {
 		path = path[0 : len(path)-1]
 	}
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
 	if err != nil {
 		pt.env.Log(environment.Error, "getFolderSeparator", err.Error())
 	}
+	if text == "" {
```
JanDeDobbeleer:
```suggestion
	if len(text) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
...
+		return pt.env.PathSeparator()
+	}
 	return text
 }
 
 func (pt *Path) getMixedPath() string {
-	var buffer strings.Builder
 	pwd := pt.getPwd()
-	splitted := strings.Split(pwd, pt.env.PathSeparator())
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
...
-
 	return buffer.String()
 }
 
 func (pt *Path) getAgnosterPath() string {
-	var buffer strings.Builder
 	pwd := pt.getPwd()
-	buffer.WriteString(pt.rootLocation())
-	pathDepth := pt.pathDepth(pwd)
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -149,67 +152,90 @@ func (pt *Path) getFolderSeparator() string {
...
-	case 0:
+	if pwd == "" {
 		return ""
-	case 1:
-		return splitted[0]
-	case 2:
-		return fmt.Sprintf("%s%s%s", splitted[0], separator, splitted[1])
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -229,35 +255,51 @@ func (pt *Path) getRelevantLetter(folder string) string {
 }
 
 func (pt *Path) getLetterPath() string {
-	var buffer strings.Builder
 	pwd := pt.getPwd()
-	splitted := strings.Split(pwd, pt.env.PathSeparator())
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -229,35 +255,51 @@ func (pt *Path) getRelevantLetter(folder string) string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -229,35 +255,51 @@ func (pt *Path) getRelevantLetter(folder string) string {
...
 	}
+	buffer.WriteString(fmt.Sprintf("%s%s", separator, elements[n-1]))
 	return buffer.String()
 }
 
 func (pt *Path) getUniqueLettersPath() string {
-	var buffer strings.Builder
 	pwd := pt.getPwd()
-	splitted := strings.Split(pwd, pt.env.PathSeparator())
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -229,35 +255,51 @@ func (pt *Path) getRelevantLetter(folder string) string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
 	}
+	buffer.WriteString(fmt.Sprintf("%s%s", separator, elements[n-1]))
 	return buffer.String()
 }
 
 func (pt *Path) getAgnosterFullPath() string {
 	pwd := pt.getPwd()
-	for len(pwd) > 1 && string(pwd[0]) == pt.env.PathSeparator() {
-		pwd = pwd[1:]
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
 	}
-	return pt.replaceFolderSeparators(pwd)
+	root = root[:len(root)-1] + pt.getFolderSeparator()
+	return root + path
 }
 
 func (pt *Path) getAgnosterShortPath() string {
 	pwd := pt.getPwd()
-	pathDepth := pt.pathDepth(pwd)
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+			buffer.WriteString(separator)
+		}
 	}
 	return buffer.String()
 }
 
 func (pt *Path) getFullPath() string {
 	pwd := pt.getPwd()
-	return pt.replaceFolderSeparators(pwd)
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+	}
+	path = pt.replaceFolderSeparators(path)
+	return root + path
 }
 
 func (pt *Path) getFolderPath() string {
 	pwd := pt.getPwd()
-	pwd = environment.Base(pt.env, pwd)
-	return pt.replaceFolderSeparators(pwd)
+	if pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pwd) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return ""
+	}
+	root, path := environment.ParsePath(pt.env, pwd)
+	if path == "" {
```
JanDeDobbeleer:
```suggestion
	if len(path) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return pt.formatRoot(root)
+	}
+	return environment.Base(pt.env, path)
+}
+
+func (pt *Path) setPwd() {
+	if pt.pwd != "" {
```
JanDeDobbeleer:
```suggestion
	if len(pt.pwd) > 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+		return
+	}
+	if pt.env.Shell() == shell.PWSH || pt.env.Shell() == shell.PWSH5 {
+		pt.pwd = pt.env.Flags().PSWD
+	}
+	if pt.pwd == "" {
```
JanDeDobbeleer:
```suggestion
	if len(pt.pwd) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
-			// -1 as we want to ignore the trailing slash
-			// set by the normalize function
-			return replacement + pwd[len(key)-1:]
+		keyRoot, keyPath := environment.ParsePath(pt.env, key)
+		if keyRoot != pwdRoot || !strings.HasPrefix(pwdPath, keyPath) {
+			continue
+		}
+		value := mappedLocations[key]
+		rem := cleanPwdPath[len(keyPath):]
+		if rem == "" {
```
JanDeDobbeleer:
```suggestion
		if len(rem) == 0 {
```

src/segments/path.go
```diff
@@ -266,140 +308,207 @@ func (pt *Path) getUniqueLettersPath() string {
...
+			// exactly match the full path
+			return value
+		}
+		if keyPath == "" {
```
JanDeDobbeleer:
```suggestion
		if len(keyPath) == 0 {
```

src/segments/path_test.go
```diff
@@ -40,237 +40,646 @@ func renderTemplate(env *mock.MockedEnvironment, segmentTemplate string, context
...
-		{Expected: "~", HomeIcon: "~", HomePath: "/home/bill/", Pwd: "/home/bill/", PathSeparator: "/"},
-		{Expected: "usr", HomePath: "/home/bill/", Pwd: "/usr/error/what", PathSeparator: "/"},
-		{Expected: "C:", HomePath: "C:\\Users\\Bill", Pwd: "C:\\Program Files\\Go", GOOS: environment.WINDOWS, PathSeparator: "\\"},
-		{Expected: "~", HomeIcon: "~", HomePath: "C:\\Users\\Bill", Pwd: "Microsoft.PowerShell.Core\\FileSystem::C:\\Users\\Bill", GOOS: environment.WINDOWS, PathSeparator: "\\"},
-		{Expected: "C:", HomePath: "C:\\Users\\Jack", Pwd: "Microsoft.PowerShell.Core\\FileSystem::C:\\Users\\Bill", GOOS: environment.WINDOWS, PathSeparator: "\\"},
-		{Expected: "", HomePath: "C:\\Users\\Jack", Pwd: "", GOOS: environment.WINDOWS, PathSeparator: "\\"},
-		{Expected: "DRIVE:", HomePath: "/home/bill/", Pwd: "/usr/error/what", Pswd: "DRIVE:", PathSeparator: "/"},
+		{
+			Case:          "Home folder",
+			Expected:      "",
```
JanDeDobbeleer: Can be omitted

JanDeDobbeleer (changes requested):
This code block is repeated a lot:

```golang
pwd := pt.getPwd()
return pt.replaceFolderSeparators(pwd)
if pwd == "" {
	return ""
}
root, path := environment.ParsePath(pt.env, pwd)
if path == "" {
	return pt.formatRoot(root)
}
```

Can we do this outside of the style logic's?

lewis-yeung:
> Can we do this outside of the style logic's?

@JanDeDobbeleer Yes, we can. Will do it later, maybe when I'm free.

lewis-yeung:
@JanDeDobbeleer Updated. Should we move on?

JanDeDobbeleer:
Let me check, don't gave time right now but I'll get to it ASAP.

src/environment/shell.go
```diff
@@ -800,37 +810,84 @@ func isPathSeparator(env Environment, c uint8) bool {
...
+	if len(path) == 0 {
 		return env.PathSeparator()
 	}
 	return path
 }
 
+// ParsePath parses an input path and returns a clean root and a clean path.
+func ParsePath(env Environment, inputPath string) (root, path string) {
+	if len(inputPath) == 0 {
+		return "", ""
```
JanDeDobbeleer: I think you can just return the variables as well, no?
```suggestion
		return root, path
```

JanDeDobbeleer (commented):
1 minor, you can argue with it :-)

src/environment/shell.go
lewis-yeung: Yes, of course. And even simpler, a naked `return` is all we need here.

PR #2951, opened 2022-10-15
Revisions for async prompt in cmd/clink.

chrisant996:
### Prerequisites

### Description

### DIFF
--- a/src/shell/scripts/omp.lua
+++ b/src/shell/scripts/omp.lua
@@ -23,10 +23,7 @@ end
 local endedit_time = 0
 local last_duration = 0
 local tooltips_enabled = ::TOOLTIPS::
-local rprompt_enabled = true
--- REVIEW:  OMP could set rprompt_enabled = ::HASRPROMPT:: which could speed up
--- performance for configs that don't have an rprompt, by not needing to invoke
--- OMP an extra time for the rprompt.
+local rprompt_enabled = ::RPROMPT::
 
 local cached_prompt = {}
 -- Fields in cached_prompt:
@@ -232,9 +229,6 @@ function p:rightfilter(prompt)
     return (cached_prompt.tooltip or cached_prompt.right), false
 end
 function p:transientfilter(prompt)
--- REVIEW: this can potentially be expensive and noticable for users; consider
--- optimizing this so the Lua script can natively generate the transient prompt
--- when feasible.
     local prompt_exe = string.format('%s print transient --shell=cmd --config=%s %s', omp_exe(), omp_config(), error_level_option())
     prompt = run_posh_command(prompt_exe)
     if prompt == "" then
--- a/src/shell/scripts/omp.lua
+++ b/src/shell/scripts/omp.lua
@@ -1,7 +1,6 @@
 -- Helper functions
 
--- REVIEW:  Why is this a global function?  Should it be local?
-function get_priority_number(name, default)
+local function get_priority_number(name, default)
 	local value = os.getenv(name)
 	if os.envmap ~= nil and type(os.envmap) == 'table' then
 		local t = os.envmap[name]
--- a/src/shell/scripts/omp.lua
+++ b/src/shell/scripts/omp.lua
@@ -46,6 +46,7 @@ end
 local function run_posh_command(command)
     command = '"'..command..'"'
     local _,ismain = coroutine.running()
+    local output
     if ismain then
         output = io.popen(command):read("*a")
     else
--- a/src/shell/scripts/omp.lua
+++ b/src/shell/scripts/omp.lua
@@ -1,5 +1,6 @@
 -- Helper functions
 
+-- REVIEW:  Why is this a global function?  Should it be local?
 function get_priority_number(name, default)
 	local value = os.getenv(name)
 	if os.envmap ~= nil and type(os.envmap) == 'table' then
@@ -18,14 +19,42 @@ function get_priority_number(name, default)
 	end
 end
 
--- Duration functions
+-- Local state
 
 local endedit_time = 0
 local last_duration = 0
-local tip
 local tooltips_enabled = ::TOOLTIPS::
-local tooltip_active = false
+local rprompt_enabled = true
+-- REVIEW:  OMP could set rprompt_enabled = ::HASRPROMPT:: which could speed up
+-- performance for configs that don't have an rprompt, by not needing to invoke
+-- OMP an extra time for the rprompt.
+
 local cached_prompt = {}
+-- Fields in cached_prompt:
+--      .cwd            = Current working directory of prompt.
+--      .left           = Left side prompt.
+--      .right          = Right side prompt.
+--      .tooltip        = Tooltip prompt.
+--      .tip_command    = Command for which to produce a tooltip.
+--      .coroutine      = Coroutine for the tooltip prompt.
+
+local function cache_onbeginedit()
+    local cwd = os.getcwd()
+    local old_cache = cached_prompt
+
+    -- Start a new table for the new edit/prompt session.
+    cached_prompt = { cwd=cwd }
+
+    -- Copy the cached left/right prompt strings if the cwd hasn't changed.
+    -- IMPORTANT OPTIMIZATION:  This keeps the prompt highly responsive, except
+    -- when changing the current working directory.
+    if old_cache.cwd == cwd then
+        cached_prompt.left = old_cache.left
+        cached_prompt.right = old_cache.right
+    end
+end
+
+-- Configuration
 
 local function omp_exe()
     return '"'..::OMP::..'"'
@@ -36,6 +65,8 @@ local function omp_config()
 end
 
 os.setenv("POSH_THEME", ::CONFIG::)
+
+-- Execution helpers
 
 local function can_async()
     if (clink.version_encoded or 0) >= 10030001 then
@@ -54,6 +85,8 @@ local function run_posh_command(command)
     end
     return output
 end
+
+-- Duration functions
 
 local function os_clock_millis()
     -- Clink v1.2.30 has a fix for Lua's os.clock() implementation failing after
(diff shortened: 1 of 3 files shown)


chrisant996:
@JanDeDobbeleer I didn't know at first about the special syntax for topic branch commit messages:
- Should I rebase and force push?
- Or abandon the PR, create a new branch with new commits, and create a new PR?

I don't have Go and am not currently able to build OMP.  I did all my testing in a `oh-my-posh.lua` script, and then applied the necessary deltas back to the `omp.lua` script.

So, I would appreciate if you can sanity check the changes.

Also, there are some `REVIEW:` comments for discussion.

JanDeDobbeleer:
@chrisant996 you can rebase for sure.

chrisant996:
@JanDeDobbeleer Rebased and force pushed.

I don't have Go and am not currently able to build OMP. I did all my testing in a oh-my-posh.lua script, and then applied the necessary deltas back to the omp.lua script. So, I would appreciate if you can sanity check the changes.

Also, there are some `REVIEW:` comments for discussion.

JanDeDobbeleer:
@chrisant996 you'll need to not start with a capital:

- fix(cmd): adjust leaked global variable
- feat(cmd): revisions for async prompt in cmd/clink

Same goes for the paragraphs in the second commit. ~~I'll go through this tomorrow as now I'm rather tired :-)~~ Well, I did it anyways...

src/shell/scripts/omp.lua
```diff
@@ -1,5 +1,6 @@
 -- Helper functions
 
+-- REVIEW:  Why is this a global function?  Should it be local?
```
JanDeDobbeleer: it can definitely be local

src/shell/scripts/omp.lua
```diff
@@ -18,14 +19,42 @@ function get_priority_number(name, default)
...
--- Duration functions
+-- Local state
 
 local endedit_time = 0
 local last_duration = 0
-local tip
 local tooltips_enabled = ::TOOLTIPS::
-local tooltip_active = false
+local rprompt_enabled = true
+-- REVIEW:  OMP could set rprompt_enabled = ::HASRPROMPT:: which could speed up
```
JanDeDobbeleer: Agreed, we do this for other shells as well, I can add that.

src/shell/scripts/omp.lua
```diff
@@ -134,50 +182,60 @@ end
...
+    return cached_prompt.left
+end
+function p:rightfilter(prompt)
+    -- Return cached tooltip if available, otherwise return cached rprompt.
+    -- Returning false as the second return value halts further prompt
+    -- filtering, to keep other things from overriding what we generated.
+    return (cached_prompt.tooltip or cached_prompt.right), false
 end
 function p:transientfilter(prompt)
+-- REVIEW: this can potentially be expensive and noticable for users; consider
```
JanDeDobbeleer: what do you mean, cache the value once and reuse? This shouldn't be a very expensive prompt as the possibilities are limited to cached values already.

chrisant996: The `transientfilter()` is called after the user presses <kbd>Enter</kbd> and it is not using a cached value.
It runs a new invocation of the OMP executable every single time.

Transient filters have the option to return whatever they want.  It's common to return something like simply `"> "`, but I don't know what range of possibilities OMP allows for there.

If the intent is for a constant string to be used, it would be great to use some kind of `::TRANSIENTPROMPT::` placeholder, to avoid invoking OMP.

chrisant996:
P.S.  I don't understand what to do about the "This branch is out-of-date with the base branch" indicator.  `git push`, `git pull`, and `git merge main` all say everything is up to date.  🤔

src/shell/scripts/omp.lua
chrisant996: Thanks -- I updated the PR.

src/shell/scripts/omp.lua
JanDeDobbeleer: Ah yes, the thing is, it can differ between runs. Instead of fetching live data, the transient prompt works with cached data from the previous run and resolves 1 template so it should be rather fast.

JanDeDobbeleer:
@chrisant996 don't mind that. Your branch is out of date with our main. To resolve that you'll need to add our repo as a remote, then rebase your branch on top of ours:

- git remote add upstream git@xxx (can't see the correct URL now)
- git fetch upstream
- git rebase upstream/main
- git push --force

chrisant996:
Vscode forked the repo for me while making the pull request. I think vscode did some things wrong. I've had to fix a couple of things already (I couldn't force push at first, because vscode's GitHub extension didn't update the tracking info for the branch to be in the fork -- the branch thought it was in your repo still).

src/shell/scripts/omp.lua
chrisant996: Ok, so it should be relatively fast, but will be noticable when holding down Enter.

Do you mean the omp exe is caching between runs? That's what I was trying to suggest as the optimal approach for async prompt, and I thought you had said that was architecturally not feasible.

chrisant996:
> @chrisant996 don't mind that. Your branch is out of date with our main. To resolve that you'll need to add our repo as a remote, then rebase your branch on top of ours:
> 
> * git remote add upstream git@xxx (can't see the correct URL now)
> * git fetch upstream
> * git rebase upstream/main
> * git push --force

Done.  Upstream was already your repo.

Here is what had happened:

I was originally in your repo, in main.  I used git pull on 10/15, and I had no local changes.  I made edits.  Then I used git switch -c to create a branch.  I tried to publish via vscode.  But oops, I forgot I didn't have a fork yet.  But wow, vscode & github are so helpful they offered to magically create a fork in-place and update everything and then push and then even create a pull request.  It seemed great!

But they left the branch in the fork still with its origin pointing at your repo instead of the fork.
And they somehow didn't fork the latest state of the repo, even though I was initially fully up to date.

I could have just concluded I did something wrong ... and maybe I did! ... but creating a fork where a topic branch points to upstream instead of origin seems like a bug in the auto-fork procedure.

chrisant996:
Just double checking:  is this blocked waiting on me?  If so then I'm not sure what step I need to do.

But as long as it's not waiting on me for something, then I'm happy and have no expectation on when/whether the PR gets applied.  😁

JanDeDobbeleer:
It's waiting on me. Will have a proper look during lunch today!

chrisant996:
> It's waiting on me. 

Oh, good.

> Will have a proper look during lunch today!

No worries, take your time.  I only wanted to make sure it wasn't waiting on me.

src/shell/scripts/omp.lua
JanDeDobbeleer: @chrisant996 no it does not cache the prompt, it does cache all the results of every segment in the "normal", previous prompt run so it actually only renders a template.

src/shell/scripts/omp.lua
JanDeDobbeleer: Done

src/shell/scripts/omp.lua
```diff
@@ -207,17 +258,47 @@ end
 
 -- Tooltips
 
+local function get_tip_command(line)
+    if USE_ENTIRE_COMMAND_LINE then
+-- REVIEW:  This is what oh-my-posh was doing -- was that intentional?
```
JanDeDobbeleer, line 263: @chrisant996 I would say it makes sense to only used the second one as we only need the first keyword. So yes, we can remove this in theory.

JanDeDobbeleer:
@chrisant996 I can't validate this as I can't use clink on arm. That's why this is still open.

chrisant996:
> @chrisant996 I can't validate this as I can't use clink on arm. That's why this is still open.

@JanDeDobbeleer I don't understand. Clink can't be used natively on arm yet; only via x86/x64 emulation on arm.

What is needed from me?

JanDeDobbeleer:
@chrisant996 did you validate the lua changes? Because I validated my changes.

chrisant996:
> @chrisant996 did you validate the lua changes? Because I validated my changes.

@JanDeDobbeleer I validated all of my Lua changes prior to creating the pull request. I haven't validated the results of merges during the PR.

Do you want me to verify the resulting merged stuff?

Also, I have Go installed now. Is building omp for Windows straightforward? If so, then I can likely do E2E validation now.

(Sidebar, for context: I installed Go to try to contribute to the fzf project, but there are no instructions for how to build it on Windows, and there seem to be comments in issues suggesting that it takes special undocumented steps.)

JanDeDobbeleer:
@chrisant996 using PowerShell, you can easily build and replace your current oh-my-posh executable using `go build -o (Get-Command oh-my-posh.exe).Source` . Would be ace if you could have a look. There's also one pending review comment.

src/shell/scripts/omp.lua
JanDeDobbeleer: @chrisant996 👀  the first one can be removed if the second one works.

chrisant996:
@JanDeDobbeleer Sorry, a family situation came up and this fell by the wayside for a bit.

I tried `go build -o (Get-Command oh-my-posh.exe).Source` from an up-to-date repo, and it reports:

```
go: cannot find main module, but found .git/config in C:\repos\oh-my-posh
        to create a module there, run:
        go mod init
```

JanDeDobbeleer:
@chrisant996 you need to be in the `src` folder 😄

chrisant996:
@JanDeDobbeleer DOH!  Success.  😎

Seems to be working well.

Here are the timings reported by Clink (<kbd>Alt-1</kbd><kbd>Ctrl-X</kbd><kbd>Ctrl-Z</kbd>):

```
prompt filters:
  filter:                                         last    avg     peak
        [string "-- Helper functions..."]:180     0 ms   18 ms  179 ms
  rightfilter:                                    last    avg     peak
        [string "-- Helper functions..."]:225     0 ms    0 ms    0 ms
  transientfilter:                                last    avg     peak
        [string "-- Helper functions..."]:231    62 ms   62 ms   62 ms
  transientrightfilter:                           last    avg     peak
        [string "-- Helper functions..."]:239     0 ms    0 ms    0 ms
```

The "18 ms" average keeps getting lower and lower with each press of <kbd>Enter</kbd>, which is consistent with the intended behavior, which is that the first prompt in a directory will invoke git immediately and take some time, but subsequent prompts will take 0 ms to invoke git in the background (git in the background takes time, but it doesn't experience responsiveness for the user).

The 62ms for the transient filter isn't ideal, but I agree it's fast enough that it's not worth addressing further unless some users report it as a problem and are able to describe the scenario and why it's problematic.

The part about `[string "-- Helper functions..."]` is because the Lua script is loaded as a string, rather than as a file, so there is no filename associated with it.  So it uses the beginning of the string as a proxy for a filename.  If the first line of the omp.lua script were changed to something like `-- oh-my-posh script` or something like that, then it would be more descriptive.

PR #2988, opened 2022-10-20
fix(scoop): innosetup package installation

vcheckzen:
scoop will extract innosetup package directly when enabling `"innosetup": true,`, conflicting with the `installer` config

### Prerequisites

### Description

### DIFF
--- a/packages/scoop/oh-my-posh.json
+++ b/packages/scoop/oh-my-posh.json
@@ -16,12 +16,6 @@
       "hash": "<HASH-386>"
     }
   },
-  "installer": {
-    "args": [
-      "/CURRENTUSER",
-      "/VERYSILENT"
-    ]
-  },
   "innosetup": true,
   "checkver": {
     "github": "https://github.com/JanDeDobbeleer/oh-my-posh"


JanDeDobbeleer:
@vcheckzen I removed that parameter in #2987, it should in fact install the same way as the other installation means. And not create a  shim at another location. That way, it doesn't matter how you proceed, everything will be updated in the same location. That said, the scoop docs could use proper guidance in all of this. Trial by error here.

vcheckzen:
Using the installer's way will make oh-my-posh escape from scoop's directory, somehow going against scoop's principle saying on its doc.

>Scoop downloads and manages packages in a portable way, keeping them neatly isolated in ~\scoop. It won't install files outside its home, and you can place a Scoop installation wherever you like.

Some packages using innosetup can be found in scoop's repository.

https://github.com/ScoopInstaller/Main/blob/master/bucket/openssl.json
https://github.com/ScoopInstaller/Main/blob/master/bucket/lua-for-windows.json

JanDeDobbeleer:
@vcheckzen I understand, but then this PR should also create the shims, no? It would also need a bin parameter I suppose.

JanDeDobbeleer:
@vcheckzen would you mind updating the definition in the main bucket to use the inno installer and stick to scoop's principles? In that way we have my way in the docs and the official way in the main bucket. Remember, everything I offer here, I need to support.

vcheckzen:
Sorry, I'm not a main bucket maintainer. I think their existing definition will continue to work if you don't change the release files structure.

JanDeDobbeleer:
@vcheckzen surely you can create a pull request? Otherwise I have to it as they do not use the signed binaries.

vcheckzen:
I'm not confident of my knownledge of their manner, sorry to bother you for that work.

PR #3005, opened 2022-10-23
feat(dotnettarget): add dotnettarget segment

### LINKED ISSUE #1568: Enhance dotnet segment to support .Net framework
### Code of Conduct

Enhance the dotnet segment to recognize if the folder contains a .Net framework project vs a dotnet core project.

Some suggestions are on the discussion such as parsing the project files and looking at the TargetFrameworkVersion element. 

I don't see an easy way to do this if only the .sln file is in the folder however, perhaps drilling into one of the projects in the solution and looking at that.

babieiev:
### Prerequisites

### Description

This PR adds a new segment to display the target .NET framework version moniker (TFM) of the current project.

Resolves #1568

### DIFF
--- a/src/engine/segment.go
+++ b/src/engine/segment.go
@@ -106,6 +106,8 @@ const (
 	DENO SegmentType = "deno"
 	// DOTNET writes which dotnet version is currently active
 	DOTNET SegmentType = "dotnet"
+	// DOTNETTARGET writes which version of .NET the current project is targeting
+	DOTNETTARGET SegmentType = "dotnettarget"
 	// EXECUTIONTIME writes the execution time of the last run command
 	EXECUTIONTIME SegmentType = "executiontime"
 	// EXIT writes the last exit code
@@ -283,6 +285,7 @@ func (segment *Segment) mapSegmentWithWriter(env environment.Environment) error 
 		DART:          &segments.Dart{},
 		DENO:          &segments.Deno{},
 		DOTNET:        &segments.Dotnet{},
+		DOTNETTARGET:  &segments.DotnetTarget{},
 		EXECUTIONTIME: &segments.Executiontime{},
 		EXIT:          &segments.Exit{},
 		FLUTTER:       &segments.Flutter{},
--- a/src/segments/dotnet_target.go
+++ b/src/segments/dotnet_target.go
@@ -2,12 +2,10 @@ package segments
 
 import (
 	"errors"
-	"fmt"
 	"oh-my-posh/environment"
 	"oh-my-posh/properties"
 	"oh-my-posh/regex"
 	"strings"
-	"time"
 )
 
 type DotnetTarget struct {
@@ -53,7 +51,6 @@ func (d *DotnetTarget) Enabled() bool {
 }
 
 func (d *DotnetTarget) getProjectFiles() []string {
-	defer d.env.Trace(time.Now(), "getProjectFiles")
 	var files []string
 	for _, extension := range d.extensions {
 		cwd := d.env.Pwd()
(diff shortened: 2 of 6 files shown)


JanDeDobbeleer:
@lesfiddler I can sure help. I'll try to go through it tomorrow. There's a few PR'a I need to help with.

babieiev:
@JanDeDobbeleer The problem is solved. I added tests and updated the pull request.

JanDeDobbeleer:
@lesfiddler thanks, now what I would do (and this is a refactor on your end I suppose), is also extend the `language` struct [like the dotnet segment](https://github.com/JanDeDobbeleer/oh-my-posh/blob/fed2d020256f19bf54c5094afc5a680bbe32a2a9/src/segments/dotnet.go#L10). That way you get a lof of things out of the box, and you can extend it with a `getVersion` function to extract the `TFM`, just like you already do. The python segment has an [example](https://github.com/JanDeDobbeleer/oh-my-posh/blob/fed2d020256f19bf54c5094afc5a680bbe32a2a9/src/segments/python.go#L38) for this. I'm still wondering if we should extend the `dotnet` segment with this one (first in the array), rather than set up a new segment. If a TFM is found, take that one, if not proceed with the dotnet CLI.

babieiev:
@JanDeDobbeleer I'm also hesitant to add a new segment. I suggested this because it seems to me that this approach is just easier to understand. There is a certain difference between this and all other `language` segments. `TFM` is not the same as version. Let me explain. The moniker can look like `net7.0`, `netcoreapp3.1`, `net6.0-ios15.0`, and in some cases (although quite rarely), like `uap10.0` for example. If the `dotnet` segment shows the version of the active SDK, then my segment shows which framework the current project is targeting.
Of course, we can extend `language` struct, but I'm not sure that combining these two segments (`dotnet` and `dotnet_target`) is a good idea. But of course you are the maintainer, so the final decision is yours.
I'm a little busy right now, but soon I'll try to make the changes you ask for.

JanDeDobbeleer:
@lesfiddler you're right, this is the same as [project](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segments/project.go) however, not dotnet. We should be able to extend that one.

babieiev:
Oh! It's a wonder I didn't notice it sooner. [Project](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segments/project.go) segment is exactly what is needed, thanks.

babieiev:
@JanDeDobbeleer `dotnet` project files can have 3 extensions: `*.csproj`, `*.vbproj` and `*.fsproj`. But `project.go` is not designed for multiple files. I am not sure that completely changing the structure, and accordingly the work, is a good decision. So, if we want to keep it as part of the `project.go`, I suggest separating them with a semicolon, like here:

New (multiple files):
```go
...
{
	Name:    "dotnet",
	File:    "*.csproj;*.vbproj;*.fsproj",
	Fetcher: n.getDotnetProject,
},
...

func (n *Project) hasProjectFile(p *ProjectItem) bool {
	files := strings.Split(p.File, ";")
	for _, file := range files {
		if n.env.HasFiles(file) {
			return true
		}
	}
	return false;
}
```

Old (single file):
```go
...
{
	Name:    "dotnet",
	File:    "*.csproj",
	Fetcher: n.getDotnetProject,
},
...

func (n *Project) hasProjectFile(p *ProjectItem) bool {
	return n.env.HasFiles(p.File)
}
```
Is that ok, or would you suggest something better?

JanDeDobbeleer:
@lesfiddler I would change File to an array. Cleaner, and only internal anyways.

babieiev:
@JanDeDobbeleer I've created a new branch for this change, so I'll open [a new PR](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3019) and close this one.

PR #3012, opened 2022-10-24
theme: add catppuccin

afmosquerare:
Sorry


JanDeDobbeleer:
👋 @anphemus, theme aditions are no longer accepted due to the ever growing set. We do however accept showcasing your custom theme in the [🎨 Themes section](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/categories/themes) in discussions.

JanDeDobbeleer:
@anphemus we already have catppucin themes in main, is there a specific reason to include this one as well?

PR #3019, opened 2022-10-26
feat(project): enhance project segment with .NET

### LINKED ISSUE #1568: Enhance dotnet segment to support .Net framework

babieiev:
### Prerequisites

### Description

Enhance `project.go` by fetching current .NET project version.

Resolves #1568 

### DIFF
--- a/src/segments/project.go
+++ b/src/segments/project.go
@@ -3,22 +3,33 @@ package segments
 import (
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"oh-my-posh/environment"
 	"oh-my-posh/properties"
+	"oh-my-posh/regex"
 	"path/filepath"
+	"strings"
 
 	"github.com/BurntSushi/toml"
 )
 
 type ProjectItem struct {
 	Name    string
-	File    string
-	Fetcher func(item ProjectItem) (string, string)
+	Files   []string
+	Fetcher func(item ProjectItem) *ProjectData
 }
 
 type ProjectData struct {
 	Version string
 	Name    string
+	Target  string
+}
+
+func (p *ProjectData) enabled() bool {
+	if p == nil {
+		return false
+	}
+	return len(p.Version) > 0 || len(p.Name) > 0 || len(p.Target) > 0
 }
 
 // Rust Cargo package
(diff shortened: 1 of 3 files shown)


src/segments/project.go
```diff
@@ -163,3 +176,29 @@ func (n *Project) getNuSpecPackage(item ProjectItem) (string, string) {
...
+		extension := filepath.Ext(file.Name())
+		if extension == ".csproj" || extension == ".fsproj" || extension == ".vbproj" {
+			name = strings.TrimSuffix(file.Name(), filepath.Ext(file.Name()))
+			content = n.env.FileContent(file.Name())
+			break
+		}
+	}
+	// the name of the parameter may differ depending on the version,
+	// so instead of xml.Unmarshal() we use regex:
+	tag := "(?P<tag><.*TargetFramework.*>(?P<tfm>.*)</.*TargetFramework.*>)"
```
JanDeDobbeleer: Can we make `tag` and `tfm` uppercase? Let's not forget to also change below in that case.
```suggestion
	tag := "(?P<TAG><.*TargetFramework.*>(?P<TFM>.*)</.*TargetFramework.*>)"
```

JanDeDobbeleer (changes requested):

babieiev:
@JanDeDobbeleer it's strange that the `ubuntu` test fails because the lines are repeated three times. I didn't know that in `golang` everything is so strict with this 😅. Should I create local constants or do something else?

JanDeDobbeleer:
@lesfiddler I'll check. One question I also have is that the behavior of this check still different compared to the others in project as they parse the project version number and not a target framework. Maybe we can extend the segment to have both values as both go and node definitely offer that. I bet there's also a `Version` property in the `Solution` file.

babieiev:
> I bet there's also a `Version` property in the `Solution` file.

@JanDeDobbeleer unfortunately, I do not think that it is possible. Regarding the `.sln` file, @PilotBob already wrote in issue #1568:

> There is no distinction in a .sln file unfortunately.

Also, there is no information about this in the [official documentation](https://learn.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file). Moreover, modern `dotnet` allows you to have projects without a `.sln` file. It is mostly needed to maintain the hierarchy of projects. The only version it contains is the Visual Studio version.

babieiev:
> One question I also have is that the behavior of this check still different compared to the others in project as they parse the project version number and not a target framework.

I know that the behavior is not similar to any segment (that's why I suggested making a separate segment in [my previous PR](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3005)). However, the project, in comparison with others, is the most suitable for this

JanDeDobbeleer:
@lesfiddler alright, so I had a good thought on this and we'll need an additional Property called `.Target` here where we add this version number. I can extend the other languages like Node and go who have this as well.

babieiev:
@JanDeDobbeleer if I understood it correctly, I should add a new property, but.. where exactly? `project.go` or `language.go`?

JanDeDobbeleer:
@lesfiddler I'll do it. Don't worry, the PR is perfect, I can take over.

JanDeDobbeleer:
@lesfiddler 👀

babieiev:
@JanDeDobbeleer looks good. Do I need to do something with it?

JanDeDobbeleer:
@lesfiddler nope, merging right away!

PR #3100, opened 2022-11-18
Add support for the Redhat icon

GerbenWelter:
Because 'redhat' wasn't recognized as the OS name, the icon always defaulted to the generic tux icon. This commit adds support for displaying the 'shadow man' icon where the OS name is 'redhat'.

### DIFF
--- a/src/segments/os.go
+++ b/src/segments/os.go
@@ -51,6 +51,8 @@ const (
 	Nixos properties.Property = "nixos"
 	// Opensuse the string/icon to use for Opensuse
 	Opensuse properties.Property = "opensuse"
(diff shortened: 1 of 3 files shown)


lewis-yeung:
@GerbenWelter You should also add an entry in [schema.json](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/schema.json). See [here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/877ac59f5f6688db4890160c3121a9e79dae4bcc/themes/schema.json#L1449-L1454) for an example. :)

GerbenWelter:
@lewis-yeung  Thx for the hint. Added and pushed.

PR #3219, opened 2022-12-17
Enable use of a prompt engine within a closed-loop program/shell

maxlandon:
This little PR aims, as the title implies, to enable the use of oh-my-posh prompt engine within a closed-loop shell. All changes are described in the unique commit of this PR. Thanks for considering it


maxlandon:
Note that the various changes to go.mod/go.sum are due to some automatic feature in my editor, and not caused by me adding some package in the PR: the changes in this file can thus probably be ignored safely.

JanDeDobbeleer:
@maxlandon let me get this right, you want to use oh-my-posh as a library, right? If so, we should maybe also use `NewEngine` in our own code.

maxlandon:
Hello !

Fastest answer ever !
Yes that is right, this is what I want.
To get some context: https://github.com/reeflective/readline
The way I use the prompt functions: https://github.com/reeflective/readline/blob/master/prompt.go

Available any time if you have questions or need help !
Thanks a lot !

JanDeDobbeleer:
> Note that the various changes to go.mod/go.sum are due to some automatic feature in my editor

@maxlandon I'm going to revert those though, these shouldn't have been committed.

maxlandon:
Yes sorry, I squashed my changes yesterday and completely forgot to do it !

src/engine/engine.go
```diff
@@ -321,7 +332,7 @@ func (e *Engine) PrintTooltip(tip string) string {
 		Segments:  []*Segment{tooltip},
 	}
 	switch e.Env.Shell() {
-	case shell.ZSH, shell.CMD, shell.FISH:
+	case shell.ZSH, shell.CMD, shell.FISH, shell.PLAIN:
```
JanDeDobbeleer, line 335: why was this added? It's not really used anywhere as it's not a valid shell name.

src/engine/config.go
```diff
@@ -218,7 +219,7 @@ func (cfg *Config) Write(format string) {
 	if len(destination) == 0 {
 		destination = cfg.origin
 	}
-	f, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
+	f, err := os.OpenFile(destination, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o755)
```
JanDeDobbeleer, line 222: this was a mistake, right?

src/engine/engine.go
```diff
@@ -410,8 +421,10 @@ func (e *Engine) PrintExtraPrompt(promptType ExtraPromptType) string {
 			return prompt
 		}
 		return str
-	case shell.PWSH, shell.PWSH5, shell.CMD, shell.BASH, shell.FISH, shell.NU:
+	case shell.PWSH, shell.PWSH5, shell.CMD, shell.BASH, shell.FISH, shell.NU, shell.PLAIN:
```
JanDeDobbeleer, line 424: Same question as above

src/engine/new.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// NewEngine returns a prompt engine initialized with the
+// given configuration options, and is ready to print any
+// of the prompt components.
+func NewEngine(config string) *Engine {
+	env := &platform.Shell{
+		CmdFlags: &platform.Flags{
+			Config: config,
+			Shell:  "shell",
```
JanDeDobbeleer, line 17: this needs to be exposed imho

JanDeDobbeleer:
@maxlandon you can find it here #3223, I continued your work, but now we also use that API internally. If that works, I'll merge it and credit you.

src/engine/config.go
maxlandon: This is due to my editor applying to some linters.

src/engine/engine.go
maxlandon: So this because if the `Shell()` call returns an empty string, the case statement is skipped and thus an empty prompt string is returned as well.

src/engine/engine.go
maxlandon: Same response as above :)
However if there is something I missed in my approach, which results in this being useless, of course feel free to ignore it !

src/engine/new.go
maxlandon: Usually package writers expose functionality through some `pkg` package directory.
I have no particular suggestion as to where this should be exposed, you probably are better placed to decide.

src/engine/engine.go
JanDeDobbeleer: That could make sense, I'll add a sane default in that case.

PR #3223, opened 2022-12-18
refactor(engine): new instantiation function

JanDeDobbeleer:
this commit adds a function to instantiate a new prompt engine to be used within a Go application (typically a readline shell)

### Prerequisites

### Description

Continuation of #3219 


maxlandon:
Give me a few minutes to test this commit, coming back with feeback soon !
Thanks a lot for your quick and kind treatment !

JanDeDobbeleer:
@maxlandon in this scenario a `NewWithConfig(flags *platform.Flags, cfg *engine.Config)` would also be a good idea.

maxlandon:
Oops sorry not yet tested it, was fixing some stuff on my readline code...
Will try to do this tonight !

maxlandon:
Not being _highly_ familiar with the code:
In such a case (with the signature you proposed), is there a function to initalize a default `platform.Environment` ? I could not find one.

Edit: Maybe I'm asking the wrong question, or not really grasping the interest of the function signature you proposed, so feel free to tell me if I'm framing this wrong.

maxlandon:
(Thinking out lout):
- `*platform.Flags` : So this is easily instantiated and set up by a consumer of the library; most of the fields can be left empty without hampering on the prompt's functions. So that seems to be good, a consumer can just ` var flags = &platform.Flags{}` and pass it to `NewEngine(flags)`.
- `*platform.Environment`: Still wandering about this one.

JanDeDobbeleer:
@maxlandon it's not the environment but the `Config`, I'd argue someone would like to simply add a `Config` struct instead of the location to a json/yaml/toml config.

maxlandon:
Oh yes sorry my mind was elsewhere, it is a config indeed.
Additionally, you are probably right: someone could just write a configuration file in a `$TEMP` file/dir and load it from here. 

However, the 'problem' seems to be exemplified here: https://github.com/JanDeDobbeleer/oh-my-posh/blob/dda12ff877de1946ad1def1c5955ba3c1f67c871/src/cli/print.go#L66

Please confirm that the following is right:
```
var flags = &platform.Flags{Config: "path/to/tmp/config.json"}

var prompt = engine.NewWithConfig(flags, nil)
```
In the following case, I assume that the engine would automatically try to load the configuration from the config path in the flags struct ? Or is there a need for a separate function ?

To be clear: I clearly see why you would want to enable a consumer to pass a `*Config` struct, I'm sure some people would be glad to be able to populate it with some segments and stuff.
I'm only thinking that most developers using the prompt as a library might be satisfied by just loading the config from somewhere.

JanDeDobbeleer:
@maxlandon I would still keep `New()`, just thinking about options here.

JanDeDobbeleer:
@all-contributors please add @maxlandon for code

maxlandon:
Hello !
Sorry I've not had time yet to test the changes and propose/settle on an exported signature !
I'm drowning in my own code at the moment, but I'm not forgetting this, as soon as I'm done cleaning my mess I'll do this !
Anyway, thanks a lot for your prompt support and help, and for the mention !
This will be very helpful !

PR #3251, opened 2022-12-26
Enable adding prompt SegmentWriters at runtime

maxlandon:
Hello Jan !
First off: #3223 works like-a-charm. Really, really great.

So I'm coming back with another, even simpler demand ! Complementing the previous one.
This commit adds a an exported function to add SegmentWriters to the engine at runtime.

To be honest, before trying to implement this I thought I was in for troubles...
But the changes to make where simple beyond imagination.

This commit would allow anyone to write prompt segments from any type provided they
implement the `SegmentWriter` interface. Once again, within the context of a readline
program, this is just fire.

Thanks for any consideration once again !
Available at any time for questions/fixes/anything.

### DIFF
--- a/src/engine/segment.go
+++ b/src/engine/segment.go
@@ -211,6 +211,76 @@ const (
 	YTM SegmentType = "ytm"
 )
 
+// PromptSegments contains all available prompt segment writers.
+// Consumers of the library can also add their own segment writer.
+var PromptSegments = map[SegmentType]SegmentWriter{
+	ANGULAR:       &segments.Angular{},
+	AWS:           &segments.Aws{},
+	AZ:            &segments.Az{},
+	AZFUNC:        &segments.AzFunc{},
+	BATTERY:       &segments.Battery{},
+	BREWFATHER:    &segments.Brewfather{},
+	CDS:           &segments.Cds{},
+	CF:            &segments.Cf{},
+	CFTARGET:      &segments.CfTarget{},
+	CMD:           &segments.Cmd{},
+	CONNECTION:    &segments.Connection{},
+	CRYSTAL:       &segments.Crystal{},
+	CMAKE:         &segments.Cmake{},
+	DART:          &segments.Dart{},
+	DENO:          &segments.Deno{},
+	DOTNET:        &segments.Dotnet{},
+	EXECUTIONTIME: &segments.Executiontime{},
+	EXIT:          &segments.Exit{},
+	FLUTTER:       &segments.Flutter{},
+	FOSSIL:        &segments.Fossil{},
+	GCP:           &segments.Gcp{},
+	GIT:           &segments.Git{},
+	GITVERSION:    &segments.GitVersion{},
+	GOLANG:        &segments.Golang{},
+	HASKELL:       &segments.Haskell{},
+	IPIFY:         &segments.IPify{},
+	ITERM:         &segments.ITerm{},
+	JAVA:          &segments.Java{},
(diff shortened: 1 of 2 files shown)


src/engine/segment.go
```diff
@@ -431,3 +443,86 @@ func (segment *Segment) SetText() {
 		segment.text = strings.NewReplacer("`", "\\`", `%`, `%%`).Replace(segment.text)
 	}
 }
+
+// userSegments is a list of segment writers that have been added at runtime.
+var userSegments = make(map[SegmentType]SegmentWriter)
```
JanDeDobbeleer: I would only initialise this when we are about to use it, so validate this in `AddSegment` and initialise if needed.

src/engine/segment.go
```diff
@@ -275,73 +275,76 @@ func (segment *Segment) background() string {
...
-		TEXT:          &segments.Text{},
-		TIME:          &segments.Time{},
-		UI5TOOLING:    &segments.UI5Tooling{},
-		WAKATIME:      &segments.Wakatime{},
-		WINREG:        &segments.WindowsRegistry{},
-		WITHINGS:      &segments.Withings{},
-		XMAKE:         &segments.XMake{},
-		YTM:           &segments.Ytm{},
-	}
+	// functions := map[SegmentType]SegmentWriter{
```
JanDeDobbeleer: this woud need to be deleted

src/engine/segment.go
```diff
@@ -431,3 +443,86 @@ func (segment *Segment) SetText() {
 		segment.text = strings.NewReplacer("`", "\\`", `%`, `%%`).Replace(segment.text)
 	}
 }
+
+// userSegments is a list of segment writers that have been added at runtime.
+var userSegments = make(map[SegmentType]SegmentWriter)
+
+func (segment *Segment) getSegmentWriters() map[SegmentType]SegmentWriter {
+	// Builtins.
+	functions := map[SegmentType]SegmentWriter{
```
JanDeDobbeleer: I would expose this map and add here rather than anything else.

src/engine/segment.go
```diff
@@ -211,6 +211,76 @@ const (
 	YTM SegmentType = "ytm"
 )
 
+// PromptSegments contains all available prompt segment writers.
+// Consumers of the library can also add their own segment writer.
+var PromptSegments = map[SegmentType]SegmentWriter{
```
JanDeDobbeleer: Can we rename this to `Segments`

maxlandon:
Also, quite unrelated. I was wandering: So the oh-my-posh module is just declared as that: `oh-my-posh`. 
Since you distribute the prompt as a binary in every release, this is not so much of a problem.

So my question, and I'm sorry by advance if this question/demand would seem astoundingly stupid: would it put a giant mess to oh-my-posh infrastructure if the its go.mod would declare itself as a fully qualified name ? 

I definitely understand that since oh-my-posh has until now been used only as a standalone binary, there was no need for such a thing. However, and in light of the recent related PRs, I'm wandering how library consumers would be able to grab the oh-my-posh code.

Of course remains the option of a git submodule... but if this hell could be avoided ?
If the answer is "yes it might be possible but the answer must be searched and thought of thoroughly", I'm obviously available to assist in finding it, or a solution.

Finally, sorry for bothering that much. You've already done more than what I wished for...

JanDeDobbeleer:
@maxlandon I think it can be changed , let me try that.

JanDeDobbeleer:
merged in #3263

PR #3258, opened 2022-12-27
feat: conditional segment styles

NoF0rte:
### Prerequisites

### Description
This pull request adds the ability to conditionally style segments. This comes in handy when there are certain segments that are only enabled in certain situations.

#### Example
The following is an example using three segments. The first and second are diamond type and the first only is enabled when the environment variable `VAR` is present.
image

Setting `style_condition` to `{{ if not (empty .Env.VAR) }}powerline{{end}}` will change the segments style to `powerline` when the environment variable `VAR` is present. 
image

**Note:** `powerline_symbol` was also set to customize the powerline style.


JanDeDobbeleer:
@NoF0rte I like this one, but what if we make `style` also a template? We don't need an additional property if we do that, right?

NoF0rte:
Sure! I didn't want to change any core aspects which is why I went with adding an additional property. Let me update this PR with that change.

JanDeDobbeleer:
@NoF0rte also make sure to change the schema for that. Not sure how to deal with autocomplete in this case but let's figure it out.

NoF0rte:
@JanDeDobbeleer I changed the schema to allow the enum values or anything (I believe). It might not be the best way to do it, but it seems to work.

JanDeDobbeleer:
@NoF0rte I will have a look tomorrow in the morning, thanks!

JanDeDobbeleer:
@NoF0rte added some changes, minor refactor. If you can validate on your end as well, we're good to go.

NoF0rte:
Seems to work just fine on my end @JanDeDobbeleer

JanDeDobbeleer:
@all-contributors please add @NoF0rte for code,design,doc

PR #3275, opened 2022-12-30
Don't load mapstructure config more than once

maxlandon:
This PR ensures that the options for decoding the configuration struct are not set more than once, so that multiple prompt engines can be loaded in the same runtime. Useful for closed-loop applications that make use of different "menus", each with its own prompt context/configuration.


maxlandon:
Sorry, I completely messed up this one: I had forgotten to pull main latest before doing this PR.

PR #3277, opened 2022-12-30
Don't load mapstructure Decode options more than once

maxlandon:
This PR ensures that the options for decoding the configuration struct are not set more than once, so that multiple prompt engines can be loaded in the same runtime. Useful for closed-loop applications that make use of different "menus", each with its own prompt context/configuration.


JanDeDobbeleer:
see #3279 as I can't edit this PR for some reason and you somehow go around creating the PR using the template and miss the conventional commit requirement.

PR #3307, opened 2023-01-04
Update Go module name and imports in code

maxlandon:
Please see the commit message for explanations ! Thanks by advance.
Third attempt...


maxlandon:
I just spent an entire night trying to find the cause of this failing Golang segment test... without success.
I just don't see where that comes from: the prompt compiles, and only import paths where changed...
Forked from latest main.

maxlandon:
I'm available if you need help to debug this. I'm not really sure what and where to look for at this point though...

JanDeDobbeleer:
@maxlandon I'm not a fan of this. Isn't it possible to install `go get github.com/jandedobbeleer/oh-my-posh/src`?

maxlandon:
## Example
Okay let's take the following:
```go
package prompt

import (
	"github.com/jandedobbeleer/oh-my-posh/src/platform"
	"github.com/jandedobbeleer/oh-my-posh/src/properties"
)

type Segment struct {
	Type string
	Path string
}

func (m Segment) Enabled() bool {
	return string(m.Path) != ""
}

func (m Segment) Template() string {
	return string(m.Path)
}

func (m Segment) Init(props properties.Properties, env platform.Environment) {
}
```

Then, without `go get` first:
```bash
❯ go build
main.go:4:2: no required module provides package github.com/jandedobbeleer/oh-my-posh/src/platform; to add it:
        go get github.com/jandedobbeleer/oh-my-posh/src/platform
main.go:5:2: no required module provides package github.com/jandedobbeleer/oh-my-posh/src/properties; to add it:
        go get github.com/jandedobbeleer/oh-my-posh/src/properties
```

So I try to grab the package:
```bash
❯  go get github.com/jandedobbeleer/oh-my-posh/src/properties
go: github.com/jandedobbeleer/oh-my-posh/src@v0.0.0-20230102190908-e957e5f8ccc6: parsing go.mod:
        module declares its path as: github.com/jandedobbeleer/oh-my-posh
                but was required as: github.com/jandedobbeleer/oh-my-posh/src
```

So now let's suppose that I modify the import path:
```go
import (
	"github.com/jandedobbeleer/oh-my-posh/platform"
	"github.com/jandedobbeleer/oh-my-posh/properties"
)
```
And that I try to grab the package again
```bash
❯ go get github.com/jandedobbeleer/oh-my-posh/properties 
go: module github.com/jandedobbeleer/oh-my-posh@upgrade found (v12.34.5+incompatible), but does not contain package github.com/jandedobbeleer/oh-my-posh/properties
```

## Probable explanation
So the problem seems to be this:
- Since the go module declares itself as `github.com/jandedobbeleer/oh-my-posh`, we can correctly grab the code with an equivalent `go get` call.
- However, when importing the lib in the code, we *must* specify the actual filesystem path, therefore appending `src/` to it.
- The problem is that the go.mod is actually in this `src/` directory, which makes the Go build process look the `src/properties` part of the import *relatively* to the go.mod, which then can't find `src/`

Currently, the only way I found around this is to give the following replace directive in the consumer's go.mod:
```go
replace github.com/jandedobbeleer/oh-my-posh v12.34.2+incompatible => github.com/jandedobbeleer/oh-my-posh/src v0.0.0-20221230190448-3559254292e5

```
Which is really prone to errors, would be required to anyone using the `oh-my-posh` module, and moreover, it requires the get the `replace` syntax with versions right, etc... Not very nice...

I hope I could explain all of this clearly !

maxlandon:
OKAY GOT IT !
The referenced relative path of `oh-my-posh` go.mod had to be changed in the `golang_test.go` file.

JanDeDobbeleer:
@maxlandon shouldn't we then simply change the mod file to use the src folder in the module definition as well?

maxlandon:
Oh I didn't think of that.
Let me check and try.

maxlandon:
I'm not sure how to proceed. Do you have an idea ?

maxlandon:
Okay so the last commit replaces the go.mod/go.sum in src/, and adds src/ to the go.mod declaration.
Seems to work fine on my side.

There is no way to go around all the import paths change in your code though...

maxlandon:
Library users would thus grab like this:
```bash
go get github.com/jandedobbeleer/oh-my-posh/src
```

And use like this:
```go
import (
      "github.com/jandedobbeleer/oh-my-posh/src/segments"
)

Issue #3388, opened 2023-01-22
Remove first newline on start and clear when `"newline": true`


CodeF53:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

With a theme that has `"newline": true` in the first block, there will be a leading newline when starting the terminal and when running `clear`
![image](https://user-images.githubusercontent.com/37855219/213928189-10378a69-2190-43cc-9390-4d36a1ea85d8.png)
![image](https://user-images.githubusercontent.com/37855219/213928261-18b309f8-5138-4a9d-9c30-14bdc75b68eb.png)
Ideally, newlines would be ignored in these cases.

CodeF53:
I apologize if this is a duplicate, searching issues for `is:issue "newline": true` gave 124 closed issues, and I couldn't look through them all.

JanDeDobbeleer:
@CodeF53 I'm working on this as we speak.

mbs-c:
I just upgraded to 13.6.0 and I do not think the second part of this bug is fixed properly: The initial prompt now omits the extra newline, but the unwanted newline after `clear` is still present in both versions of PowerShell.

JanDeDobbeleer:
@mbs-c Clear isn't something we capture. It moves everything to the top of the screen, not sure if there's a low impact way to see if we're on the first line that's also fool proof.

This isn't a bug either, it's just behavior.

mbs-c:
@JanDeDobbeleer
I can imagine it's hard to capture, I just wanted to point out that the initial description mentions `clear` as well and that this part hasn't changed. Personally, the behavior after `clear` even bothers me more than the behavior on initial load, because I didn't mind the newline after all the verbose PowerShell messages as much.

You're also absolutely right that this isn't a "bug" per se, but rather a feature request. Personally, I am trying to re-create powerlevel10k's behavior, which only shows a newline before the prompt if there's a transient prompt from a previous command on the screen, and that does not seem to be possible at the moment.

**Edit**: I am basically halfway there with the following configuration that's supposed to mimic a single-line powerlevel10k prompt:
```js
{
  // [snip]
  "blocks": [
    {
      "newline": true,
      "segments": [
        // [snip] long prompt configuration
      ],
      "type": "prompt"
    },
    {
      "alignment": "right",
      "segments": [
        // [snip] long rprompt configuration
      ],
      "type": "rprompt"
    }
  ],
  "transient_prompt": {
[...]

PR #3393, opened 2023-01-24
feat: ignore empty header line on clear

JanDeDobbeleer:
relates to #3388

### Prerequisites

### Status

- [x]  ZSH
- [x] Bash
- [x] PowerShell
- [x] CMD
- [x] Fish - blocked on https://github.com/fish-shell/fish-shell/issues/6610
- [x] Nushell - no builtin for the line number and can't yet handle echo-ing the ANSI sequence

### DIFF
--- a/src/engine/engine.go
+++ b/src/engine/engine.go
@@ -64,8 +64,9 @@ func (e *Engine) canWriteRightBlock(rprompt bool) bool {
 func (e *Engine) PrintPrimary() string {
 	// cache a pointer to the color cycle
 	cycle = &e.Config.Cycle
+	firstLine := e.Env.Getenv("POSH_CURSOR_LINE") == "1"
 	for i, block := range e.Config.Blocks {
-		e.renderBlock(block, (i == 0 && e.PromptCount == 1))
+		e.renderBlock(block, (i == 0 && (e.PromptCount == 1 || firstLine)))
 	}
 	if len(e.Config.ConsoleTitleTemplate) > 0 {
 		title := e.getTitleTemplateText()
--- a/src/shell/scripts/omp.bash
+++ b/src/shell/scripts/omp.bash
@@ -9,6 +9,14 @@ PS0='${omp_start_time:0:$((omp_start_time="$(_omp_start_timer)",0))}'
 # set secondary prompt
 PS2="$(::OMP:: print secondary --config="$POSH_THEME" --shell=bash --shell-version="$BASH_VERSION")"
 
+function _set_posh_cursor_position() {
+  echo -ne "\033[6n"            # ask the terminal for the position
+  read -s -d\[ garbage          # discard the first part of the response
+  read -s -d R pos              # store the position in bash variable 'pos'
+  export POSH_CURSOR_LINE=${pos%;*}
+  export POSH_CURSOR_COLUMN=${pos#*;}
+}
+
 function _omp_start_timer() {
     ::OMP:: get millis
 }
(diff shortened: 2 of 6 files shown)


JanDeDobbeleer:
@chrisant996 challenge/question for you. I'm looking for a way to store the cursor's position (and primarily the line we're on) in an environment variable called `POSH_CURSOR_LINE`. In bash and zsh we leverage an ANSI sequence to print the location, then parse the values from that output. Is there a way to get the information from lua, or can we also read the output by printing that ANSI sequence silently, fetch and parse the content (and will the lua process have that information)?

chrisant996:
@JanDeDobbeleer 

re: Clink on ARM -- Some users have confirmed that Clink is working on ARM in v1.4.7 and higher. Have you tried with a recent version?

re: "store the cursor's position" -- I don't quite understand what's desired or how it will be used. Can you elaborate? Why does omp need to store the cursor? What does it do with the position? Is omp trying to simulate the `ESC [ s` and `ESC [ u` codes manually? Is omp trying to write stuff above or below (or in??) the input line?

JanDeDobbeleer:
@chrisant996 for this feature we need to know what line we're on so that we can decide whether or not to print a first line break (empty line at the top). The way to do this is by getting the cursor position from the shell (builtin for PowerShell but also retrievable using a specific ANSI sequence. And CMD is still on my list to see if this can be solved.

> Some users have confirmed that Clink is working on ARM in v1.4.7 and higher. Have you tried with a recent version?

Works! Awesome!

chrisant996:
> @chrisant996 for this feature we need to know what line we're on so that we can decide whether or not to print a first line break (empty line at the top).

@JanDeDobbeleer When?

- Do you mean the cursor position while the user is moving the cursor around while editing in the input line?  That isn't available, and I don't think that's something to mess around with.
- Or do you mean the cursor position right before the prompt gets displayed?  In a callback function registered via [clink.onbeginedit()](https://chrisant996.github.io/clink/clink.html#clink.onbeginedit) you could use [console.getnumlines()](https://chrisant996.github.io/clink/clink.html#console.getnumlines) minus [console.gettop()](https://chrisant996.github.io/clink/clink.html#console.gettop).

Is the goal to control how much blank space shows up between the end of the previous command, and the prompt for the current input command line?  If so, then why would omp need to know the cursor position?

If that's the goal, then with Clink you can even compact multiple blank lines before printing the prompt.  E.g. [clink-flex-prompt](https://github.com/chrisant996/clink-flex-prompt) does fancy stuff with that; [see here](https://github.com/chrisant996/clink-flex-prompt/blob/bec2bff1133bb028da2dc08a98e992130b9e8c38/flexprompt.lua#L1697-L1717).  That's the only reason I can think of how omp might want to use the cursor position, but none of the other shells can read from the terminal display, so that can't be how omp is using the cursor position.

JanDeDobbeleer:
@chrisant996 we only want to know if we're on the first line, so if those functions allow for that I can see how to implement that. Thanks.

chrisant996:
@JanDeDobbeleer Oh, do you want to know if the prompt is being shown at the top of the terminal window's visible area?

Something like this, though I haven't tried it and might be off by 1 or something:

```
if console.getnumlines() == console.get top() then
```

But I don't know for sure what console.getnumlines() reports before anything has been printed yet.

JanDeDobbeleer:
@chrisant996 yes! I'll give that a whirl. Thanks a lot!

PR #3528, opened 2023-02-25
feat: add Unity segment

cabauman:
### Prerequisites

### Description

adds a new segment that displays the active Unity and C# versions

I'm sure there will be feedback to address but ready to kick off the review process!

### DIFF
--- a/src/segments/unity.go
+++ b/src/segments/unity.go
@@ -25,7 +25,7 @@ func (u *Unity) Enabled() bool {
 		u.env.Error(err)
 		return false
 	}
-	if unityVersion == "" {
+	if len(unityVersion) == 0 {
 		return false
 	}
 	u.UnityVersion = unityVersion
@@ -42,7 +42,8 @@ func (u *Unity) Enabled() bool {
 func (u *Unity) GetUnityVersion() (version string, err error) {
 	versionFilePath := filepath.Join("ProjectSettings", "ProjectVersion.txt")
 	if !u.env.HasFiles(versionFilePath) {
-		return "", nil
+		u.env.Debug("No ProjectVersion.txt file found")
+		return
 	}
 
 	versionFileText := u.env.FileContent(versionFilePath)
--- a/src/segments/unity.go
+++ b/src/segments/unity.go
@@ -40,13 +40,20 @@ func (u *Unity) Enabled() bool {
 }
 
 func (u *Unity) GetUnityVersion() (version string, err error) {
-	versionFilePath := filepath.Join("ProjectSettings", "ProjectVersion.txt")
-	if !u.env.HasFiles(versionFilePath) {
+	projectDir, err := u.env.HasParentFilePath("ProjectSettings")
+	if err != nil {
+		u.env.Debug("No ProjectSettings parent folder found")
+		return
+	}
+
+	if !u.env.HasFilesInDir(projectDir.Path, "ProjectVersion.txt") {
 		u.env.Debug("No ProjectVersion.txt file found")
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


src/segments/unity.go
```diff
@@ -0,0 +1,104 @@
...
+	CSharpVersion string
+}
+
+func (u *Unity) Enabled() bool {
+	unityVersion, err := u.GetUnityVersion()
+	if err != nil {
+		u.env.Error(err)
+		return false
+	}
+	if unityVersion == "" {
```
JanDeDobbeleer:
```suggestion
	if len(unityVersion) == 0 {
```

src/segments/unity.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	u.CSharpVersion = csharpVersion
+
+	return true
+}
+
+func (u *Unity) GetUnityVersion() (version string, err error) {
+	versionFilePath := filepath.Join("ProjectSettings", "ProjectVersion.txt")
+	if !u.env.HasFiles(versionFilePath) {
+		return "", nil
```
JanDeDobbeleer: I would log this as it helps with debugging more explicitly.

```suggestion
        u.env.Debug("No ProjectVersion.txt file found")
		return "", nil
```

And I'm pretty sure this can be a return statement only as you have named variables.

src/segments/unity.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	versionFileText := u.env.FileContent(versionFilePath)
+	firstLine := strings.Split(versionFileText, "\n")[0]
```
JanDeDobbeleer, line 57: Do we want to validate the length before splitting? This will always work, but do we expect more lines?

JanDeDobbeleer (changes requested):

src/segments/unity.go
cabauman: From my experience the file always has 3 lines (last one empty):

```
m_EditorVersion: 2019.4.13f1
m_EditorVersionWithRevision: 2019.4.13f1 (518737b1de84)

```

cabauman: Would you like me to make the check, and abort if there are not exactly 2 lines? I figured as long as that first line passes the check then it wouldn't hurt to disregard the rest.

src/segments/unity.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		u.env.Error(err)
+	}
+	u.CSharpVersion = csharpVersion
+
+	return true
+}
+
+func (u *Unity) GetUnityVersion() (version string, err error) {
+	versionFilePath := filepath.Join("ProjectSettings", "ProjectVersion.txt")
```
JanDeDobbeleer: This will only work when you're inside the root folder, correct? If so, what we can also do is use `u.env.HasParentFilePath("ProjectSettings")` and then validate if have the correct file using `u.env.HasFiles("ProjectVersion.txt")`. For example:

```golang
projectDir, err := g.env.HasParentFilePath("ProjectSettings")
if err != nil {
    u.env.Debug("No ProjectSettings parent folder found")
	return
}

if !u.env.HasFilesInDir(projectDir.Path, "ProjectVersion.txt") {
    u.env.Debug("No ProjectVersion.txt file found")
	return
}

versionFilePath := filepath.Join(projectDir.Path, "ProjectVersion.txt")
versionFileText := u.env.FileContent(versionFilePath)
```

cabauman: Ah, yes. Good suggestion.

JanDeDobbeleer:
@all-contributors please add @cabauman for code,design,doc

PR #3563, opened 2023-03-07
feat: add node, python, dotnet to context

leiyu66:
### Prerequisites

### Description

### DIFF
--- a/themes/cloud-context.omp.json
+++ b/themes/cloud-context.omp.json
@@ -157,6 +157,30 @@
           "powerline_symbol": "\ue0b4",
           "template": "{{ .HEAD }}{{ if .Staging.Changed }} \uf046 {{ .Staging.String }}{{ end }}{{ if and (.Working.Changed) (.Staging.Changed) }} |{{ end }}{{ if .Working.Changed }} \uf044 {{ .Working.String }}{{ end }}{{ if gt .StashCount 0 }} \uf692 {{ .StashCount }}{{ end }} ",
           "type": "git"
+        },
+        {
+          "background": "p:background-color",
+          "foreground": "p:symbol-color",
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@GavinOsborn can you review, it's your theme? Approving is also fine.

GavinOsborn:
Would love to review. To set expectations though - I won't get to it until the weekend.

GavinOsborn:
Thanks for suggesting enhancements to my theme 👍

I did actually think about adding runtime support to the theme when I first created it. I was concerned about a few different things:

- Signal -> Noise. I suppose it's a little subjective, but I tried to set a high bar for the value of the information the theme contained to keep it clean yet dense with high value info.
  - Most of the information in the theme provides context to help save an engineer from making costly mistakes - deleting resources in the wrong environment, pushing to the wrong branch, etc.
  - How often to engineers truly need to know the runtime version they are using where having it available justifies it appearing on every line in the terminal?
  - For those moments where you truly do need to know the full version - you're probably going to reach for `dotnet info`, `node version` anyway, right?
- If we add `node`, `dotnet` and `python` where does it stop?
  - do we end up supporting all of the languages?
  - do we then also add other platform specific tooling like package managers, Functions and Terraform all into this theme?
  - at what point does this grow to introduce significant performance penalties with all these new segment types?
- The quality of the iconography available for some platforms is quite poor.
  - dotnet in particular. Rather than being symbolic of anything I felt most folks would mistake it for a `~` - I probably would have.
  - It took me forever to find icons that I was happy with for the initial theme 😅 

I would love to know your thoughts.

JanDeDobbeleer:
@GavinOsborn I agree, @leiyu66 I would adjust the config locally to serve your needs.

PR #3569, opened 2023-03-10
fix(wakatime)!: fix struct `wtData` binding from API response, from `Cummulative` to `Cumulative`

CodexLink:
### Prerequisites

- [X] I have read and understood the [contributing guide][CONTRIBUTING.md]
- [X] Docs have been added/updated (for bug fixes/features)

### Description

**This is a breaking change.**

It's been a while since I noticed that the Wakatime segment for my theme has disappeared. Upon inspection, Wakatime has changed one of the returned JSON's parent key names from `cummulative_total` to `cumulative_total`.

The following is just the excerpt of the payload as of now:

```json
{
  "cumulative_total": {
    "decimal": "0.02",
    "digital": "0:01",
    "seconds": 109.270097,
    "text": "1 min"
  },
  "daily_average": {},
  "data": [ ],
  "end": "2023-03-11T15:59:59Z",
  "start": "2023-03-10T16:00:00Z"
}
```

The unexpected change was _assumed in the first place_ since I have [snapshots of my theme](https://github.com/CodexLink/chips.omp.json#git-states-segment) that shows the Wakatime segment. After finding the problem, it is somehow true that the API changed. But upon (light) research, I can't seem to find any evidence that their API has changed, but either way ...

This PR fixes that by **correcting the spelling** from **cummulative** to **cumulative** for both **component** and **documentation** hence, _becoming a breaking change_. Other than that, the PR also contains the following:

- Added an additional checking that emits an error stating that the response returned from the Wakatime API should be checked due to struct `wtData` containing _an empty context_/_is an empty struct_.
- Added an additional context from the fake payload (on _wakatime_test.go_) to avoid tripping the test cases.

> Note that the additional checker (for the `wtData`) from the _wakatime.go_ was not added to the part of test cases since it's going to be a bit of a hassle vs. adding the checker itself from the main code segment instead.

> Plus, the test cases (_exclusive to wakatime_) are just emulating the context that the API returned. Declaring another test case will have other test cases to adjust because the `response` must be compatible with every test case declared. _But in the end, it's possible to include it in the test case, but there will be adjustments_.

### Disclaimer

This is **my first time using Golang**, suggestions and requests for changes are extremely welcome.

### DIFF
--- a/src/segments/wakatime.go
+++ b/src/segments/wakatime.go
@@ -2,6 +2,7 @@ package segments
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/platform"
 	"github.com/jandedobbeleer/oh-my-posh/src/properties"
@@ -21,13 +22,13 @@ type wtTotals struct {
 }
 
 type wtData struct {
-	CummulativeTotal wtTotals `json:"cummulative_total"`
-	Start            string   `json:"start"`
-	End              string   `json:"end"`
+	CumulativeTotal wtTotals `json:"cumulative_total"`
+	Start           string   `json:"start"`
+	End             string   `json:"end"`
 }
 
 func (w *Wakatime) Template() string {
-	return " {{ secondsRound .CummulativeTotal.Seconds }} "
+	return " {{ secondsRound .CumulativeTotal.Seconds }} "
 }
 
 func (w *Wakatime) Enabled() bool {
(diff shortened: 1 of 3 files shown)


src/segments/wakatime.go
```diff
@@ -63,6 +64,13 @@ func (w *Wakatime) setAPIData() error {
 		return err
 	}
 
+	// In case of Wakatime API changes, validate if one of the fields of `w.wtData`, is empty.
+	if (w.wtData.CumulativeTotal == wtTotals{}) ||
```
JanDeDobbeleer: this isn't necessary as users will report a change and the logs will allow us to easily see what the issue is

JanDeDobbeleer (changes requested):

src/segments/wakatime.go
CodexLink: Oh understood, I also find this quite awkward to put as well (_though I just still put it anyway yesterday_), this will also remove the additional items from the fake payload that I have added from the test then.

JanDeDobbeleer:
@all-contributors please add @CodexLink for code,doc

PR #3627, opened 2023-03-24
feat(diamond): allow joining diamonds on empty leading

### LINKED ISSUE #3623: Add option powerline_skipLeadingIfFirst to skip the leading symbol if it is the first item in the line
### Code of Conduct

```
blocks:
  - type: prompt
    alignment: left
    segments:
      - type: session
        style: powerline
        background: "#569CD6"
        foreground: "#333333"
        template: "{{ if .SSHSession }}🖥️ {{ .UserName }}@{{ .HostName }}{{ end }}"
      - type: path
        style: powerline
        powerline_symbol: 
        powerline_skipLeadingIfFirst: true
        background: "#333333"
        template: "{{ .Path }}"
```

Expected:
image

If the first segment session is not rendered (shows as false in the print debug), then the next segment should basically take on the shape of "diamond" with trailing_diamond set to the powerline character and no leading_diamond. The code I used to make the above is as follows:
```
      - type: path
        style: diamond
        trailing_diamond: 
        background: "#333333"
        template: "{{ .Path }}"
```

If there is something there, then it should behave like powerline as normal

Actual:
image

JanDeDobbeleer:
feat(diamond): allow joining diamonds on empty leading

resolves #3623

### DIFF
--- a/src/engine/block.go
+++ b/src/engine/block.go
@@ -150,7 +150,6 @@ func (b *Block) renderActiveSegment() {
 	case Diamond:
 		b.writer.Write(ansi.Transparent, ansi.Background, b.activeSegment.LeadingDiamond)
 		b.writer.Write(ansi.Background, ansi.Foreground, b.activeSegment.text)
-		b.writer.Write(ansi.Transparent, ansi.Background, b.activeSegment.TrailingDiamond)
 	case Accordion:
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
started over in #3629

PR #3684, opened 2023-04-06
feat(argocd): init commit for argocd context

ksdpmx:
### Prerequisites

### Description

add support for prompting the current argocd context in oh-my-posh

### How

add this new argocd segment in the configuration

### DIFF
--- a/src/engine/segment.go
+++ b/src/engine/segment.go
@@ -98,6 +98,8 @@ const (
 
 	// ANGULAR writes which angular cli version us currently active
 	ANGULAR SegmentType = "angular"
+	// ARGOCD writes the current argocd context
+	ARGOCD SegmentType = "argocd"
 	// AWS writes the active aws context
 	AWS SegmentType = "aws"
 	// AZ writes the Azure subscription info we're currently in
@@ -246,6 +248,7 @@ const (
 // Consumers of the library can also add their own segment writer.
 var Segments = map[SegmentType]func() SegmentWriter{
 	ANGULAR:       func() SegmentWriter { return &segments.Angular{} },
+	ARGOCD:        func() SegmentWriter { return &segments.Argocd{} },
 	AWS:           func() SegmentWriter { return &segments.Aws{} },
 	AZ:            func() SegmentWriter { return &segments.Az{} },
 	AZFUNC:        func() SegmentWriter { return &segments.AzFunc{} },
--- a/src/segments/argocd.go
+++ b/src/segments/argocd.go
@@ -1,0 +1,110 @@
+package segments
+
+import (
+	"errors"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+	"github.com/spf13/pflag"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	argocdOptsEnv     = "ARGOCD_OPTS"
+	argocdInvalidFlag = "invalid flag"
+	argocdInvalidYaml = "invalid yaml"
+	argocdNoCurrent   = "no current context"
+
+	NameTemplate = " {{ .Name }} "
+)
+
+type ArgocdContext struct {
+	Name   string `yaml:"name"`
+	Server string `yaml:"server"`
+	User   string `yaml:"user"`
+}
+
+type ArgocdConfig struct {
+	Contexts       []*ArgocdContext `yaml:"contexts"`
+	CurrentContext string           `yaml:"current-context"`
+}
+
+type Argocd struct {
+	props properties.Properties
+	env   platform.Environment
+
+	ArgocdContext
+}
+
+func (a *Argocd) Template() string {
+	return NameTemplate
+}
+
+func (a *Argocd) Init(props properties.Properties, env platform.Environment) {
+	a.props = props
+	a.env = env
+}
+
+func (a *Argocd) Enabled() bool {
+	// always parse config instead of using cli to save time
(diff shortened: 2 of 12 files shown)


src/engine/segment.go
```diff
@@ -99,7 +99,8 @@ const (
 	// ANGULAR writes which angular cli version us currently active
 	ANGULAR SegmentType = "angular"
 	// AWS writes the active aws context
-	AWS SegmentType = "aws"
+	ARGOCD SegmentType = "argocd"
```
JanDeDobbeleer, line 102: Can you add the same (I know) comment for consistency?

src/segments/argocd.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if len(cpo) > 0 {
+		cp = cpo
+	}
+	return cp
+}
+
+func (a *Argocd) getConfigFromOpts() (string, error) {
+	// don't exit/panic when encountering invalid flags
+	a.flags = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
```
JanDeDobbeleer: why do we use all flags here? Can't we only add the config flag and get that value? I would also not make that part of the struct, but an inline variable as it's only used here.

src/segments/argocd_test.go
```diff
@@ -0,0 +1,233 @@
...
+		env := new(mock.MockedEnvironment)
+		env.On("Getenv", argocdOptsEnv).Return(tc.Opts)
+
+		argocd := &Argocd{
+			env:   env,
+			props: properties.Map{},
+		}
+		if len(tc.ExpectedError) > 0 {
+			_, err := argocd.getConfigFromOpts()
+			assert.EqualError(t, err, tc.ExpectedError, tc.Case)
```
JanDeDobbeleer: can you do an early return here instead of if/else?

```suggestion
			assert.EqualError(t, err, tc.ExpectedError, tc.Case)
			continue
```

src/segments/argocd_test.go
```diff
@@ -0,0 +1,233 @@
...
+		argocd := &Argocd{
+			env:   env,
+			props: properties.Map{},
+		}
+		assert.Equal(t, tc.Expected, argocd.getConfigPath())
+	}
+}
+
+func TestArgocdParseConfig(t *testing.T) {
+	configFile := "../test/argocd.yml"
```
JanDeDobbeleer: this doesn't exist, right?

src/segments/argocd_test.go
```diff
@@ -0,0 +1,233 @@
...
+		env.On("FileContent", configFile).Return(tc.Config)
+		env.On("Error", mock2.Anything).Return()
+
+		argocd := &Argocd{
+			env:   env,
+			props: properties.Map{},
+		}
+		if len(tc.ExpectedError) > 0 {
+			_, err := argocd.parseConfig(configFile)
+			assert.EqualError(t, err, tc.ExpectedError, tc.Case)
```
JanDeDobbeleer, line 166: early return instead of if/else

```suggestion
			assert.EqualError(t, err, tc.ExpectedError, tc.Case)
			continue
```

website/docs/segments/argocd.mdx
```diff
@@ -0,0 +1,39 @@
...
+## Template ([info][templates])
+
+:::note default template
+
+```template
+{{ .Context.Name }}
+```
+### Properties
+
+| Name       | Type     | Description                  |
```
JanDeDobbeleer: the table needs to be formatted correctly (so all columns align).

JanDeDobbeleer (changes requested):

website/docs/segments/argocd.mdx
ksdpmx: columns have been aligned.

src/engine/segment.go
ksdpmx: fixed with the correct comment.

src/segments/argocd_test.go
ksdpmx: correct. it's just some kind of placeholder here. because we may need to parse different configs in this UT.

src/segments/argocd_test.go
ksdpmx: updated with early return.

src/segments/argocd_test.go
ksdpmx: updated with early return.

src/segments/argocd.go
ksdpmx: > why do we use all flags here

~because i'm trying to tell the difference between valid argocd flags and invalid ones. so that we can handle the use case like [this](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3684/files#diff-b58057c229fe7937c24b999999dad7b3bdefbb75e3513b6d0922d1b1b9eaf412R24). if those valid flags got removed, pflags would fail to retrieve the config flag we need.~

or we may use `FlagSet.ParseErrorsWhitelist`..

ksdpmx: > why do we use all flags here

removed those unused valid flags and removed the error. updated UT.

ksdpmx: > an inline variable as it's only used here

agreed and updated.

ksdpmx:
all issues have been fixed.

PTAL again @JanDeDobbeleer. Thanks:)

src/segments/argocd_test.go
ksdpmx: updated with the same config file like other UT.

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @ksdpmx for code,doc

PR #3703, opened 2023-04-09
Update customize.mdx

mavaddat:
Update the Lua for the cmd config so it uses the env var `POSH_THEMES_PATH` set by `oh-my-posh.exe`.

### Prerequisites

### Description

Updated the `oh-my-posh.lua` so that it uses the environment variable `POSH_THEMES_PATH` which is set my `oh-my-posh`.

### How

Using `os.getenv(POSH_THEMES_PATH")` as in 

```lua
load(io.popen('oh-my-posh init cmd --config ' .. os.getenv("POSH_THEMES_PATH") .. '/jandedobbeleer.omp.json'):read("*a"))()
```

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1643,6 +1643,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "kema-dev",
+      "name": "kema",
+      "avatar_url": "https://avatars.githubusercontent.com/u/54537427?v=4",
+      "profile": "http://www.kemadev.fr/fr/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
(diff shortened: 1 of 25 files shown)


JanDeDobbeleer:
@mavaddat there are two commits in here and they should both respect the conventional commit guidelines. I'll review first as the change in the install script is interesting. I don't see why we would support older PowerShell versions in 2023.

mavaddat:
> @mavaddat there are two commits in here and they should both respect the conventional commit guidelines.

Yes, I was going to come back and update the commit messages to comply with the Conventional Commit style. Sorry about my do-first-then-fix contribution approach... 😳 I am updating the commits now.

> I'll review first as the change in the install script is interesting. I don't see why we would support older PowerShell versions in 2023.

Some of us (e.g., _me_) are locked into enterprise version freeze environments where we cannot install or update our Powershell.

JanDeDobbeleer:
@mavaddat gotcha, I'll try to review tomorrow, could be I adjust and merge myself.

JanDeDobbeleer:
@mavaddat see #3724. As for the change in the docs, that's not a valid one. It clearly states we're dealing with a local file and not the theme's folder.

PR #3726, opened 2023-04-16
feat(tooltips zsh): match aliases

kema-dev:
### Prerequisites

### Description

When using zsh, show tooltips even on aliases of specified tips

### How

Get the first word of the command line and get its alias if possible before feeding tooltips engine


JanDeDobbeleer:
@kema-dev this is supported using the tips array, if we do this for one shell we have to for every shell but it's once again more logic that gets executed during execution so I'm not in favour of adding this.

PR #3743, opened 2023-04-22
fix(pwsh): erase a previous tooltip before a new one is rendered

lewis-yeung:
### Prerequisites

### Description

Fix a bug that a previous tooltip can be cut off and partially preserved in PowerShell, if the new one is shorter.

### How

Use the "CSI K" escape sequence to clear from cursor to the end of the line but preserve the cursor position so that a previous tooltip can be completely erased before a new one is rendered.

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -133,7 +133,12 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
             }
             $script:ToolTipCommand = $command
             $standardOut = @(Start-Utf8Process $script:OMPExecutable @("print", "tooltip", "--error=$script:ErrorCode", "--shell=$script:ShellName", "--pswd=$cleanPSWD", "--config=$env:POSH_THEME", "--command=$command", "--shell-version=$script:PSVersion"))
-            Write-Host $standardOut -NoNewline
+            if ($standardOut -ne '') {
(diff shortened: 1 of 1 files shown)


src/shell/scripts/omp.ps1
```diff
@@ -133,7 +133,12 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
             }
             $script:ToolTipCommand = $command
             $standardOut = @(Start-Utf8Process $script:OMPExecutable @("print", "tooltip", "--error=$script:ErrorCode", "--shell=$script:ShellName", "--pswd=$cleanPSWD", "--config=$env:POSH_THEME", "--command=$command", "--shell-version=$script:PSVersion"))
-            Write-Host $standardOut -NoNewline
+            if ($standardOut -ne '') {
+                # clear from cursor to the end of the line in case a previous tooltip is cut off and partially preserved,
+                # if the new one is shorter
+                Write-Host "`e[K" -NoNewline
```
JanDeDobbeleer, line 139: @lewis-yeung maybe we should add this to the engine rather than here? We can easily do it there using the ansi helper functions.

lewis-yeung: Makes sense.

lewis-yeung: @JanDeDobbeleer See #3745.

PR #3796, opened 2023-05-05
feat(python): add support for pyvenv.cfg prompt

javidcf:
### Prerequisites

### Description

Python virtual environments can be created with a user-specified prompt associated to them that overrides the default behaviour of using the environment directory name (see [`venv` module documentation](https://docs.python.org/3/library/venv.html)). This PR implements using that user-specified prompt instead of the environment directory name when provided.

Please advise if the documentation for `.Venv` should be updated (the current description, "the virtual environment name (if present)", seems still appropriate).

### How

If a custom prompt is given on environment creation, it is stored in `pyvenv.cfg`, a file that, according to [PEP 405](https://peps.python.org/pep-0405/) may be found either adjacent to the Python executable or one directory above it (if the executable is a symlink, it is not dereferenced)".

The changes search for a `pyvenv.cfg` file in the relevant locations and for a `prompt` value within that file. If found, the environment directory name resolution is omitted and the retrieved value is used as `.Venv` instead.

### DIFF
--- a/src/segments/python.go
+++ b/src/segments/python.go
@@ -63,6 +63,10 @@ func (p *Python) loadContext() {
 	if !p.language.props.GetBool(FetchVirtualEnv, true) {
 		return
 	}
+	if prompt := p.pyvenvCfgPrompt(); len(prompt) > 0 {
+		p.Venv = prompt
+		return
+	}
 	venvVars := []string{
 		"VIRTUAL_ENV",
 		"CONDA_ENV_PATH",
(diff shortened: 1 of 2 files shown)


src/segments/python.go
```diff
@@ -63,21 +63,26 @@ func (p *Python) loadContext() {
...
-		"VIRTUAL_ENV",
-		"CONDA_ENV_PATH",
-		"CONDA_DEFAULT_ENV",
-	}
-	var venv string
-	for _, venvVar := range venvVars {
-		venv = p.language.env.Getenv(venvVar)
-		if len(venv) == 0 {
-			continue
+	prompt := p.pyvenvCfgPrompt()
```
JanDeDobbeleer: This can be an early return instead a full if/else block:

```go
prompt := p.pyvenvCfgPrompt()
if len(prompt) > 0 {
    p.Venv = prompt
    return
}
```

src/segments/python.go
```diff
@@ -142,3 +147,32 @@ func (p *Python) pyenvVersion() (string, error) {
 	}
 	return parts[0], nil
 }
+
+func (p *Python) pyvenvCfgPrompt() string {
+	path := p.language.env.CommandPath("python")
+	if len(path) == 0 {
+		path = p.language.env.CommandPath("python3")
+	}
+	if len(path) > 0 {
```
JanDeDobbeleer: this one can also be an early return:

```go
if len(path) == 0 {
    return ""
}
```

JanDeDobbeleer (changes requested):

javidcf (commented):
Added early returns

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @javidcf for code

This is pure magic 🪄! @holopin-bot @javidcf wizard

PR #3804, opened 2023-05-06
Update legacy folder icon `f74a`

tzuhanchang:
`f74a` (nf-mdi-folder) has been removed from Nerd Fonts, see https://www.nerdfonts.com/cheat-sheet. It is replaced with `\udb80\ude4b` (nf-md-folder).


JanDeDobbeleer:
@tzuhanchang I'm going to update the entire set now that V3 is live. It needs changes everywhere.

PR #3882, opened 2023-05-20
Fix theme schema & template cache

lewis-yeung:
### Prerequisites

- [ ] Docs have been added/updated (for bug fixes / features)

### Description

1. Add `upgrade` to the list of valid segment types in the theme schema.
2. Avoid duplicate creations of template cache. Related to #2885.

### How

1. Update the theme schema (`themes/schema.json`).
2. In package `platform`:
   - Add an `initialized` field to `TemplateCache` as an indicator.
   - Assign an initial value to the `tmplCache` field of `Shell` in the `Init` method.
   - When calling `(*Shell).TemplateCache`, use the R/W lock and check the `initialized` flag to determine whether the initialization should be performed.

### DIFF
--- a/src/platform/shell.go
+++ b/src/platform/shell.go
@@ -190,6 +190,7 @@ type TemplateCache struct {
 	SHLVL        int
 	Segments     SegmentsCache
 
+	initialized bool
 	sync.RWMutex
 }
 
@@ -312,6 +313,7 @@ func (env *Shell) Init() {
 	env.cmdCache = &commandCache{
 		commands: NewConcurrentMap(),
 	}
+	env.tmplCache = &TemplateCache{}
 	env.SetPromptCount()
 }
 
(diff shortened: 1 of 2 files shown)


src/platform/shell.go
```diff
@@ -785,60 +788,61 @@ func (env *Shell) Logs() string {
...
-		Segments:     make(map[string]interface{}),
-		PromptCount:  env.CmdFlags.PromptCount,
-	}
-	tmplCache.Env = make(map[string]string)
-	tmplCache.Var = make(map[string]interface{})
+	tmplCache := env.tmplCache
+	tmplCache.Lock()
+	defer tmplCache.Unlock()
+
+	if !tmplCache.initialized {
```
JanDeDobbeleer: can we invert this one to avoid having this whole block indented?

```go
if tmplCache.initialized {
    return tmplCache
}
// init here
```

JanDeDobbeleer (changes requested):

src/platform/shell.go
lewis-yeung: Yes, we can.

PR #3926, opened 2023-06-01
feat(Sitecore): upgraded Sitecore segment to support Sitecore CLI context switching 

dorokhin-bohdan:
### Prerequisites

### <samp>🤖 Generated by Copilot at c14f4a8</samp>

Refactored and improved the `sitecore` segment to use a new logic based on the user.json file and added a new option to display the default icon. Updated the tests, schema, and documentation accordingly.

### <samp>🤖 Generated by Copilot at c14f4a8</samp>

*  Simplify import statements and remove unused packages in `sitecore.go` and `sitecore_test.go` ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-e4e61b9f252fb291a7129580c917d7dfe5178f2d12c9bdb8c3a28a0fdcfe802dL5-R5), [link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-7ba0e717460746e6e69e829d8abb6927bc2834beb817f3da6948f25f2285b80dL4-R8))
*  Modify `Sitecore` and `SitecoreConfig` structs to use different fields and types for Sitecore environment configuration ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-e4e61b9f252fb291a7129580c917d7dfe5178f2d12c9bdb8c3a28a0fdcfe802dL12-R62))
*  Replace `shouldDisplay` method with `Enabled` method that implements new logic for checking and parsing sitecore.json and user.json files and setting `EndpointName` and `CmHost` fields ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-e4e61b9f252fb291a7129580c917d7dfe5178f2d12c9bdb8c3a28a0fdcfe802dL45-R96))
*  Rewrite `TestSitecoreSegment` function to use different test cases and mock data for the new logic of the segment ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-7ba0e717460746e6e69e829d8abb6927bc2834beb817f3da6948f25f2285b80dL17-R161))
*  Add `sitecore` segment name to the list of valid segment names in `themes/schema.json` ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-892c540c9bfebcb3bb8a0be1714201cd17eb6f0d40367c09f80d56e883b7335eR295))
*  Add `display_default` property to the schema of the `sitecore` segment in `themes/schema.json` with description, type, and default value ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-892c540c9bfebcb3bb8a0be1714201cd17eb6f0d40367c09f80d56e883b7335eL1941-R1954))
*  Update sample configuration and properties sections of the `sitecore` segment documentation in `website/docs/segments/sitecore.mdx` to match the new logic and output of the segment ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-3af710d4f9882cc603875c32893b1e701198ef8218f9be38d2a4ffb3a1348819L9-R27))
*  Simplify template section of the `sitecore` segment documentation in `website/docs/segments/sitecore.mdx` to use `EndpointName` and `CmHost` fields and remove advanced xmcloud template and `.Cloud` and `.Environment` properties ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3926/files?diff=unified&w=0#diff-3af710d4f9882cc603875c32893b1e701198ef8218f9be38d2a4ffb3a1348819L36-R44))

### DIFF
--- a/src/segments/sitecore.go
+++ b/src/segments/sitecore.go
@@ -2,39 +2,64 @@ package segments
 
 import (
 	"encoding/json"
-	"path/filepath"
-	"strings"
+	"path"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/platform"
 	"github.com/jandedobbeleer/oh-my-posh/src/properties"
 )
 
-type Sitecore struct {
-	props properties.Properties
-	env   platform.Environment
+const (
+	sitecoreFileName = "sitecore.json"
+	sitecoreFolderName = ".sitecore"
+	userFileName = "user.json"
+	defaultEnpointName = "default"
+)
 
-	Environment string
-	Cloud       bool
+type Sitecore struct {
+    props   	properties.Properties
+    env     	platform.Environment
+
+    EndpointName 	string
+	CmHost 			string
 }
 
-type SitecoreConfig struct {
-	Endpoints struct {
-		Default struct {
-			Ref        string `json:"ref"`
-			AllowWrite bool   `json:"allowWrite"`
-			Host       string `json:"host"`
-			Variables  struct {
-			} `json:"variables"`
-		} `json:"default"`
-	} `json:"endpoints"`
+type EndpointConfig struct {
+	Host string `json:"host"`
+}
+
+type UserConfig struct {
+	DefaultEndpoint string 						`json:"defaultEndpoint"`
+	Endpoints 		map[string]EndpointConfig 	`json:"endpoints"`
 }
 
 func (s *Sitecore) Enabled() bool {
-	return s.shouldDisplay()
+	if s.env.HasFiles(sitecoreFileName) && s.env.HasFiles(path.Join(sitecoreFolderName, userFileName)) {
+		var userConfig = getUserConfig(s)
+
+		if (userConfig == nil) {
+			return false
+		}		
+
+		s.EndpointName = userConfig.getDefaultEndpoint()
+		
+		displayDefault := s.props.GetBool(properties.DisplayDefault, true)
+        
+		if !displayDefault && s.EndpointName == defaultEnpointName {
+			return false
+		} 
+
+		if endpoint := userConfig.getEndpoint(s.EndpointName); endpoint != nil && len(endpoint.Host) > 0 {
+			s.CmHost = endpoint.Host
+		}
+		
+		return true
+	}
+
+	return false
 }
 
 func (s *Sitecore) Template() string {
-	return " {{ if .Cloud }}\uf65e{{ else }}\uf98a{{ end }} {{ .Environment }} "
+    return "{{ .EndpointName }} {{ if .CmHost }}({{ .CmHost }}){{ end }}"
 }
 
 func (s *Sitecore) Init(props properties.Properties, env platform.Environment) {
@@ -42,37 +67,31 @@ func (s *Sitecore) Init(props properties.Properties, env platform.Environment) {
 	s.env = env
 }
 
-func (s *Sitecore) shouldDisplay() bool {
-	sitecoreDir, err := s.env.HasParentFilePath(".sitecore")
-	if err != nil {
-		return false
+func getUserConfig(s *Sitecore) *UserConfig {
+	userJson := s.env.FileContent(path.Join(sitecoreFolderName, userFileName))
+	var userConfig UserConfig
+
+	if err := json.Unmarshal([]byte(userJson), &userConfig); err != nil {
+		return nil
 	}
 
-	if !sitecoreDir.IsDir {
-		return false
+	return &userConfig
+}
+
+func (u *UserConfig) getDefaultEndpoint() string {
+	if len(u.DefaultEndpoint) > 0  {
+		return u.DefaultEndpoint
+	} 
+	
+	return defaultEnpointName
+}
+
+func (u *UserConfig) getEndpoint(name string) *EndpointConfig {
+	endpoint, exists := u.Endpoints[name]
+
+	if exists {
+		return &endpoint
 	}
 
-	if !s.env.HasFilesInDir(sitecoreDir.Path, "user.json") {
-		return false
-	}
-
-	sitecoreConfigFile := filepath.Join(sitecoreDir.Path, "user.json")
-
-	if len(sitecoreConfigFile) == 0 {
-		return false
-	}
-
-	content := s.env.FileContent(sitecoreConfigFile)
-
-	var config SitecoreConfig
-	if err := json.Unmarshal([]byte(content), &config); err != nil {
-		return false
-	}
-
-	// sitecore xm cloud always has sitecorecloud.io as domain
-	s.Cloud = strings.Contains(config.Endpoints.Default.Host, "sitecorecloud.io")
-
-	s.Environment = config.Endpoints.Default.Host
-
-	return true
+	return nil
 }
--- a/src/segments/sitecore.go
+++ b/src/segments/sitecore.go
@@ -33,29 +33,29 @@ type UserConfig struct {
 }
 
 func (s *Sitecore) Enabled() bool {
-	if s.env.HasFiles(sitecoreFileName) && s.env.HasFiles(path.Join(sitecoreFolderName, userFileName)) {
-		var userConfig = getUserConfig(s)
-
-		if (userConfig == nil) {
-			return false
-		}		
-
-		s.EndpointName = userConfig.getDefaultEndpoint()
-		
-		displayDefault := s.props.GetBool(properties.DisplayDefault, true)
-        
-		if !displayDefault && s.EndpointName == defaultEnpointName {
-			return false
-		} 
-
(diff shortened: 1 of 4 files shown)


src/segments/sitecore.go
```diff
@@ -2,77 +2,96 @@ package segments
...
-		} `json:"default"`
-	} `json:"endpoints"`
+type UserConfig struct {
+	DefaultEndpoint string 						`json:"defaultEndpoint"`
+	Endpoints 		map[string]EndpointConfig 	`json:"endpoints"`
 }
 
 func (s *Sitecore) Enabled() bool {
-	return s.shouldDisplay()
+	if s.env.HasFiles(sitecoreFileName) && s.env.HasFiles(path.Join(sitecoreFolderName, userFileName)) {
```
JanDeDobbeleer: this can be inverted so we do not indent the full block. Return false early and continue on the same level below.

src/segments/sitecore.go
```diff
@@ -2,77 +2,96 @@ package segments
...
-	props properties.Properties
-	env   platform.Environment
+    props   	properties.Properties
+    env     	platform.Environment
+
+    EndpointName 	string
+	CmHost 			string
+}
 
-	Environment string
```
JanDeDobbeleer, line 16: This is a breaking change, why was this done?

JanDeDobbeleer (changes requested):
@BasLijten this is a complete rewrite and as it's your baby I would love your review here too. I'm not comfortable with sitecore to understand the implications of this. I also don't understand why the current properties wouldn't be valid anymore .

src/segments/sitecore.go
dorokhin-bohdan: Thanks!
Fixed

dorokhin-bohdan:
Hello @JanDeDobbeleer @BasLijten 
First of all thanks for spending your time on review PR.

The Sitecore CLI 5.2+ provides a new ability to change the default environment. More information about the new command [here](https://doc.sitecore.com/xp/en/developers/103/developer-tools/the-environment-command.html).
The segment provided by @BasLijten works only with the default environment. 
The new segment implementation extends it and provides the ability to work with context-switching commands.

According to the new implementation authors of themes can show not only the host but also the environment name. 
Possible theme and example:
image

src/segments/sitecore.go
dorokhin-bohdan: According to the Sitecore CLI terminology Environment is a complex object which contains cm-host, auth-host, ref, etc.
According to the current implementation, Environment has a value of cm-host.

The new implementation provides the same value but used for it a new property called CmHost and extends the Sitecore struct by providing a new property called EndpointName which contains the name of the environment configuration.

It was done to align with Sitecore CLI terminology.

BasLijten:
Hi, @dorokhin-bohdan , first of all: cool that you added extra functionality, good to know that I am not the only person in the world that wanted to use this segment. I wasn't aware that this was added in a new version of the CLI: I might have completely missed it, or it wasn't available (yet) at the time of writing, as I was planning on making a context switcher, great the Sitecore introduced this. So I can find myself completely in the functionality, that you added. I do have a few comments/questions. It's based on reading the code and not actual testing, as I am not in the possibility to do this at this very moment.

- my _assumption_ is that it is backwards compatible: when the defaultEndpoint entry is missing, it always defaults back to "default"
- I'd love to keep the IsCloud boolean, to determine if you are working locally, xmcloud, or something else. If you have a suggestion on how to make this work, i'd love to hear that from you, based on complexity I could agree ;)

dorokhin-bohdan:
@BasLijten 
> my assumption is that it is backwards compatible: when the defaultEndpoint entry is missing, it always defaults back to "default"

This segment will work for previous versions of CLI (without context switching command) and return "default" name in that case.
For the new version you can easily disable showing the default endpoint in the segment by passing "false" value to the "display_default" properties. In that case segment will be displayed if your current env name is not default.
   
> I'd love to keep the IsCloud boolean, to determine if you are working locally, xmcloud, or something else. If you have a suggestion on how to make this work, i'd love to hear that from you, based on complexity I could agree ;)

Did you consider an option to pass this logic to the theme's template?
Template example: {{ if contains \"sitecorecloud.io\" .CmHost }}\uf65e{{ else }}\uf98a{{ end }}
image

The main concern regarding to .IsCloud property is a hardcoded part of the hostname in the segment code.
In case of changing xmcloud hostname, the segment should be updated and users will wait for a fix/update.

src/segments/sitecore.go
```diff
@@ -2,77 +2,96 @@ package segments
...
 	s.props = props
 	s.env = env
 }
 
-func (s *Sitecore) shouldDisplay() bool {
-	sitecoreDir, err := s.env.HasParentFilePath(".sitecore")
-	if err != nil {
-		return false
-	}
+func getUserConfig(s *Sitecore) *UserConfig {
```
JanDeDobbeleer: can we have this return an error on issues?

```suggestion
func getUserConfig(s *Sitecore) (*UserConfig, error) {
```

JanDeDobbeleer (changes requested):
If @BasLijten is OK with the response (I agree with the proposal to leverage template logic for cloud), I only have one remaining comment in my final review.

src/segments/sitecore.go
dorokhin-bohdan: yes, we can, but let me clarify one thing
is it necessary just to display an error message instead of hiding the whole segment? Or there is another aim of it?

example:
image

dorokhin-bohdan: After update:
the method getUserConfig return error and this value is used for the if statement instead of userConfig value.

src/segments/sitecore.go
```diff
@@ -2,77 +2,96 @@ package segments
...
 
 func (s *Sitecore) Enabled() bool {
-	return s.shouldDisplay()
+	if !s.env.HasFiles(sitecoreFileName) || !s.env.HasFiles(path.Join(sitecoreFolderName, userFileName)) {
+		return false
+	}
+
+	var userConfig, err = getUserConfig(s)
+
+	if (err != nil) {
```
JanDeDobbeleer: the brackets aren't needed here so they can be removed

JanDeDobbeleer (changes requested):

src/segments/sitecore.go
dorokhin-bohdan: habits from dotnet

Thanks

JanDeDobbeleer (approved):

src/segments/sitecore.go
JanDeDobbeleer: > the method getUserConfig return error and this value is used for the if statement instead of userConfig value

@dorokhin-bohdan that's indeed was I was after 👍🏻

JanDeDobbeleer:
@BasLijten just need you blessing/feedback

BasLijten:
> @BasLijten
> 
> > my assumption is that it is backwards compatible: when the defaultEndpoint entry is missing, it always defaults back to "default"
> 
> This segment will work for previous versions of CLI (without context switching command) and return "default" name in that case. For the new version you can easily disable showing the default endpoint in the segment by passing "false" value to the "display_default" properties. In that case segment will be displayed if your current env name is not default.
> 
> > I'd love to keep the IsCloud boolean, to determine if you are working locally, xmcloud, or something else. If you have a suggestion on how to make this work, i'd love to hear that from you, based on complexity I could agree ;)
> 
> Did you consider an option to pass this logic to the theme's template? Template example: {{ if contains "sitecorecloud.io" .CmHost }}\uf65e{{ else }}\uf98a{{ end }} image
> 
> The main concern regarding to .IsCloud property is a hardcoded part of the hostname in the segment code. In case of changing xmcloud hostname, the segment should be updated and users will wait for a fix/update. 

I agree on the hardcoded part. Was planning to remove that in a future update to a property. Although I am not to fond of moving a lot of logic into the template, I can totally live with it. Thanks for the effort in making this a (much) better segment! @JanDeDobbeleer I'm not sure if you have the same saying in the Belgium, but I'll give it a try: gaan met die banaan!

BasLijten:
> @BasLijten just need you blessing/feedback

I dont have any remarks anymore

JanDeDobbeleer:
> gaan met die banaan!

@BasLijten can do 😂

JanDeDobbeleer:
@dorokhin-bohdan test should succeed, I can fix the commit message errors by squashing, that's no issue.

JanDeDobbeleer:
@all-contributors please add @dorokhin-bohdan for code,design,doc

This is pure magic 🪄! @holopin-bot @dorokhin-bohdan wizard

PR #3990, opened 2023-06-23
Change the Chips Theme .Path to agnoster_short with 2 levels

junhan-z:
### Prerequisites

### Description
I really like the chips theme, however, I think the path part can be improved - to include 2 levels instead of just current directory, which provides much more information and not too long meanwhile.

**As I'm not the original author of this theme, I'm not sure if this PR is appropriate**. On the other hand, I don't want to create yet another theme that is highly similar with this theme.

Any suggestions are welcome, thanks.

### <samp>🤖 Generated by Copilot at 9ea227c</samp>

Adjusted the `leading_diamond` segment style and folder depth in the `chips` theme. This improves the prompt appearance and usability.

### <samp>🤖 Generated by Copilot at 9ea227c</samp>

*  Modify the `properties` of the `leading_diamond` segment in the `chips` theme to use the `agnoster_short` style and limit the folder depth to 2 ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/3990/files?diff=unified&w=0#diff-d6f44ea85e61865b70c07cc65a2577a45dad4d68d3f9f421cd96e515111b6b49L16-R17)). This change makes the prompt more compact and readable when navigating deep directories in the `themes/chips.omp.json` file.

### DIFF
--- a/themes/chips.omp.json
+++ b/themes/chips.omp.json
@@ -13,7 +13,8 @@
           ],
           "leading_diamond": "\uE0B6",
           "properties": {
-            "style": "folder"
+            "style": "agnoster_short",
+            "max_depth": 2
           },
           "style": "diamond",
           "template": "\uF07B {{ .Path }}",


JanDeDobbeleer:
@CodexLink care to review this one?

CodexLink:
Sure, thanks for letting me know @JanDeDobbeleer.

@junhan-z Thanks for the suggestion.

I agree that it could have been more minimally detailed than what it had. The current one sometimes made me do `pwd` because I usually forgot where was I.

Anyway, the suggestion is good but I wanna add some additional property which I will be reviewing the code to add that though. The reason for that is because the **drive letter is visually cluttering for me**, or the suggestion **does not _quite_ fit within the aesthetic** of the theme in terms of displaying such contexts. _(path being too long along with git, may potentially eradicate other badges on the right)_

Further context, the initial goal of the theme was to basically have these contexts displayed without visually cluttering itself and one another by displaying exactly what they are supposed to (strictly) represent; hence the main reason why I set the `Path` to display `Folder`; but since I understand users' confusion in regards to what `Path` is their terminal on, this suggestion LGTM. 

Anyway, the property I wanted to add is just a touch-up if users really wanted a driver letter in the `Path` or not, which for me I prefer not having, but users will have the choice now. See the output below.

## With `hide_root_location` as `false`
cKCGp9zXSE

## With `hide_root_location` as `true`
nSU03MYOnb

### Extra

@JanDeDobbeleer may I ask if it would still be possible to create variants of the theme, I may plan on creating `double-width` _(this was for compatibility issues with fonts like Iosevka Term wherein single space truncates the icons)_ and `minimal` _(is for people who wanted lesser visual clutter, few icons, more strict context)_ variants just in case people have an issue with the current theme and prefer just having the other. I remembered that it is not allowed to post such themes because we have too many themes, but I'm not quite sure if variants are part of the restriction as they are also theme with few adjustments.

Let me know if there are concerns with the idea, as well as further suggestions and recommendations.

junhan-z:
I see, thank you!
OK, since it's partially your plan, I'm looking forward to the future changes from you and closing this PR.

Thank you for both of your time @JanDeDobbeleer @CodexLink

PR #4108, opened 2023-08-01
Added linux section in customize.mdx

Melyns:
I added a linux section with copy pastable command for linux users, for example ubuntu users using bash. I was confused at first with the theme install process since i use ubuntu. So i figured to update the docs & make it easier for others.

### DIFF
--- a/website/docs/installation/customize.mdx
+++ b/website/docs/installation/customize.mdx
@@ -39,6 +39,7 @@ following way to reference them directly. This will keep them up-to-date and com
   values={[
     { label: 'windows', value: 'windows', },
     { label: 'homebrew', value: 'homebrew', },
+    { label: 'linux', value: 'linux', },
   ]
 }>
 <TabItem value="windows">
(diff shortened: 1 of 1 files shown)


website/docs/installation/customize.mdx
```diff
@@ -49,13 +50,22 @@ For example, to use the [jandedobbeleer][jandedobbeleer] theme, alter the init l
 oh-my-posh init pwsh --config "$env:POSH_THEMES_PATH/jandedobbeleer.omp.json" | Invoke-Expression
 ```
 
+</TabItem>
+<TabItem value="linux">
+
+To use the [jandedobbeleer][jandedobbeleer] theme, alter the init line like this (`bash`):
+
+```bash
+eval "$(oh-my-posh init bash --config POSH_THEMES_PATH/jandedobbeleer.omp.json)"
```
JanDeDobbeleer, line 59: @Melyyn the linux installation doesn't create the `POSH_THEMES_PATH` environment variable so I think this will confuse people even more.

Melyns: Understood!

PR #4352, opened 2023-10-16
feat(NBA): Add a new NBA segment to track schedule + scores for your favorite NBA team

joadoumie:
Creating a draft PR as I have not yet implemented the tests for the change. I still have to think through the best way to actually do the testing because the JSON that is retrieved from the endpoint is very large, so I don't necessarily want to replicate the entire JSON... I may have to though.
 
### Prerequisites

### <samp>🤖 Generated by Copilot at e93706e</samp>

This pull request adds a new NBA segment for oh-my-posh, which shows NBA game data in the prompt. It implements the segment logic in `nba.go`, the segment type and creation in `segment.go`, and the segment documentation in `nba.mdx`. It also uses the NBA API and the Docusaurus framework as dependencies.

### <samp>🤖 Generated by Copilot at e93706e</samp>

*  Add a new segment type `NBA` to display NBA game data ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/4352/files?diff=unified&w=0#diff-cd158856c1b1aa6b2acece8041eb07352e4247b700e0fc62798a6be35d5f1c7dR171-R172), [link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/4352/files?diff=unified&w=0#diff-cd158856c1b1aa6b2acece8041eb07352e4247b700e0fc62798a6be35d5f1c7dR296))
*  Implement the `NBA` struct and its methods in `nba.go` to fetch, parse, and render the NBA API response ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/4352/files?diff=unified&w=0#diff-34cae96d58fdd9d69f24bdec1be456f7cd877135dceb025fdc13918b54255666R1-R257))
*  Document the NBA segment and its configuration, properties, and template in `nba.mdx` ([link](https://github.com/JanDeDobbeleer/oh-my-posh/pull/4352/files?diff=unified&w=0#diff-9e6a8c778e3ef0f52088c4a2f1ce7b56ad0f6b6b1acac5e0ccb67ac3446cf941R1-R99))

### DIFF
--- a/src/segments/nba.go
+++ b/src/segments/nba.go
@@ -1,0 +1,418 @@
+package segments
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+)
+
+// segment struct, makes templating easier
+type Nba struct {
+	props properties.Properties
+	env   platform.Environment
+
+	NBAData
+}
+
+// NBA struct contains parsed API data that care about for the segment
+type NBAData struct {
+	HomeTeam       string
+	AwayTeam       string
+	Time           string
+	GameDate       string
+	StartTimeUTC   string
+	GameStatus     GameStatus // 1 = scheduled, 2 = in progress, 3 = finished
+	HomeScore      int
+	AwayScore      int
+	HomeTeamWins   int
+	HomeTeamLosses int
+	AwayTeamWins   int
+	AwayTeamLosses int
+}
+
+func (nba *NBAData) HasStats() bool {
+	return nba.HomeTeamWins != 0 || nba.HomeTeamLosses != 0 || nba.AwayTeamWins != 0 || nba.AwayTeamLosses != 0
+}
+
+func (nba *NBAData) Started() bool {
+	return nba.GameStatus == InProgress || nba.GameStatus == Finished
+}
+
+const (
+	NBASeason  properties.Property = "season"
+	TeamName   properties.Property = "team"
+	DaysOffset properties.Property = "days_offset"
+
+	ScheduledTemplate  properties.Property = "scheduled_template"
+	InProgressTemplate properties.Property = "in_progress_template"
+	FinishedTemplate   properties.Property = "finished_template"
+
+	NBAScoreURL      string = "https://cdn.nba.com/static/json/liveData/scoreboard/todaysScoreboard_00.json"
+	NBASchedURLPart1 string = "https://stats.nba.com/stats/internationalbroadcasterschedule?LeagueID=00&Season="
+	NBASchedURLPart2 string = "&RegionID=1&EST=Y"
+
+	Unknown = "Unknown"
+)
+
+// Custom type for GameStatus
+type GameStatus int
+
+// Constants for GameStatus values
+const (
+	Scheduled  GameStatus = 1
+	InProgress GameStatus = 2
+	Finished   GameStatus = 3
+	NotFound   GameStatus = 4
+)
+
+// Int() method for GameStatus to get its integer representation
+// This is a helpful method if people want to come up with their own templates
+func (gs GameStatus) Int() int {
+	return int(gs)
+}
+
+func (gs GameStatus) Valid() bool {
+	return gs == Scheduled || gs == InProgress || gs == Finished
+}
+
+func (gs GameStatus) String() string {
+	switch gs {
+	case Scheduled:
+		return "Scheduled"
+	case InProgress:
+		return "In Progress"
+	case Finished:
+		return "Finished"
+	case NotFound:
+		return "Not Found"
+	default:
+		return Unknown
+	}
+}
+
+// All of the structs needed to retrieve data from the live score endpoint
+type ScoreboardResponse struct {
+	Scoreboard Scoreboard `json:"scoreboard"`
+}
+
+type Scoreboard struct {
+	GameDate string `json:"gameDate"`
+	Games    []Game `json:"games"`
+}
+
+type Game struct {
+	GameStatus     int    `json:"gameStatus"`
+	GameStatusText string `json:"gameStatusText"`
+	GameTimeUTC    string `json:"gameTimeUTC"`
+	HomeTeam       Team   `json:"homeTeam"`
+	AwayTeam       Team   `json:"awayTeam"`
+}
+
+type Team struct {
+	TeamTricode string `json:"teamTricode"`
+	Wins        int    `json:"wins"`
+	Losses      int    `json:"losses"`
+	Score       int    `json:"score"`
+}
+
+// All the structs needed to get data from the schedule endpoint
+type ScheduleResponse struct {
+	ResultSets []ResultSet `json:"resultSets"`
+}
+
+type ResultSet struct {
+	CompleteGameList []ScheduledGame `json:"CompleteGameList,omitempty"`
+}
+
+type ScheduledGame struct {
+	VtAbbreviation string `json:"vtAbbreviation"`
+	HtAbbreviation string `json:"htAbbreviation"`
+	Date           string `json:"date"`
+	Time           string `json:"time"`
+}
+
+func (nba *Nba) Template() string {
+	return " \U000F0806 {{ .HomeTeam}}{{ if .HasStats }} ({{.HomeTeamWins}}-{{.HomeTeamLosses}}){{ end }}{{ if .Started }}:{{.HomeScore}}{{ end }} vs {{ .AwayTeam}}{{ if .HasStats }} ({{.AwayTeamWins}}-{{.AwayTeamLosses}}){{ end }}{{ if .Started }}:{{.AwayScore}}{{ end }} | {{ if not .Started }}{{.GameDate}} | {{ end }}{{.Time}} " //nolint:lll
+}
+
+func (nba *Nba) Enabled() bool {
+	data, err := nba.getResult()
+	if err != nil || !data.GameStatus.Valid() {
+		return false
+	}
+
+	nba.NBAData = *data
+
+	return true
+}
+
+// Returns an empty Game Data struct with the GameStatus set to NotFound
+// Helpful for caching the fact that a game was not found for a team
+func (nba *Nba) getGameNotFoundData() string {
+	return `{
+		"HomeTeam":"",
+		"AwayTeam":"",
+		"Time":"",
+		"GameDate":"",
+		"StartTimeUTC":"",
+		"GameStatus":4,
+		"HomeScore":0,
+		"AwayScore":0,
+		"HomeTeamWins":0,
+		"HomeTeamLosses":0,
+		"AwayTeamWins":0,
+		"AwayTeamLosses":0
+	}`
+}
+
+// parses through a set of games from the score endpoint and looks for props.team in away or home team
+func (nba *Nba) findGameScoreByTeamTricode(games []Game, teamTricode string) (*Game, error) {
+	for _, game := range games {
+		if game.HomeTeam.TeamTricode == teamTricode || game.AwayTeam.TeamTricode == teamTricode {
+			return &game, nil
+		}
+	}
+
+	return nil, errors.New("no game score found for team")
+}
+
+// parses through a set of games from the schedule endpoint and looks for props.team in away or home team
+func (nba *Nba) findGameSchedulebyTeamTricode(games []ScheduledGame, teamTricode string) (*ScheduledGame, error) {
+	for _, game := range games {
+		if game.VtAbbreviation == teamTricode || game.HtAbbreviation == teamTricode {
+			return &game, nil
+		}
+	}
+
+	return nil, errors.New("no scheduled game found for team")
+}
+
+// parses the time and date from the schedule endpoint into a UTC time
+func (nba *Nba) parseTimetoUTC(timeEST, date string) string {
+	combinedTime := date + " " + timeEST
+	timeUTC, err := time.Parse("01/02/2006 03:04 PM", combinedTime)
+	if err != nil {
+		return ""
+	}
+
+	return timeUTC.UTC().Format("2006-01-02T15:04:05Z")
+}
+
+// retrieves data from the score endpoint
+func (nba *Nba) retrieveScoreData(teamName string, httpTimeout int) (*NBAData, error) {
+	body, err := nba.env.HTTPRequest(NBAScoreURL, nil, httpTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	var scoreboardResponse *ScoreboardResponse
+	err = json.Unmarshal(body, &scoreboardResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	gameInfo, err := nba.findGameScoreByTeamTricode(scoreboardResponse.Scoreboard.Games, teamName)
+	if err != nil {
+		return nil, err
+	}
+
+	return &NBAData{
+		AwayTeam:       gameInfo.AwayTeam.TeamTricode,
+		HomeTeam:       gameInfo.HomeTeam.TeamTricode,
+		Time:           gameInfo.GameStatusText,
+		GameDate:       scoreboardResponse.Scoreboard.GameDate,
+		StartTimeUTC:   gameInfo.GameTimeUTC,
+		GameStatus:     GameStatus(gameInfo.GameStatus),
+		HomeScore:      gameInfo.HomeTeam.Score,
+		AwayScore:      gameInfo.AwayTeam.Score,
+		HomeTeamWins:   gameInfo.HomeTeam.Wins,
+		HomeTeamLosses: gameInfo.HomeTeam.Losses,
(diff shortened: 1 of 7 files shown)


src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// NightscoutData struct contains the API data
+type NBAData struct {
+	HomeTeam       string
+	AwayTeam       string
+	Time           string
+	GameDate       string
+	StartTimeUTC   string
+	GameStatus     int // 1 = scheduled, 2 = in progress, 3 = finished
```
JanDeDobbeleer: If you make this into a custom type, you can add a `String()` function to get the representation.

src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+	// parses through a set of games and looks for props.team in away or home team
+	findGameByTeamTricode := func(games []Game, teamTricode string) (*Game, error) {
+		for _, game := range games {
+			if game.HomeTeam.TeamTricode == teamTricode || game.AwayTeam.TeamTricode == teamTricode {
+				return &game, nil
+			}
+		}
+		return nil, errors.New("no game found for team")
+	}
+	getCacheValue := func(key string) (*NBAData, error) {
```
JanDeDobbeleer: add a space here for readability

src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+			err := json.Unmarshal([]byte(val), &nbaData)
+			if err != nil {
+				return nil, err
+			}
+			return nbaData, nil
+		}
+		return nil, errors.New("no data in cache")
+	}
+
+	cachedScheduleKey := nba.props.GetString(TeamName, "") + "schedule"
```
JanDeDobbeleer: team name is fetched twice, I would cache it above and use a `fmt.Sprintf("%sschedule", team)` below for both keys.

src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+	// How often you want to query the API to get live score information, defaults to 2 minutes
+	cacheScoreTimeout := nba.props.GetInt(properties.CacheTimeout, 2)
+
+	// Cache the schedule information for a day so we don't call the API too often
+	cacheScheduleTimeout := nba.props.GetInt(properties.CacheTimeout, 14400)
+
+	teamName := nba.props.GetString(TeamName, "")
+
+	if cacheScheduleTimeout > 0 {
+		if data, err := getCacheValue(cachedScheduleKey); err == nil {
```
JanDeDobbeleer: Make this a separate function so we can improve readability.

src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+
+	if cacheScoreTimeout > 0 {
+		if data, err := getCacheValue(cachedScoreKey); err == nil {
+			return data, nil
+		}
+	}
+
+	body, err := nba.env.HTTPRequest(NBAScoreURL, nil, httpTimeout)
+	if err != nil {
+		// use oh my posh debugger
```
JanDeDobbeleer: useless comment :-)

src/segments/nba.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, err
+	}
+
+	gameInfo, err := findGameByTeamTricode(scoreboardResponse.Scoreboard.Games, teamName)
+	if err != nil {
+		nba.env.Debug(fmt.Sprintf("Error: %s", err))
+		return nil, err
+	}
+
+	data := NBAData{}
```
JanDeDobbeleer: any reason to use a second struct and not embed the `scoreboardResponse` in the NBA struct?

JanDeDobbeleer (changes requested):
a few comments, about the tests, you don't need every field as that's not really useful. You can return a simple json that parses (or not) to validate the logic.

src/segments/nba.go
joadoumie: Thanks for all the thoughtful feedback so far! I'll work on tweaking the other pieces based on your feedback :)

The reason for this was that the scoreboardResponse itself is quite large and can contain in theory (if every team is playing) 15 different game responses. Then in order to display the relevant information for the **specific team** you are hoping to follow you'd need to pull out the information for the specific team you are hoping to get schedule + stats for. Then inside of the NBA struct you'd have so much information for teams that are not relevant. There'd be a ton of extra info here and that to me felt a bit overwhelming to manage.

So the thought here was to parse the scoreboard response and then just cache the relevant information that we need for the team we are trying to get info for. I thought (not familiar with Go perf + considerations) this would be better for caching + performance. Two primary reasons:

1. If we store the whole scoreboardResponse in the cache we'd have a big chunk of data in the cache that we need to parse each time to get to the proper team info and most of that data may be irrelevant. We'd have to know where in the Games[] array the particular game is that we are trying to get information for.

2. If people want to create their own templates, they'd have a tough time because they'd have to know exactly where the data is for their team inside of the scoreboardResponse. With the NBAData struct it feels easy to just get any of the most relevant information that you'd actually care about.

That said ^, I'm completely new to the project, so lmk if you think there is a better way and I'm happy to learn from you and make the necessary changes :-)

src/segments/nba.go
joadoumie: Which part should be a different function? The conditional or the time check?

joadoumie:
I am new to contributing - I made a wide range of changes based on your feedback @JanDeDobbeleer. Thanks again for all your help here. I am very excited to get this thing out the door!

I am concurrently working on adding the nba_test.go functionality :)

nopeless:
What a feature contribution. I am amazed to see such unique ideas really (never imagined id see nba on this).

What motivated you to write this? Just curious

src/segments/nba.go
JanDeDobbeleer: the entire block in the conditional

src/segments/nba.go
```diff
@@ -0,0 +1,511 @@
...
+		HomeScore:      0,
+		AwayScore:      0,
+		HomeTeamWins:   0,
+		HomeTeamLosses: 0,
+		AwayTeamWins:   0,
+		AwayTeamLosses: 0,
+	}
+	// get string representation of res, nba data
+	res, _ := json.Marshal(gameNotFound)
+	return string(res)
```
JanDeDobbeleer: this can be a string constant instead of a runtime parsing of an empty struct, no?

src/segments/nba.go
```diff
@@ -0,0 +1,511 @@
...
+	// parses through a set of games from the score endpoint and looks for props.team in away or home team
+	findGameScoreByTeamTricode := func(games []Game, teamTricode string) (*Game, error) {
+		for _, game := range games {
+			if game.HomeTeam.TeamTricode == teamTricode || game.AwayTeam.TeamTricode == teamTricode {
+				return &game, nil
+			}
+		}
+		return nil, errors.New("no game score found for team")
+	}
+	// parses through a set of games from the schedule endpoint and looks for props.team in away or home team
```
JanDeDobbeleer: this needs an additional space

src/segments/nba.go
JanDeDobbeleer: @jadoumie gotcha, but as we only care about what we will map towards the NBA struct, I would only create the relevant struct and properties instead of the entire JSON object. It's overhead that doesn't do anything except consume CPU cycles. So if this is what we need:

```go
data := NBAData{}
data.AwayTeam = gameInfo.AwayTeam.TeamTricode
data.HomeTeam = gameInfo.HomeTeam.TeamTricode
data.Time = gameInfo.GameStatusText
data.GameDate = scoreboardResponse.Scoreboard.GameDate
data.StartTimeUTC = gameInfo.GameTimeUTC
data.GameStatus = GameStatus(gameInfo.GameStatus)
data.HomeScore = gameInfo.HomeTeam.Score
data.AwayScore = gameInfo.AwayTeam.Score
data.HomeTeamWins = gameInfo.HomeTeam.Wins
data.HomeTeamLosses = gameInfo.HomeTeam.Losses
data.AwayTeamWins = gameInfo.AwayTeam.Wins
data.AwayTeamLosses = gameInfo.AwayTeam.Losses
```

Then this is all we need to parse:

```go
type Game struct {
	AwayTeam          Team        `json:"awayTeam"`
	HomeTeam          Team        `json:"homeTeam"`
	GameStatusText    string      `json:"gameStatusText"`
	GameTimeUTC       string      `json:"gameTimeUTC"`
	GameStatus        int         `json:"gameStatus"`
}

type Team struct {
	TeamTricode       string   `json:"teamTricode"`
	Score             int      `json:"score"`
	Wins              int      `json:"wins"`
	Losses            int      `json:"losses"`
}

type ScoreboardResponse struct {
	Scoreboard Scoreboard `json:"scoreboard"`
}

type Scoreboard struct {
	GameDate   string `json:"gameDate"`
}
```

JanDeDobbeleer (changes requested):
I would also, instead of having inline functions inside `getResult()`, expose those as internal functions of the `NBA` struct. There's a lof of context in there that makes the entire block hard to read. Extracting those into smaller context will make it a lot easier to follow what's going on.

joadoumie:
@nopeless I love basketball! And I was inspired by @JanDeDobbeleer and @shanselman's Nightscout segment: https://www.hanselman.com/blog/a-nightscout-segment-for-ohmyposh-shows-my-realtime-blood-sugar-readings-in-my-git-prompt

I figured the shell is a really fun place to be able to surface information I care about and was hoping I wasn't the only person who'd find this interesting :) 

image
image

src/segments/nba.go
joadoumie: Oh wow! I didn't understand how this worked in go. I thought we needed to have a complete struct so we could map the byte array from the http request... makes sense that this would be way better though!

I'll go ahead and make the changes here!

src/segments/nba.go
joadoumie: True! It felt nice to always cache the same type of object and to take advantage of the GameStatus struct, but sounds like it may not be worth the extra performance hit?

src/segments/nba.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+type ScheduledGame struct {
+	VtAbbreviation string `json:"vtAbbreviation"`
+	HtAbbreviation string `json:"htAbbreviation"`
+	Date           string `json:"date"`
+	Time           string `json:"time"`
+}
+
+func (nba *NBA) Template() string {
```
JanDeDobbeleer: We usually only have a single template. Now, I get that this could be a bit too challenging in this use-case so I'm down to going this direction. But, then we would have to allow people to override the templates below, and I'd rather have that logic in 1 function instead of having it split across multiple. This logic `if nba.HomeTeamWins == 0 && nba.HomeTeamLosses == 0 && nba.AwayTeamWins == 0 && nba.AwayTeamLosses == 0` can be a boolean function on the nba struct, but I would add every other template directly in the `Template` function AND, if the user has its own `template` property, we simply return that one first without running any additional logic.

src/segments/nba.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		// Check if we can find a game for the team
+		gameInfo, err := nba.findGameSchedulebyTeamTricode(scheduleResponse.ResultSets[1].CompleteGameList, teamName)
+		if err != nil {
+			// We didn't find a game for the team on this day, so we need to check the next day
+			t = t.AddDate(0, 0, 1)
+			numDaysToSearch--
+			continue
+		} else {
```
JanDeDobbeleer: This `else` is redundant. If we hit the first `if`, we will never need this `else` due to the `continue`, so we can simply remove that and "continue" on the same level below.

JanDeDobbeleer (changes requested):

src/segments/nba.go
joadoumie: Good catch :)

src/segments/nba.go
joadoumie: I'm a bit confused. I was under the impression that a user can still have whatever template they'd like to use and override the template even in the way it is currently constructed. For example, I'm finding that something like this:

```json
"segments": [
        {
          "background": "#e9ac2f",
          "foreground": "#8748dc",
          "leading_diamond": "\ue0b6",
          "style": "diamond",
	      "template": "I want to override the NBA semgent with {{ .HomeTeam }}"
          "trailing_diamond": "\ue0b0",
          "type": "nba",
          "properties": {
            "team":"LAL",
            "http_timeout": 1500,
	    "season": "2023",
	    "days_offset": 8
          }
        }
      ],
      "type": "prompt"
    }
  ],
```

Will render the template provided by the user "Test this out {{ .HomeTeam }}" and not the template that is being returned in Template(). 

As always, thanks for the help and taking the time to teach me what's going on here!

joadoumie: image

JanDeDobbeleer: You're right, it's defined in the engine to do that. But, I would still export the included partial templates for override so the user can decide to override one or more without having to resort to a full override.

joadoumie: Can you please explain how I should handle this? What do you mean by export the included partial templates?

joadoumie:
I have added test cases in nba_test.go. 

I still have to: 

1. Figure out how to best handle the Template situation with all of the different possible variations. 
2. Follow up on other refactoring as needed.

Hopefully getting really close to wrapping it up :)

JanDeDobbeleer:
@joadoumie I will take the last ones, still this week. You've done an amazing job already. I want to go through the code in vscode myself and play around to see what works best.

joadoumie:
> @joadoumie I will take the last ones, still this week. You've done an amazing job already. I want to go through the code in vscode myself and play around to see what works best.

@JanDeDobbeleer that sounds great! Thank you so much for pushing this through to the finish line. It was a really fun project to work on and I'd love to keep contributing to this project moving forward. I appreciate all of your help and guidance. I was able to learn a lot :).

JanDeDobbeleer:
@joadoumie I'm done. If you can validate my changes, we have a go.

joadoumie:
@JanDeDobbeleer looks great! Thanks for all the changes. Looking through the code gave me some good ideas for the next time I work on a project like this. I'm going to change this to ready for review so you can merge it. Ship it!

JanDeDobbeleer:
@joadoumie I forgot to extend the docs, will do that tomorrow. Will ship ASAP!

src/segments/nba.go
```diff
@@ -0,0 +1,425 @@
...
+
+// Gets the data from the cache for a scheduled game if it exists
+// Checks whether the game should have started and if so, removes the cache entry
+func (nba *Nba) getCachedScheduleValue(key string) (*NBAData, error) {
+	data, err := nba.getCacheValue(key)
+	if err != nil {
+		return nil, errors.New("no data in cache")
+	}
+
+	// check if the game was previously not found and we should wait to check again
```
joadoumie, line 331: @JanDeDobbeleer - I'm actually experiencing a minor bug after testing a fake team tri-code "DSF". 

I think with this addition to the code that we return an error in the case that we have a game that was previously not found we are continuing to hit the endpoints. I can't leave a code review on my own PR GitHub says, so I'm tagging you in this comment instead. 

The end result is what you'd expect, but since it tries to find a game every single time, it's pretty slow and doesn't take advantage of caching periodically for a game not found.

src/segments/nba.go
```diff
@@ -0,0 +1,425 @@
...
+	cacheScoreTimeout := nba.props.GetInt(properties.CacheTimeout, 2)
+
+	// Cache the schedule information for a day so we don't call the API too often
+	cacheScheduleTimeout := nba.props.GetInt(properties.CacheTimeout, 1440)
+
+	// Cache the fact a game was not found for 30 minutes so we don't call the API too often
+	cacheNotFoundTimeout := nba.props.GetInt(properties.CacheTimeout, 30)
+
+	if cacheScheduleTimeout > 0 {
+		if data, err := nba.getCachedScheduleValue(cachedScheduleKey); err == nil {
```
joadoumie, line 372: @JanDeDobbeleer - now when we check the getCachedScheduleValue, if we already have a game that was not found in the cache (so we should not check again) we will actually keep going with the rest of the code because err != nil

I have some ideas for how to fix it, but figured maybe the quickest way would be for you to resolve it. Otherwise, I'm more than happy to push some changes.

src/segments/nba.go
JanDeDobbeleer: I'll check to see what we can do about that.

src/segments/nba.go
JanDeDobbeleer: I'll check to see what we can do about that.

src/segments/nba.go
JanDeDobbeleer: This is solved now, it will only do the call again on the cache timeout and the context is logged.

src/segments/nba.go
JanDeDobbeleer: also fixed

JanDeDobbeleer:
@all-contributors please add @joadoumie for code,doc

This is pure magic 🪄! @holopin-bot @joadoumie wizard

JanDeDobbeleer:
@joadoumie if there's still an issue, let me know and we can fix forward. Seems to work as expected now.

PR #4400, opened 2023-10-29
feat(python): Use parent directory as name if venv name is "venv" or ".venv"

SriRamanujam:
### Prerequisites

### Description

It is somewhat common to create Python virtualenvs in the working directory of the Python project. Usually people simply create these virtualenvs with the name "venv" or ".venv". This is common enough that it's referenced in the official Python documentation: https://docs.python.org/3/library/venv.html

This change makes it so that if the base name of the determined virtualenv path is "venv" or ".venv", it will instead use the parent directory as the virtualenv's name.

ex. if your venv is in `/path/to/folder/.venv`, the value of `{{ .Venv }}` will be `folder` instead of `.venv`.

### How

Updated `src/segments/python.go` with a check in `loadContext()` to use the parent directory of the venv file if the initially calculated name is `venv` or `.venv`.
Updated python.mdx to note this behavior. 
Added a new test to `src/segments/python_test.go` for this new behavior.

### DIFF
--- a/src/segments/python.go
+++ b/src/segments/python.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/platform"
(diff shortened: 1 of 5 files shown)


src/segments/python.go
```diff
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"golang.org/x/exp/slices"
```
JanDeDobbeleer:
```suggestion
	"slices"
```

as of 1.21 slices is now part of go itself

src/segments/python.go
```diff
@@ -72,13 +74,23 @@ func (p *Python) loadContext() {
 		"CONDA_ENV_PATH",
 		"CONDA_DEFAULT_ENV",
 	}
+	defaultVenvNames := []string{
```
JanDeDobbeleer, line 77: can you add a space between this new block? Above and below, will improve readability

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@all-contributors please add @SriRamanujam for code,doc

This is pure magic 🪄! @holopin-bot @SriRamanujam wizard

PR #4421, opened 2023-11-01
Ignore TFM if not found

paulomorgado:
### Prerequisites

### Description

Visual Studio solution and solution filter files do not have a TFM (`Target`) and fail because of that.

### How

Defaulted to an empty string for the `Target` when not found.

I can add unit tests for this, but I'm no GO programmer and I haven't found any.

If someone points me at a similar feature, I can add the tests.

### DIFF
--- a/src/segments/project.go
+++ b/src/segments/project.go
@@ -3,7 +3,7 @@ package segments
 import (
 	"encoding/json"
 	"encoding/xml"
-	"errors"
+	"fmt"
 	"path/filepath"
 	"strings"
 
@@ -208,13 +208,14 @@ func (n *Project) getNuSpecPackage(_ ProjectItem) *ProjectData {
 func (n *Project) getDotnetProject(_ ProjectItem) *ProjectData {
 	var name string
 	var content string
+	var extension string
 
 	extensions := []string{".sln", ".slnf", ".csproj", ".fsproj", ".vbproj"}
 	files := n.env.LsDir(n.env.Pwd())
 
 	// get the first match only
 	for _, file := range files {
-		extension := filepath.Ext(file.Name())
+		extension = filepath.Ext(file.Name())
 		if slices.Contains(extensions, extension) {
 			name = strings.TrimSuffix(file.Name(), filepath.Ext(file.Name()))
 			content = n.env.FileContent(file.Name())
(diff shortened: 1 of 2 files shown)


src/segments/project.go
```diff
@@ -226,11 +225,10 @@ func (n *Project) getDotnetProject(_ ProjectItem) *ProjectData {
 	// so instead of xml.Unmarshal() we use regex:
 	tag := "(?P<TAG><.*TargetFramework.*>(?P<TFM>.*)</.*TargetFramework.*>)"
 	values := regex.FindNamedRegexMatch(tag, content)
-	if len(values) == 0 {
-		n.Error = errors.New("cannot extract TFM from " + name + " project file").Error()
-		return nil
+	target := ""
+	if len(values) != 0 {
+		target = values["TFM"]
```
JanDeDobbeleer, line 233: can you also log a debug statement here?

paulomorgado: I can do that, but it's not an error for a solution file to not have that pattern, or even for a project. I frequently set the target frameworks in a Directory.Build.props and have it applied to all projects.

JanDeDobbeleer: I know, but it's a good idea to log this in case people create an issue because they expect this somehow (even for `csproj` etc).

paulomorgado: How do I log a debug statement?

On my working copy here I'm working on a different logic (the correct one :smile:). If there's a `.sln` file, that's the one; otherwise, if there's a `.slnf` file, that's the one; otherwise, it's the first project found.

JanDeDobbeleer: @paulomorgado but you never have a `.sln` file next to a project file in the same directory, right?

paulomorgado: A solution is a directory of projects and other files with the folder structure of how it should be seen on Visual Studio's Solution Explorer.

A solution filter is a manifest that points to the solution file and lists all the projects from that solution that should be loaded. If you browse https://github.com/dotnet/aspnetcore/, you'll see one single **.sln** file and lots of **.slnf** files.

Usually, each project (**.csproj**, **.fsproj**, **.vbproj**) is in its own folder and solution or solution filters are in parent directories.

But that is not required. In fact, Visual Studio requires a solution file to open .NET projects. If you open a project, it creates a solution with the same name as the project in the same folder.

As for the target, building a **C#**, **F#**, **Visual Basic** using **MSBuild** includes a lookup strategy of **MSBuild** files that does not require the project file itself to declare the **TargetFramework** or **TargetFrameworks**. It's wrong to assume that if it's not there, there's an error.

So, the correct lookup stratagy should be:

* If any **.sln** file is found: that's it.
  * If any **.slnf** file is found: that's it.
    * If any **.csproj**, **.fsproj** or **.vbproj** is found: that's
      * It's a nice to have reporting the target, if found, but not an error to not have it.
 
I can implement this logic.

JanDeDobbeleer: But in my experience, `sln` and `slnf` are found in the root of the project, and `csproj` etc in subfolders. So all you need to in that case is not return an error on no version found. A log statement is fine. There's no need to follow that logical path in case of no collisions.

paulomorgado: How do I add the log statement?

JanDeDobbeleer: The `env` property has a function called `Error` to log this.

paulomorgado: Done! But it's an error to consider an error a TFM not being present.

Also added the extension as `Type`. It's useful in templates.

JanDeDobbeleer: @paulomorgado but type needs to stay what it is (`dotnet`). Otherwise you have a breaking change in behaviour. It's overridden anyways elsewhere, that's also why it isn't set here.

PR #4460, opened 2023-11-14
fix(shell): network config cache fallback

randombenj:
### Prerequisites

### Description

When running `oh-my-posh init SHELL http://example.com/config.yml` and the user does 
not have network connection, the default config is used, even though the config is cached locally.

This change fixes this and loads the cached config if it exists.

### How

If the network call fails and the config is already cached, this config is loaded.

### DIFF
--- a/src/platform/shell.go
+++ b/src/platform/shell.go
@@ -351,6 +351,12 @@ func (env *Shell) downloadConfig(location string) error {
 	configPath := filepath.Join(env.CachePath(), filename)
 	cfg, err := env.HTTPRequest(location, nil, 5000)
 	if err != nil {
+		if _, osErr := os.Stat(configPath); !os.IsNotExist(osErr) {
+			// use the already cached config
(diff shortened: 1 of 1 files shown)


randombenj:
@JanDeDobbeleer Is the macos `golangci-lint` failure in any way related to this change, as the same command runs successfully on Windows and Linux and the error message doesn't seem to be related to any of these changes?

JanDeDobbeleer:
@randombenj seems like their latest has an issue.

JanDeDobbeleer:
@all-contributors please add @randombenj for code

This is pure magic 🪄! @holopin-bot @randombenj wizard

randombenj:
@JanDeDobbeleer thanks for the fast merge, super cool contributor experience! :tada:

JanDeDobbeleer:
@randombenj thank you for the useful addition. Makes sense to do this, if I used remote configs more often, I'd have the same need at times.

PR #4493, opened 2023-11-27
feat(project): add PEP 621 support for Python projects

### LINKED ISSUE #4461: Project segment:  Follow PEP 621 to retrieve project metadata from pyproject.toml
### Code of Conduct

Template properties retrieved from pyproject.toml should follow [PEP621](https://peps.python.org/pep-0621/#table-name) and [PyPA specification](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/). 

Maybe a "Python" project type could be added to the .Type property, which could read the .Name and .Version fields from the "[project]" table in pyproject.toml.

babieiev:
### Prerequisites

### Description

Add [PEP 621][pep621-standard] support for getting metadata from `pyproject.toml` ([project segment][project-segment]).

Resolves #4461

[pep621-standard]: https://peps.python.org/pep-0621/
[project-segment]: https://ohmyposh.dev/docs/segments/project


babieiev:
@JanDeDobbeleer can I hope for a happy little review? 😅

src/segments/project.go
```diff
@@ -95,9 +96,9 @@ func (n *Project) Init(props properties.Properties, env platform.Environment) {
 			Fetcher: n.getCargoPackage,
 		},
 		{
-			Name:    "poetry",
+			Name:    "python",
```
JanDeDobbeleer, line 99: this is a breaking change, can we duplicate this entry so it works for `python` and `poetry` still?

src/segments/project.go
```diff
@@ -175,9 +176,15 @@ func (n *Project) getPoetryPackage(item ProjectItem) *ProjectData {
 		return nil
 	}
 
+	if data.Tool.Poetry.Version != "" || data.Tool.Poetry.Name != "" {
```
JanDeDobbeleer:
```suggestion
	if len(data.Tool.Poetry.Version) != 0 || len(data.Tool.Poetry.Name) != 0 {
```

JanDeDobbeleer (changes requested):

src/segments/project.go
babieiev: Unfortunately, I don't think this is possible at the moment. As I understand, the project segment detects a specific file (or file type) and passes it to the appropriate "fetcher". That is, if I'm not mistaken, more than one fetcher cannot be responsible for the same file (in our case it is `pyproject.toml`).
But I doubt that this is a breaking change. The change does not seem to break anything, but adds the ability to read data according to a different standard. Yes, the name of the fetcher changes from "Poetry" to "Python", but as far as I know nothing is "bound" to it (other than what I already changed).

JanDeDobbeleer: The `Name` is exposed as `.Type` in the template. So if people have templates matching `poetry` as value, it will break. So we need this in the release notes. It's a breaking change in that case. I'll adjust that.

JanDeDobbeleer:
@all-contributors please add @oleksbabieiev for code,doc

This is pure magic 🪄! @holopin-bot @oleksbabieiev wizard

PR #4514, opened 2023-12-02
feat(cmd): optionally fetch status from =ExitCode

### LINKED ISSUE #4478: CMD.exe sets a hidden `=ExitCode` envvar, which the ExitCode module could use
### Code of Conduct

This isn't a "request" per se -- It's just a courtesy "heads-up" in case you didn't yet know about the following:

Whenever `cmd.exe` runs a program, it sets a hidden envvar named `=ExitCode` to an 8 digit hexadecimal representation of the exit code from the program.

While the `set` command doesn't list hidden envvars whose names begin with `=`, programs written in Go or C/C++ or etc are able to get the hidden envvars by name.  And even `echo %=exitcode%` prints the value.

Since on Windows launching OMP inherits the environment from its parent, maybe the ExitCode module could automatically get the exit code from `%=ExitCode%` when the `-e` flag isn't specified?

Note, however, that the `=ExitCode` envvar only seems to be set for programs that were executed, but not for internal commands.

JanDeDobbeleer:
### Prerequisites

### Description

resolves #4478

### DIFF
--- a/src/platform/shell.go
+++ b/src/platform/shell.go
@@ -35,6 +35,7 @@ const (
 	WINDOWS = "windows"
 	DARWIN  = "darwin"
 	LINUX   = "linux"
+	CMD     = "cmd"
 )
 
 func pid() string {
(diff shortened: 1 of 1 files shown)


chrisant996:
@JanDeDobbeleer I'm not sure I'm reading it right, but it looks like maybe `%=ExitCode%` supersedes whatever is passed to OMP in the command line flag.

If so, then I think the command line flag should always supersede `%=ExitCode%`.  Because the envvar is never cleared, and it's only set when CMD launches programs.  CMD internal commands set the actual exit code into `%ERRORLEVEL%` _without_ setting `%=ExitCode%`.  So, when using Clink, the Lua script can get the real exit code from `%ERRORLEVEL%` or `os.geterrorlevel()` and that will be more accurate than `%=ExitCode%`.

So I think `%=ExitCode%` should only be used if the exit code hadn't been passed in via the OMP command line flag.

PR #4628, opened 2024-01-15
feat: react segment

14096:
### Prerequisites

### Description

Support for React.
Enabled when package.json is present.

### DIFF
--- a/src/engine/segment.go
+++ b/src/engine/segment.go
@@ -208,6 +208,8 @@ const (
 	QUASAR SegmentType = "quasar"
 	// R version
 	R SegmentType = "r"
+	// REACT writes the current react version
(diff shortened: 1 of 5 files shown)


JanDeDobbeleer:
@srpmtt I want to make a proposal to change this from a language specific segment to a `nodepackage` segment. As it needs to always parse `package-lock.json`, just like the `quasar` segment does, it would be good to have a segment that can do this. [This](https://github.com/JanDeDobbeleer/oh-my-posh/blob/16b542eb8e4cb909c07d2967f927114e4e808d7b/src/segments/quasar.go#L60) is the logic we use in the `quasar` segment, I'm sure it can be moved so it allows an array of package names to validate and parse.

PR #4826, opened 2024-04-08
 feat(svn): Add repository name to segment

R00dRallec:
Add the name of the current svn repository to the svn segment.

### Prerequisites

### Description

This feature allows to add and display the SVN repository name as a segment property.
It works similar to the already existing 'Branch' property, but resolves the root repository name based on the `svn info` command.

### DIFF
--- a/src/segments/svn.go
+++ b/src/segments/svn.go
@@ -1,6 +1,7 @@
 package segments
 
 import (
+	"path"
 	"strconv"
 	"strings"
 
@@ -121,6 +122,18 @@ func (s *Svn) setSvnStatus() {
 		// element is the element from someSlice for where we are
 		s.Working.add(line[0:1])
 	}
+}
+
+func (s *Svn) Repo() string {
+	// Get the repository name as the last path element of the repository root URL
+	repo := s.getSvnCommandOutput("info", "--show-item", "repos-root-url")
+	base := path.Base(repo)
+
+	if base == "." {
+		return ""
+	}
+
+	return base
 }
 
 func (s *Svn) getSvnCommandOutput(command string, args ...string) string {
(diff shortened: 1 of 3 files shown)


ivan-the-terrible (approved):

src/segments/svn.go
```diff
@@ -101,6 +102,13 @@ func (s *Svn) setSvnStatus() {
 		s.Branch = branch[2:]
 	}
 
+	// Get the repository name as the last path element of the repository root URL
+	repo := s.getSvnCommandOutput("info", "--show-item", "repos-root-url")
```
JanDeDobbeleer: If you make this a function called `Repo`, you can avoid this being called every time, it's only relevant when the user adds it to a template.

JanDeDobbeleer (changes requested):

src/segments/svn.go
R00dRallec: Makes sense, thanks for the input!
Improved in commit 2716c490d016157ba4845a0375634026b3d55517.

R00dRallec: Makes sense, thanks for the input.
Resolved in commits 2716c490d016157ba4845a0375634026b3d55517 and 1b71249acc9e4abffc2d071dfa20755172cc30bb.

src/segments/svn.go
```diff
@@ -46,7 +46,7 @@ type Svn struct {
 }
 
 func (s *Svn) Template() string {
-	return " \ue0a0{{.Branch}} r{{.BaseRev}} {{.Working.String}} "
+	return " \uf401 {{.Repo}} \ue0a0{{.Branch}} r{{.BaseRev}} {{.Working.String}} "
```
JanDeDobbeleer: I would remove it from here and create a separate test, I don't think everyone wants/needs this information.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@all-contributors please add @R00dRallec for code,doc

This is pure magic 🪄! @holopin-bot @R00dRallec wizard

Issue #4850, opened 2024-04-18
Firebase segment does not work on child folders


QuentinPhrl:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Issue

The new Firebase segment proposed by @ivan-the-terrible only works at the root of a Firebase project. 

Here's an example in a firebase function project

<img width="894" alt="Prompt" src="https://github.com/JanDeDobbeleer/oh-my-posh/assets/87812493/d41dd716-768e-4d61-ac99-5fcfc87edaf7">

### Solution
As in the script I made to provide a solution before the official segment was created, all you need is the go equivalent of a startwith on the current prompt path

[https://github.com/QuentinPhrl/Omp_firebase_prompt/blob/6b6705349a0aeaccd2aaaef342bdb9e19ba27b4a/src/.firebase_project.zsh](url)

### Theme

cloud-native-azure

### What OS are you seeing the problem on?

macOS

### Which shell are you using?

zsh

### Log output

```Shell
N/A
```

ivan-the-terrible:
Working on a patch as we speak! It's as easy as `|| strings.HasPrefix(key, f.env.Pwd())`.

I'll get the code out and a proper test case for this asap!

ivan-the-terrible:
https://github.com/JanDeDobbeleer/oh-my-posh/pull/4852
Fix has been raised.

PR #4852, opened 2024-04-19
fix: Firebase Subpaths Check

ivan-the-terrible:
### Prerequisites

### Description

As per issue https://github.com/JanDeDobbeleer/oh-my-posh/issues/4850 raised, subpaths were not returning the Firebase project name when they should. 

### Work

Since Active Projects are stored as the "root", then subpaths can be checked with `strings.HasPrefix()`, which resolved this issue.
A test case has been added and a mental note will be kept at the back of my mind forever. 

### DIFF
--- a/src/segments/firebase.go
+++ b/src/segments/firebase.go
@@ -54,7 +54,7 @@ func (f *Firebase) Enabled() bool {
 	// Test if the current directory is a project path
 	// and if it is, return the project name
 	for key, value := range data.ActiveProject {
-		if key == f.env.Pwd() || strings.HasPrefix(f.env.Pwd(), key) {
+		if strings.HasPrefix(f.env.Pwd(), key) {
 			f.Project = value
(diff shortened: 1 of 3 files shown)


src/segments/firebase.go
```diff
@@ -53,7 +54,7 @@ func (f *Firebase) Enabled() bool {
 	// Test if the current directory is a project path
 	// and if it is, return the project name
 	for key, value := range data.ActiveProject {
-		if key == f.env.Pwd() {
+		if key == f.env.Pwd() || strings.HasPrefix(f.env.Pwd(), key) {
```
JanDeDobbeleer: Can't this only be the HasPrefix check? That includes equals, no?

ivan-the-terrible: You're absolutely correct! Ive updated the conditional to just the `HasPrefix()` call

PR #4857, opened 2024-04-19
fix(pulumi): improve reliability of filepath hash calculation

publicfacingusername:
Changing this line of code to use filepath.Join rather than relying on concatenation with a "/" symbol. Windows uses "\" in its file paths. So the use of "/" leads to an incorrect hash calculation and prevents this segment from working in Windows. The filepath.Join() function is already used in the getPulumiStackName() function, so this change is only needed in the getProjectName() function.

I'm not certain where docs would be added, so I'm leaving that unchecked.

### Prerequisites

### Description

### DIFF
--- a/src/segments/pulumi.go
+++ b/src/segments/pulumi.go
@@ -154,7 +154,7 @@ func (p *Pulumi) getProjectName() error {
 		return hex.EncodeToString(h.Sum(nil))
 	}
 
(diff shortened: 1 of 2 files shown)


JanDeDobbeleer:
@all-contributors please add @publicfacingusername for code

This is pure magic 🪄! @holopin-bot @publicfacingusername wizard

Issue #4879, opened 2024-04-28
New segment request: Talosctl current context


divanikus:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

It would be cool to have Talos OS's talosctl current context (https://www.talos.dev/v1.7/reference/cli/)

It should be quite simple - it's just `talosctl config info` or `context:` in `.talos/config`

ivan-the-terrible:
Reading through the documentation and trying to get familiarized with this. I see the references to the `~/.talos/config` file that `talosctl config info` would pull from.

@divanikus do you have a sample config file you could provide that contains the `context` property? I see a flag that shows it can output the context info in different formats, but I'd rather read directly from the config file.

There are similar segments that pull from config files, so this isn't a difficult thing to implement.

If you can show me the format of the data, I can build the segment for you. Talos Linux sounds neat btw. Never heard of it before seeing this issue.

divanikus:
Sure, it's a pretty simple YAML:
```
context: dev
contexts:
    dev:
        endpoints:
            - 10.1.0.10
        nodes:
            - 10.1.0.10
            - 10.1.0.20
[...]

PR #4889, opened 2024-05-01
Talosctl Segment

ivan-the-terrible:
### Prerequisites

### Description

Per https://github.com/JanDeDobbeleer/oh-my-posh/issues/4879, the Talosctl segment makes its debut.

### Work Done

As described in the above issue, `talosctl` creates a file in `.talos/config` that registers the current context. This context name (similar to the idea of `kubectl`) can now be displayed in a segment.

### Images

Example of `talosctl` segment working:

Talosctl Working

Inclusion of the segment within the `cloud-context` theme:

cloud-context theme

New documentation page:

website doc


src/segments/talosctl.go
```diff
@@ -0,0 +1,75 @@
...
+		return false
+	}
+
+	data, err := t.getTalosctlData(configFile)
+	if err != nil {
+		t.env.Error(err)
+		return false
+	}
+
+	if len(data.ActiveContext) == 0 {
```
ivan-the-terrible: @JanDeDobbeleer I feel like this `if` check is somewhat superfluous, BUT it's more readable than:
```go
t.Context = data.ActiveContext
return len(data.ActiveContext) == 0
```
Since `t.Context` will be initialized as the empty string anyways, there is no harm in writing the above, but again, readability...

src/segments/talosctl.go
```diff
@@ -0,0 +1,75 @@
...
+
+type Talosctl struct {
+	props properties.Properties
+	env   platform.Environment
+
+	Context string
+}
+
+type TalosctlData struct {
+	ActiveContext string `yaml:"context"`
```
JanDeDobbeleer: i wonder if we can use the `TalosCTL` struct for this instead of creating a new struct just for one field.

```golang
type Talosctl struct {
	props properties.Properties
	env   platform.Environment

	Context string `yaml:"context"`
}

func (t *Talosctl) getTalosctlData(configFile string) error {
	err := yaml.Unmarshal([]byte(configFile), t)
	if err != nil {
		return nil, err
	}
	return &data, nil
}
```

If that works, we might as well remove the functions as they're too small to split.

src/segments/talosctl.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	TALOSCTLNOACTIVECONFIG = "NO ACTIVE CONFIG FOUND"
+)
+
+type Talosctl struct {
```
JanDeDobbeleer:
```suggestion
type TalosCTL struct {
```

src/segments/talosctl.go
```diff
@@ -0,0 +1,75 @@
...
+	"errors"
+	"path/filepath"
+
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	TALOSCTLNOACTIVECONFIG = "NO ACTIVE CONFIG FOUND"
```
JanDeDobbeleer: as this is only used once, this can be added inline. No need to use a constant for this

JanDeDobbeleer (changes requested):

src/segments/talosctl.go
ivan-the-terrible: Added inline

src/segments/talosctl.go
ivan-the-terrible: Yup, that worked. I deleted the unnecessary function and moved the unmarshalling call within `Enabled()`.

Issue #4702, opened 2024-02-21
New segment request: Azure Developer CLI (`azd`)


SvenAelterman:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

It would be very useful to see the config details, especially the selected subscription, of the Azure Developer CLI.

JanDeDobbeleer:
@SvenAelterman can you share the command(s) you use for this? I don't use that CLI.

ivan-the-terrible:
@SvenAelterman I see the documentation on https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/overview

I haven't read entirely through it, but I assume it's akin to the existing `az` and `azfunc` commands that exist here already.

PR #5009, opened 2024-06-05
Segment for Azure Developer CLI (azd)

aaronpowell:
### Prerequisites

### Description

I wanted to show the current environment as per [`azd`](https://aka.ms/azd), so I've added a segment that will show that.

Here's an example of the prompt with it:

Prompt sample

### DIFF
--- a/src/segments/azd.go
+++ b/src/segments/azd.go
@@ -1,0 +1,78 @@
+package segments
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"strings"
+
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+)
+
+type Azd struct {
+	props properties.Properties
+	env   platform.Environment
+
+	azdConfig
+}
+
+type azdConfig struct {
+	Version            int    `json:"version"`
+	DefaultEnvironment string `json:"defaultEnvironment"`
+}
+
+func (t *Azd) Template() string {
+	return " \uebd8 {{ .DefaultEnvironment }} "
+}
+
+func (t *Azd) Init(props properties.Properties, env platform.Environment) {
+	t.props = props
+	t.env = env
+}
+
+func (t *Azd) Enabled() bool {
+	var parentFilePath string
+
+	folders := t.props.GetStringArray(LanguageFolders, []string{".azure"})
+	for _, folder := range folders {
+		if file, err := t.env.HasParentFilePath(folder); err == nil {
+			parentFilePath = file.ParentFolder
+			break
+		}
+	}
+
+	if len(parentFilePath) == 0 {
+		t.env.Debug("No .azure folder found in parent directories")
+		return false
+	}
+
+	dotAzureFolder := filepath.Join(parentFilePath, ".azure")
+	files := t.env.LsDir(dotAzureFolder)
+
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+
+		if strings.EqualFold(file.Name(), "config.json") {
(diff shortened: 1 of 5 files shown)


JanDeDobbeleer:
@aaronpowell this resolves https://github.com/JanDeDobbeleer/oh-my-posh/issues/4702, right?

aaronpowell:
Yeah, I don't have subscription in there, it's not as relevant in an azd environment

src/segments/azd.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (t *Azd) Init(props properties.Properties, env platform.Environment) {
+	t.props = props
+	t.env = env
+}
+
+func (t *Azd) Enabled() bool {
+	var location string
+
+	folders := []string{".azure"}
```
JanDeDobbeleer: if for now this is one folder only, we can design for that today and still proceed to an array later. Unless if we allow the user to override this logic:

```suggestion
	folders := t.props.GetStringArray(LanguageFolders, []string{".azure"}) 
```

src/segments/azd.go
```diff
@@ -0,0 +1,77 @@
...
+	return " {{ .DefaultEnvironment }} "
+}
+
+func (t *Azd) Init(props properties.Properties, env platform.Environment) {
+	t.props = props
+	t.env = env
+}
+
+func (t *Azd) Enabled() bool {
+	var location string
```
JanDeDobbeleer:
```suggestion
	var parentFolder string
```

src/segments/azd.go
```diff
@@ -0,0 +1,77 @@
...
+			break
+		}
+	}
+
+	if len(location) == 0 {
+		t.env.Debug("No .azure folder found in parent directories")
+		return false
+	}
+
+	files := t.env.LsDir(filepath.Join(location, ".azure"))
```
JanDeDobbeleer:
```suggestion
    dotAzureFolder := filepath.Join(location, ".azure")
	files := t.env.LsDir(dotAzureFolder)
```

`filepath.Join` is repeated

src/segments/azd.go
```diff
@@ -0,0 +1,77 @@
...
+
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+
+		if strings.EqualFold(file.Name(), "config.json") {
+			return t.TryReadConfigJSON(filepath.Join(location, ".azure", file.Name()))
```
JanDeDobbeleer:
```suggestion
			return t.TryReadConfigJSON(filepath.Join(dotAzureFolder, file.Name()))
```

src/segments/azd_test.go
```diff
@@ -0,0 +1,70 @@
...
+		env.On("Debug", mock2.Anything)
+
+		if tc.IsInited {
+			fileInfo := &platform.FileInfo{
+				Path:         "test/.azure",
+				ParentFolder: "test",
+				IsDir:        true,
+			}
+			env.On("HasParentFilePath", ".azure").Return(fileInfo, nil)
+			dirEntries := []fs.DirEntry{
```
JanDeDobbeleer, line 48: instead of having `IsInited` I would expose `DirEntries` so you can also test that logic (no `config.json` for example).

JanDeDobbeleer (changes requested):

src/segments/azd_test.go
aaronpowell: To the best of my knowledge, there's no way that you could have the `.azure` file without a `config.json` file, without having something really broken with `azd`. Would you still want to handle that scenario then?

aaronpowell:
hmm, I should have tested after accepting the changes 🤣

JanDeDobbeleer:
@all-contributors please add @aaronpowell for code,design,doc

This is pure magic 🪄! @holopin-bot @aaronpowell wizard

PR #5028, opened 2024-06-08
docs: Update share-theme.md to clarify output filename

EDIflyer:
### Prerequisites

### Description
Added clarification that the output file will literally be `.png` - when I first ran the command I couldn't see an export then realised it was called this and hidden by default as it's a dotfile.

Not sure if it might be worth changing it so that if no export filename is specified a default (e.g., `omptheme.png` or similar) is used to stop it ended up being a hidden file?  For just now I've just updated the docs though.

### DIFF
--- a/src/cli/config_export.go
+++ b/src/cli/config_export.go
@@ -88,11 +88,13 @@ func cleanOutputPath(path string, env runtime.Environment) string {
 		path = strings.TrimPrefix(path, "~")
 		path = filepath.Join(env.Home(), path)
 	}
+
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer:
@EDIflyer it makes sense to add a sane default name. Also, for configurations that start with a dot, we best strip those as well.

JanDeDobbeleer:
@all-contributors please add @EDIflyer for code,doc

This is pure magic 🪄! @holopin-bot @EDIflyer wizard

PR #5031, opened 2024-06-09
feat(segment): add `bun`

Joxtacy:
### Prerequisites

### Description

Adds a segment for [`bun`](https://bun.sh/).

### DIFF
--- a/src/segments/bun.go
+++ b/src/segments/bun.go
@@ -1,0 +1,34 @@
+package segments
+
+import (
+	"github.com/jandedobbeleer/oh-my-posh/src/platform"
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+)
+
+type Bun struct {
+	language
+}
+
+func (b *Bun) Template() string {
+	return languageTemplate
+}
+
+func (b *Bun) Init(props properties.Properties, env platform.Environment) {
+	b.language = language{
+		env:        env,
+		props:      props,
+		extensions: []string{"bun.lockb"},
+		commands: []*cmd{
+			{
+				executable: "bun",
+				args:       []string{"--version"},
+				regex:      `(?:(?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+))))`,
+			},
+		},
+		versionURLTemplate: "https://github.com/oven-sh/bun/releases/tag/bun-v{{.Full}}",
+	}
+}
+
+func (b *Bun) Enabled() bool {
+	return b.language.Enabled()
+}
--- a/src/engine/segment.go
+++ b/src/engine/segment.go
@@ -111,6 +111,8 @@ const (
 	BREWFATHER SegmentType = "brewfather"
 	// Buf segment writes the active buf version
 	BUF SegmentType = "buf"
+	// BUN writes the active bun version
+	BUN SegmentType = "bun"
 	// CARBONINTENSITY writes the actual and forecast carbon intensity in gCO2/kWh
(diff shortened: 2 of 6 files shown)


Joxtacy:
I tried to add 🥟 as the emoji using the code point `\0001f95f`, but it doesn't render correctly in the docs.
3145da00de787b932b7ecf6f175e97a7


It works as expected in my terminal (alacritty) when I add it in my template `template = "[\U0001f95f {{ .Full }}]"` though.
Screenshot 2024-06-09 at 11 51 00


Any input on how I should proceed here?

JanDeDobbeleer (approved):

src/segments/bun.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (b *Bun) Template() string {
+	return languageTemplate
+}
+
+func (b *Bun) Init(props properties.Properties, env platform.Environment) {
+	b.language = language{
+		env:        env,
+		props:      props,
+		extensions: []string{"*.js", "*.ts", "bun.lockb"},
```
JanDeDobbeleer: do we want to use the same extensions as the node segment? Or only trigger on `bun.lock`?

JanDeDobbeleer:
@Joxtacy maybe the docs Nerd Font isn't V3 yet?

src/segments/bun.go
Joxtacy: I was thinking about this as well. When testing things out, I noticed myself setting `extensions = ["bun.lockb"]` anyway. 
I more or less just copied what we have for `Deno`. Technically you could use the bun cli outside of a bun project.
I don't really have any strong preferences here. Personally I think just having the `bun.lockb` is fine.

Joxtacy:
> @Joxtacy maybe the docs Nerd Font isn't V3 yet?

It seems like it's the `JSON.stringify`, `YAML.stringify`, `TOML.stringify` functions that aren't replacing the code point correctly. I could create a makeshift replacer function for both JSON and YAML to manually call `String.fromCodePoint("0x0001f95f")` and that then renders the emoji correctly.

Not sure what to do about that. Maybe I can just find another icon that plays more nicely. 🙃

src/segments/bun.go
JanDeDobbeleer: I would indeed keep `bun.lockb` only as the default, people can override as they see fit.

JanDeDobbeleer:
@Joxtacy what if you paste in the actual emoji rather than the code point?

Joxtacy:
> @Joxtacy what if you paste in the actual emoji rather than the code point?

That works just fine. I thought you wanted to have the code points, but maybe I misunderstood the contribution docs.

JanDeDobbeleer:
> I thought you wanted to have the code points

@Joxtacy back in the day, check, but now as they're being rendered, this isn't required anymore.

src/segments/bun.go
JanDeDobbeleer: @Joxtacy if you adjust this part, we're ready to merge 👍🏻

Joxtacy: Done! :) If you'd just like to double check to make sure I didn't miss anything, I'd be happy as well.

JanDeDobbeleer (approved):

src/segments/bun.go
JanDeDobbeleer: already did, and if not, that's just a small PR away 😄

Joxtacy:
I forgot to update my test... 🙃

JanDeDobbeleer:
@all-contributors please add @Joxtacy for code,design,doc

This is pure magic 🪄! @holopin-bot @Joxtacy wizard

PR #5060, opened 2024-06-12
Update install.sh

Dartypier:
### Prerequisites

### Description

These are the corrected behaviors:

- ANSI \x1b breaks output in some cases (ex. try to execute the script when oh-my-posh is already installed: it will print out also \x1b). Better use `\e` that doesn't give this strange result in Linux terminal.

Example of the bug
```bash
\x1b[31mcurl is required to install Oh My Posh. Please install curl and try again.
```

- executing the script with sudo will check the permission on regular user, not for sudo/root. So it will cancel the installation when you actually have writing permission. Now also a sudo check is done.

- `==` operator is not intended for string check inside test `[]`. REF: https://linuxize.com/post/how-to-compare-strings-in-bash/. Replaced with `=`. Note: zsh `==` is correct to check strings, but that's not the correct syntax for bash; and the install script is intended to be executed by bash (as it should be as standard).

### DIFF
--- a/website/static/install.sh
+++ b/website/static/install.sh
@@ -5,7 +5,7 @@ themes_dir=""
 executable=""
 
 error() {
-    printf "\x1b[31m$1\e[0m\n"
+    printf "\e[31m$1\e[0m\n"
     exit 1
 }
 
@@ -14,7 +14,7 @@ info() {
 }
 
 warn() {
-    printf "⚠️  \x1b[33m$1\e[0m\n"
+    printf "⚠️  \e[33m$1\e[0m\n"
 }
 
 help() {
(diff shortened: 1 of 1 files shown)


Dartypier:
Also a shebang for bash should be included at the top of the file, to make clear is a bash file and not a zsh file:

```bash
#!/bin/bash
```

JanDeDobbeleer:
> Also a shebang for bash should be included at the top of the file

@Dartypier that wasn't added yet, right?

Dartypier:
> > Also a shebang for bash should be included at the top of the file
> 
> @Dartypier that wasn't added yet, right?

It was present actually, but the shell called is `sh` that might not be `bash`. Changed from `#!/usr/bin/env sh` to `#!/usr/bin/env bash`.

Ref: https://askubuntu.com/questions/141928/what-is-the-difference-between-bin-sh-and-bin-bash

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @Dartypier for doc

This is pure magic 🪄! @holopin-bot @Dartypier wizard

JanDeDobbeleer:
@Dartypier there's an issue with sudo not always being available. You can read about it [here](https://discord.com/channels/1023597603331526656/1023597603331526659/1250535176342343780).

Issue #2692, opened 2022-08-22
Fish $( ... ) is not supported


ehawman:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

Had OMP installed on an RPI. Config got borked so I wiped it. Now Fish is crying that oh-my-posh init fish doesn't conform.

![image](https://user-images.githubusercontent.com/52979227/185998818-606da962-1cf1-4a53-9e11-da0acaf98bcd.png)

```
- (line 59): $(...) is not supported. In fish, please use '(string)'.
if test "$(string match -e '_render_transient' $(bind \r --user 2>/dev/null))" != ''
         ^
from sourcing file -
source: Error while reading file “<stdin>”
```

Installed via

```
sudo wget https://github.com/JanDeDobbeleer/oh-my-posh/releases/latest/download/posh-linux-arm -O /usr/local/bin/oh-my-posh
sudo chmod +x /usr/local/bin/oh-my-posh
```

I'm very confused since this was working before. Also Bash can run it just fine.

<details>
 <summary>Output of oh-my-posh init fish</summary>

```
set --export POSH_THEME /home/abcdefg/.cache/oh-my-posh/config.omp.json
set --global POWERLINE_COMMAND "oh-my-posh"
set --global CONDA_PROMPT_MODIFIER false
set --global omp_tooltip_command ""
set --global omp_transient 0

function fish_prompt
    set --local omp_status_cache_temp $status
    # clear from cursor to end of screen as
    # commandline --function repaint does not do this
    # see https://github.com/fish-shell/fish-shell/issues/8418
    printf \e\[0J
    if test "$omp_transient" = "1"
      /usr/local/bin/oh-my-posh print transient --config $POSH_THEME --shell fish --error $omp_status_cache --execution-time $omp_duration --stack-count $omp_stack_count --shell-version $FISH_VERSION
      return
    end
    set --global omp_status_cache $omp_status_cache_temp
    set --global omp_stack_count (count $dirstack)
    set --global omp_duration "$CMD_DURATION$cmd_duration"
    # check if variable set, < 3.2 case
    if set --query omp_lastcommand; and test "$omp_lastcommand" = ""
      set omp_duration 0
    end
    # works with fish >=3.2
    if set --query omp_last_status_generation; and test "$omp_last_status_generation" = "$status_generation"
      set omp_duration 0
    end
    if set --query status_generation
      set --global --export omp_last_status_generation $status_generation
    end

[...]

Issue #2475, opened 2022-07-02
Repaint only the right prompt for rendering a tooltip and preserve the previous in fish


lewis-yeung:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

Currently, if we set a tip of `git` for a `git` segment, and get it rendered as a toolip in fish, after we enter `status` and a space character, the previously rendered tooltip will be overwritten by the original right prompt (or in my case, it will disappear if no right prompt is set), as the screenshots show below:

![tooltip-rendered](https://user-images.githubusercontent.com/83903009/176994345-1478c189-0405-412e-9c3d-395b1552ad64.png)
![tooltip-overwritten](https://user-images.githubusercontent.com/83903009/176994344-2dc62e9a-44a0-480b-ba32-ff4ee815c06d.png)

Actually, not only the right prompt, but the whole is re-rendered in this case. That means the previous contents of all segments will lose, as the screenshots show below:

![before-repainting](https://user-images.githubusercontent.com/83903009/176994137-bd394f56-442a-47ae-a387-b91db74ee510.png)
![after-repainting](https://user-images.githubusercontent.com/83903009/176994135-575b86ce-6435-470b-ad0a-6ddd64437cb2.png)

That's a little unreasonable, isn't it?

Related code:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/2710f796131a5502e4c8c6b63adc6578ba59bd07/src/shell/scripts/omp.fish#L64

I just wonder if it is possible to repaint only the right prompt in fish. Besides, I'd like the previously rendered tooltip to remain when there is no new one.

JanDeDobbeleer:
@lewis-yeung that's a limitation of fish as there's no way to repaint the right prompt only (looking at the [docs](https://fishshell.com/docs/current/cmds/bind.html)). So you'll need a feature request there.

PR #5070, opened 2024-06-13
Bug fixes and more consistent prompt behaviors across different shells

lewis-yeung:
### Prerequisites

### Description

- **Major change**: Avoid unnecessary CLI calls for prompt repaint in ~PowerShell (unsupported with the `ConstrainedLanguage` mode),~ Zsh and Fish (#2475), making the prompt behaviors (mainly transient prompts and tooltips) more consistent across different shells. This also improves performance to some extent.
- **Minor changes/refactorings**:
  - ~The right prompt rendering has been separated from the primary for PowerShell.~
  - In the initialization script for Fish (`omp.fish`), now we do not use command substitution with a prefix dollar sign (`$`) so that an error of `$(...) is not supported` (#2692) will never occur even in a version of Fish lower than v3.4.0. The documentation is also updated.
  - The [suggestion list bug][suggestion-list-bug] no longer appears in the latest PSReadLine, so the workaround (#2567) is now safely removed from the initialization script for PowerShell (`omp.ps1`).
  - The initialization scripts for Fish (`omp.fish`) has been formatted using consistent indentations.

Seems like heavy stuff but all carefully validated. ✅

**Updated:** The PowerShell prompt caching changes have been extracted from this PR and may be further reworked.

**Updated:** This PR will also resolve a regression introduced in #5222 ([related change][related-change]), where a tooltip is rendered as an "empty" block in PowerShell.

[suggestion-list-bug]: https://github.com/JanDeDobbeleer/oh-my-posh/pull/2567#issue-1317067611
[related-change]: https://github.com/JanDeDobbeleer/oh-my-posh/commit/dcc439c941bcb34135f8d153f35861d2ddae07ad#diff-13d678abd11e6f042974803b799b26e028b7e68e48358fa4ee0a93bf765d47c7L48

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -14,20 +14,29 @@ function global:Get-PoshStackCount {
 }
 
 New-Module -Name "oh-my-posh-core" -ScriptBlock {
+    # Check `ConstrainedLanguage` mode.
+    $script:ConstrainedLanguageMode = $ExecutionContext.SessionState.LanguageMode -eq "ConstrainedLanguage"
+
+    # Prompt related backup.
+    $script:OriginalPromptFunction = $Function:prompt
+    $script:OriginalContinuationPrompt = (Get-PSReadLineOption).ContinuationPrompt
+    $script:OriginalPromptText = (Get-PSReadLineOption).PromptText
+
+    $script:NoExitCode = $true
     $script:ErrorCode = 0
     $script:ExecutionTime = 0
     $script:OMPExecutable = ::OMP::
     $script:ShellName = "::SHELL::"
     $script:PSVersion = $PSVersionTable.PSVersion.ToString()
     $script:TransientPrompt = $false
-    $script:ToolTipCommand = ""
+    $script:TooltipCommand = ''
     $env:POWERLINE_COMMAND = "oh-my-posh"
     $env:POSH_SHELL_VERSION = $script:PSVersion
     $env:POSH_PID = $PID
     $env:CONDA_PROMPT_MODIFIER = $false
 
     # set the default theme
-    if ((::CONFIG:: -ne '') -and (Test-Path -LiteralPath ::CONFIG::)) {
+    if (::CONFIG:: -and (Test-Path -LiteralPath ::CONFIG::)) {
         $env:POSH_THEME = (Resolve-Path -Path ::CONFIG::).ProviderPath
     }
 
@@ -56,7 +65,7 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
             [string[]]$Arguments = @()
         )
 
-        if ($ExecutionContext.SessionState.LanguageMode -eq "ConstrainedLanguage") {
+        if ($script:ConstrainedLanguageMode) {
             $standardOut = Invoke-Expression "& `$FileName `$Arguments 2>&1"
             $standardOut -join "`n"
             return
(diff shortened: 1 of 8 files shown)


src/shell/scripts/omp.ps1
```diff
@@ -120,108 +128,105 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
...
+                        return
+                    }
+
+                    # Save the tooltip prompt to avoid unnecessary CLI calls.
+                    $script:TooltipPrompt = $standardOut.TrimStart()
+                    $paddingLength = $standardOut.Length - $script:TooltipPrompt.Length
+                    $script:TooltipPromptWidth = $terminalWidth - $position.X - $paddingLength
+
+                    # Remove the right-aligned prompt (if any) from the saved prompt.
+                    $script:CurrentPrompt = $script:CurrentPrompt -replace "^([\s\S]+)`e7\s*.+`e8([\s\S]+)?$", '$1$2'
```
JanDeDobbeleer: this feels so hacky. I know it will work but it's a recipe for disaster in my opinion. The previous implementation was easier to understand.

lewis-yeung: This workaround exists because for PowerShell the part of right prompt rendering is merged into that of the primary prompt. Or, should we separate out the logic in the OMP engine, making it just like other shells that have native right prompt support? Seems less hacky but does require one more CLI call, which may lead to slight performance degradation of increasing latency when rendering full prompts.

JanDeDobbeleer: @lewis-yeung we didn't do this before though, it's the prompt caching that introduces this. I wonder if we can do this differently as this isn't code I like seeing. I get the challenge here, but I think we can do better.

lewis-yeung: @JanDeDobbeleer If we directly append a tooltip to the cached prompt, an existing right prompt (if any) will be shown in a flash and be instantly overwritten by the tooltip during a repaint. Anyway, to implement this, separating the right prompt rendering from the primary seems to be our only choice and is easy for future maintenance.

lewis-yeung: Reworked and ready for a review now.

JanDeDobbeleer:
@lewis-yeung I did not forget about this. Although you might need to update this based on my latest refactor. Sorry about that.

lewis-yeung:
@JanDeDobbeleer Never mind.

lewis-yeung:
Freshly rebased.

JanDeDobbeleer:
Reviewing tomorrow 🙏🏻

src/shell/scripts/omp.ps1
```diff
@@ -120,115 +134,128 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
...
+
+    function Update-PoshRPrompt {
+        param([string]$NewRPrompt)
+
+        $position = $Host.UI.RawUI.CursorPosition
+        $terminalWidth = Get-TerminalWidth
+        if ($NewRPrompt) {
+            # Save the right/tooltip prompt to avoid unnecessary CLI calls.
+            $script:CurrentRPrompt = $NewRPrompt.TrimStart()
+            $paddingLength = $NewRPrompt.Length - $script:CurrentRPrompt.Length
```
JanDeDobbeleer: These prompts contain ANSI characters, how is the length calculation correct?

lewis-yeung: Just as the code does. The prompt rendered by the OMP executable always contains leading spaces as padding for right-alignment, so we can calculate *padding_length*, and the **real** tooltip/right prompt length is *terminal_length* - *cursor_horizontal_position* - *padding_length*, where counting ANSI characters is avoided.

JanDeDobbeleer:
@lewis-yeung Can we extract the powershell changes from this one in a separate PR? I want to look at caching the prompt inside the oh-my-posh cache rather than doing length calculations here. I don't feel like maintaining that logic in two spots.

lewis-yeung:
@JanDeDobbeleer Yes, but in a few days since I am not available now.

lewis-yeung:
> I want to look at caching the prompt inside the oh-my-posh cache rather than doing length calculations here.

@JanDeDobbeleer That way we are not able to avoid an OMP CLI call when repainting the prompt in PowerShell because we have to read stuff from the cache file.

JanDeDobbeleer:
> That way we are not able to avoid an OMP CLI call when repainting the prompt in PowerShell because we have to read stuff from the cache file.

@lewis-yeung check. But it will make a difference to fetch from cache rather than render the entire part again. I really do not want to see any positioning logic outside of the executable, purely for maintaining purposes. It'll become a maze before we know it.

lewis-yeung:
> purely for maintaining purposes

Fair point. Will rebase again.

JanDeDobbeleer (approved):

PR #5080, opened 2024-06-14
feat(upgrade): auto upgrade

JanDeDobbeleer:
- **feat(upgrade): use built-in upgrade by default**
- **feat(upgrade): auto upgrade**
- **chore(go): update dependencies**

### Prerequisites

### Description

This PR introduces the `auto_upgrade` configuration parameter which will trigger the
upgrade using the same interval as the upgrade notice on shell startup.

Currently a darkly launched feature so it can be experimented with.


.vscode/launch.json
```diff
@@ -74,7 +74,7 @@
         "prompt",
         "init",
         "cmd",
-        "--print"
+        "--config=/Users/jan/.posh.omp.jsonc"
```
JanDeDobbeleer: needs to be removed

kimsey0:
@JanDeDobbeleer, thanks for the new auto-upgrade feature. I don't know where, if anywhere, you take feedback on it, but it would be great to remove or at least allow disabling the `✅ no new version available` notice in every new terminal window. It doesn't really convey any information, so it's just noise to me.

https://github.com/JanDeDobbeleer/oh-my-posh/blob/170e5ac9e5b12d918240add12078a0fadd71330b/src/cli/upgrade.go#L42

JanDeDobbeleer:
@kimsey0 that's solved by adding `disable_notice: true`

kimsey0:
But does that still show the actual upgrade message?

JanDeDobbeleer:
Yes, when it upgrades it will still show that.

kimsey0:
OK. I think this interaction sounds a bit confusing when spelled out ("`disable_notice` disables the notice that a new upgrade is available when `auto_upgrade` is disabled, and the notice that no new upgrade is available when `auto_upgrade` is enabled"), but would it still make sense to document, as in #5175?

PR #5096, opened 2024-06-16
Allow nushell to have set_poshcontext like other shells

Nibodhika:
### Prerequisites

### Description

This change allows nushell to execute `set_poshcontext` like other shells do.

Following the same example on the docs for nushell would be:

```nu
def --env set_poshcontext [] {
	$env.POSH = ( date now );
}
```

I don't know how to update the doc

### DIFF
--- a/src/shell/scripts/omp.nu
+++ b/src/shell/scripts/omp.nu
@@ -28,6 +28,10 @@ $env.PROMPT_COMMAND = { ||
         $clear = (history | is-empty) or ((history | last 1 | get 0.command) == "clear")
     }
 
+    if ($env.SET_POSHCONTEXT? | is-not-empty) {
+        do --env $env.SET_POSHCONTEXT
+    }
+
     ^::OMP:: print primary $"--config=($env.POSH_THEME)" --shell=nu $"--shell-version=($env.POSH_SHELL_VERSION)" $"--execution-time=(posh_cmd_duration)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=(posh_width)" $"--cleared=($clear)"
 }
 
--- a/website/docs/configuration/templates.mdx
+++ b/website/docs/configuration/templates.mdx
@@ -91,6 +91,15 @@ end
 ```
 
 </TabItem>
+<TabItem value="nu">
+
+```bash
+$env.SET_POSHCONTEXT = {
+	$env.POSH = ( date now );
+}
+```
+
+</TabItem>
 </Tabs>
 
 :::


src/shell/scripts/omp.nu
```diff
@@ -21,6 +21,7 @@ def posh_width [] {
 $env.PROMPT_MULTILINE_INDICATOR = (^::OMP:: print secondary $"--config=($env.POSH_THEME)" --shell=nu $"--shell-version=($env.POSH_SHELL_VERSION)")
 
 $env.PROMPT_COMMAND = { ||
+    try { set_poshcontext };
```
JanDeDobbeleer: Is there a performance impact with a try catch rather than overriding an empty function (if possible)?

Nibodhika: I'm relatively new to nushell, couldn't find any information on performance penalties for try/catch blocks, but testing it on my machine I didn't spot any penalty (but also didn't ran any benchmark). As for the alternative I also couldn't get it to work, it complains about the redeclaration of the function.

JanDeDobbeleer: Thanks. I'll have a quick benchmark myself and if there's no real penalty we'll go ahead and merge.

JanDeDobbeleer: @Nibodhika this adds around 25ms to prompt invocation which is unacceptable. Can we have a look at using a mutable [closure](https://www.nushell.sh/book/types_of_data.html#closures) instead?

JanDeDobbeleer:
```
mut set_poshcontext = { return }

// in prompt:

do $set_poshcontext
```

override using 


```
$set_poshcontext = { echo world }
```

this executes in 21 microseconds

Nibodhika: Cool, I'll change accordingly, how do you run the benchmark so I can run it myself for testing?

JanDeDobbeleer: @Nibodhika there's a `timeit` function in nushell.

JanDeDobbeleer: @Nibodhika so far I tried to solve this efficiently but I'm stuck as there's two things we can't do in nushell:

1. override functions
2. use a mutable closure inside the prompt environment variable

As we don't have any options that are fast, I'm about to close this and the issue due to not being possible at the time.

@fdncred do you have any pointers on how we could achieve this without using `try { set_poshcontext };`?

Nibodhika: I've been trying to find another way as well, I also thought about trying to check if the function exists but couldn't find any way.

But I think I found a way that would only require change in documentation, i.e. having something like this on your config after the source of the installed file:

```bash
let original_prompt = $env.PROMPT_COMMAND

$env.PROMPT_COMMAND = { ||
  # new code here
  do $original_prompt
}
```

Need to run some tests to make sure, but seems to be working.

Nibodhika: I just finished running some tests, the solution I proposed on the last comment seems to work well, but also I was not able to find the 25ms extra for the try block. This is how I'm testing the speed in case you think there's something wrong here:

```nushell
let amount = 1000; mut total_time = 0sec; for $x in 1..$amount { let time = timeit do $env.PROMPT_COMMAND; $total_time += $time }; let avg_time = $total_time / $amount ; print "Average time: " + $avg_time;
```
I get very close values (less than 1ms difference) for all 3 scenarios (No try block, try block without the function being defined, try block with the function being defined).

No try block: 12ms 826µs 289ns
Try block without the function defined: 13ms 243µs 74ns
Try block with the function defined: 12ms 654µs 903ns

There is a consistent performance degradation from the try block, but it's less than 1ms when the function is not implemented, and considering that with the function defined it ran almost 200µs faster, the only way that this makes sense is a variation of more than 200µs is expected between runs of this test depending on other conditions, so the 1ms difference might be even less.

I think it's possible that you did your test with a function that did things, so you added the time of whatever you were doing to the test which is not a fair comparison, I tried it with an empty function and a function that set the env variable I needed hardcoded and obtained very similar 12ms times, having the function do some work to get the data I need did elevated the time to 15ms but that's to be expected since there's more stuff being executed.

fdncred: Hey @JanDeDobbeleer! Happy to help if I can. I confess, I don't fully understand what we're trying to achieve here but it looks like the problem is that we want to call into a `set_poshcontext` nushell custom command but we don't know if that custom command is defined. 

One way to achieve this might be to check and see if it's defined before trying to call it. Example: `let set_poshcontext_exists = scope commands | where type == custom and name == set_poshcontext | length` then `if $set_poshcontext_exists > 0 { set_poshcontext } else { null }`. You could also try `help commands` with the same pipeline, it could have different performance than `scope commands`.

Another option maybe is to always have `set_poshcontext` defined and shadow it with another function of the same name when you need it for real.

One thing to know is that `$env.PROMPT_COMMAND` can contain a pretty complicated script. Here's what mine looks like below, where [oh-my.nu](https://github.com/nushell/nu_scripts/blob/main/modules/prompt/oh-my.nu) calls a `git_prompt` function. I've used this prompt for years and it's not meant to demonstrate the perfect prompt, I'll leave that to you. It was really meant to imitate omp/omzsh with some nushell hackery.
```nushell
use "modules/prompt/oh-my.nu" git_prompt
$env.PROMPT_COMMAND = {|| (git_prompt).left_prompt }
$env.PROMPT_COMMAND_RIGHT = {|| (git_prompt).right_prompt }
$env.PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) {$"(ansi green_bold)\n❯ (ansi reset)"} else {$"(ansi red_bold)\n❯ (ansi reset)"}}
$env.TRANSIENT_PROMPT_COMMAND = { "" }
$env.TRANSIENT_PROMPT_COMMAND_RIGHT = { || (git_prompt).right_prompt }
$env.TRANSIENT_PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) {$"(ansi light_yellow_bold)❯ (ansi reset)"} else {$"(ansi light_red_bold)❯ (ansi reset)"}}
```

JanDeDobbeleer: > Another option maybe is to always have set_poshcontext defined and shadow it with another function of the same name when you need it for real.

@fdncred how do we do this? Because that's exactly what we do for other shells.

Nibodhika: What we're trying to achieve is a way to have generic code execute before each prompt is rendered, like it's doable on other terminals by defining the `set_poshcontext` function.  Unfortunately overwriting the function doesn't seem possible, if the function is declared with def or let is not mutable, if it's declared with mut it can't be stored inside another variable.

Personally I prefer the try method as that keeps nu in parity with the other shells, but I'm also not seeing the performance degradation that @JanDeDobbeleer does so it might not be an option if the performance degradation happens for other people

JanDeDobbeleer: @Nibodhika it's not a user noticeable degradation, it just adds an invisible 23ms to the prompt execution which I dislike.

fdncred: > @fdncred how do we do this? Because that's exactly what we do for other shells.

It would have to be defined in the nushell script that omp already `source`s or `use`s and then right before using it, if needed, redefine it. For instance, in the beginning of the script you'd have something like
```nushell
def --env set_poshcontext [] {}
```
and then right before usage you'd just define it differently like
```nushell
def --env set_poshcontext [] {
  # do something different here
}
```
This second definition of the custom command would _not_ overwrite the first but it would shadow the first. I believe in the scope where it's used the shadowed custom command could be used but when outside of that scope, the initially defined custom command would be used.

It also seems like this would work https://github.com/JanDeDobbeleer/oh-my-posh/pull/5096#discussion_r1641810408 as already suggested but not making it mutable, just shadowing it with another let to shadow the closure.

I guess all this is just nushell script so maybe I can play around with it after while and understand better what the problem really is. I think I'm missing something fundamental with just talking about it vs trying it.

fdncred: If I define a `set_poshcontext` custom command _inside_ the .oh-my-posh.nu script, you can see it's being called.
```nushell
... stuff above this
def --env set_poshcontext [] {
    # if you comment the line below out, this function will still run but not return anything
    $env.POSH = (date now | format date "%Y-%m-%d %H:%M:%S")
}

# PROMPTS
$env.PROMPT_MULTILINE_INDICATOR = (^"C:/Users/us991808/AppData/Local/Programs/oh-my-posh/bin/oh-my-posh.exe" print secondary $"--config=($env.POSH_THEME)" --shell=nu $"--shell-version=($env.POSH_SHELL_VERSION)")

$env.PROMPT_COMMAND = { ||
    # hack to set the cursor line to 1 when the user clears the screen
    # this obviously isn't bulletproof, but it's a start
    mut clear = false
    set_poshcontext
    if POSH in $env { print $env.POSH }
    if $nu.history-enabled {
        $clear = (history | is-empty) or ((history | last 1 | get 0.command) == "clear")
    }

    ^"C:/Users/us991808/AppData/Local/Programs/oh-my-posh/bin/oh-my-posh.exe" print primary $"--config=($env.POSH_THEME)" --shell=nu $"--shell-version=($env.POSH_SHELL_VERSION)" $"--execution-time=(posh_cmd_duration)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=(posh_width)" $"--cleared=($clear)"
}
... stuff below this
```
and in the repl
```nushell
2024-06-17 09:45:39
~\source\repos\forks\nushell
❯

Comment (Nibodhika) [reply src/shell/scripts/omp.nu]: @JanDeDobbeleer I showed above my methodology for calculating average time, and the PROMPT_COMMAND function is getting called in 12/13ms regardless of having the try or not, I'm not seeing anything take anywhere close to 25ms more for adding the try. Which is why I showed how I tested this to see if there's something wrong with it, essentially I ran the timeit on it 1000 times (because I saw lots of fluctuation between one run and the next) and took the average, I did this on 3 different cases:
* No try block
* try block without the function existing
* try block with the function existing
All 3 of them ran on an average of 12/13ms at most it added 1ms, nothing close to the 25 you're mentioning. Which is why I want to know if I should be testing it differently or how are you seeing those 25ms.

Comment (JanDeDobbeleer) [reply src/shell/scripts/omp.nu]: I ran `timeit` straight on the try block. If we can work with @fdncred 's suggestion, that would be preferred as it follows the same logic as the other shells.

image

**Comment by kubouch** [comment]
For debugging performance you can try the `debug profile` command.

File: src/shell/scripts/omp.nu (hunk 1 above)
Comment (Nibodhika) [reply src/shell/scripts/omp.nu]: The if method is much slower than the try, it runs in 18ms vs the 13ms of the try method, so I imagine this would be potentialized on your machine since this is how the same try test runs on my machine:
2024-06-17-213655_hyprshot

I think I found a way that's cleaner. Env is allowed to be mutable, so if the closure is defined inside an env variable it can be changed afterwards, i.e.

```nushell
$env.SET_POSHCONTEXT = {|| }

$env.PROMPT_COMMAND = { ||
    do --env $env.SET_POSHCONTEXT
    ...
```

and inside the config.nu
```nushell
source ~/.oh-my-posh.nu

$env.SET_POSHCONTEXT = {
	$env.POSH = ( date now );
}
```

The only caveat is that the overwriting of `SET_POSHCONTEXT` needs to happen after the sourcing of the .oh-my-posh file, but I imagine this is the same for other shells as well. As for times it should be the fastest way since the function is always there, but is empty by default. Is this solution acceptable?

devyn: > I think I found a way that's cleaner. Env is allowed to be mutable, so if the closure is defined inside an env variable it can be changed afterwards

This sounds probably like the best solution to me.

The reason why your

```nushell
timeit { try { nope } }
```

takes so long is that when a command doesn't exist in Nushell, it has to try to see if it might exist as an external command, which unlike the internal commands is *not* something we resolve at parse time.

The other challenge for you is that even if you redefine a custom command, it won't actually affect anything that had previously been parsed that was referencing it. I think probably using the `$env` as global mutable state is the only option we really have that works in your case.

By the way, if you don't want to have it pre-defined, this also shouldn't take too much time:

```nushell
if ($env.SET_POSHCONTEXT? | is-not-empty) {
  do --env $env.SET_POSHCONTEXT
}
```

JanDeDobbeleer:
Thanks all for your contributions on this one. Amazing work.

@holopin-bot @fdncred helping

JanDeDobbeleer:
@holopin-bot @devyn helping

JanDeDobbeleer:
@all-contributors please add @Nibodhika for code,doc

This is pure magic 🪄! @holopin-bot @Nibodhika wizard

Issue #5090, opened 2024-06-15
Right-aligned prompts are not right-aligned in Bash/mintty on Windows


Bilge:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

The right-aligned portions of the prompt are never right-aligned; they sit somewhere in the middle of the screen. Moreover, they keep changing their alignment depending on the width of the left-aligned segments, which is weird.

![image](https://github.com/JanDeDobbeleer/oh-my-posh/assets/470626/3f6d811e-2435-46ca-80b6-e3a2af97e29f)

In case you are supposing this is a limitation with Bash/mintty, the same feature works fine in [Starship](https://starship.rs/).

### Theme

[powerlevel10k_rainbow](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/powerlevel10k_rainbow.omp.json)

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

bash

### Log output

```Shell
[DEBUG] 12:56:06.103 shell.go:Getenv:326 → mintty
[TRACE] 12:56:06.103 shell.go:Getenv(TERM_PROGRAM) - 519.2µs
[DEBUG] 12:56:06.103 shell.go:Getenv:326 → C:\Users\Bilge\AppData\Local
[TRACE] 12:56:06.103 shell.go:Getenv(LOCALAPPDATA) - 0s
[TRACE] 12:56:06.103 shell_windows.go:CachePath() - 0s
[DEBUG] 12:56:06.103 shell.go:Getenv:326 → C:/Users/Bilge/.p10k.omp.yaml
[TRACE] 12:56:06.103 shell.go:Getenv(POSH_THEME) - 0s
[DEBUG] 12:56:06.103 shell.go:Shell:601 → no shell name provided in flags, trying to detect it
[DEBUG] 12:56:06.109 shell.go:Shell:609 → process name: bash.exe
[TRACE] 12:56:06.109 shell.go:Shell() - 5.667ms
[DEBUG] 12:56:06.109 shell.go:resolveConfigPath:260 → Cygwin detected, using full path for config
[TRACE] 12:56:06.109 shell.go:resolveConfigPath() - 5.667ms
[TRACE] 12:56:06.109 shell.go:Init() - 6.1862ms
[TRACE] 12:56:06.109 shell.go:Flags() - 0s
[TRACE] 12:56:06.112 config.go:loadConfig() - 2.6119ms
[...]

PR #5102, opened 2024-06-17
fix(bash): only use COLUMNS when it's present

JanDeDobbeleer:
### Prerequisites

### Description

relates to #5090

### DIFF
--- a/src/shell/scripts/omp.bash
+++ b/src/shell/scripts/omp.bash
@@ -54,6 +54,11 @@ function _omp_hook() {
     local omp_stack_count=$((${#DIRSTACK[@]} - 1))
     local omp_elapsed=-1
     local no_exit_code="true"
+    local omp_columns=0
+
+    if [[ -n "$COLUMNS" ]]; then
+        omp_columns=$COLUMNS
+    fi
 
     if [[ -n "$omp_start_time" ]]; then
         local omp_now=$(::OMP:: get millis --shell=bash)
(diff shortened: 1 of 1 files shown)


src/shell/scripts/omp.bash
```diff
@@ -54,6 +54,11 @@ function _omp_hook() {
     local omp_stack_count=$((${#DIRSTACK[@]} - 1))
     local omp_elapsed=-1
     local no_exit_code="true"
+    local omp_columns=0
+
+    if [[ -n "$COLUMNS" ]]; then
```
Bilge, line 59: Just a note that `-n` is the default test, so there is no need to specify it. That is, `[[ -n $COLUMNS ]]` is the same as `[[ $COLUMNS ]]`. Moreover, since `[[` is a language construct (as opposed to `[`), it is not vulnerable to word splitting, so there is no need to quote if you don't want to (though there is no harm in doing so).

JanDeDobbeleer: I learned something today! Thanks 🙏🏻

src/shell/scripts/omp.bash
```diff
@@ -65,7 +70,7 @@ function _omp_hook() {
     set_poshcontext
     _set_posh_cursor_position
 
-    PS1="$(::OMP:: print primary --config="$POSH_THEME" --shell=bash --shell-version="$BASH_VERSION" --status="$ret" --pipestatus="${pipeStatus[*]}" --execution-time="$omp_elapsed" --stack-count="$omp_stack_count" --no-status="$no_exit_code" --terminal-width="$COLUMNS" | tr -d '\0')"
+    PS1="$(::OMP:: print primary --config="$POSH_THEME" --shell=bash --shell-version="$BASH_VERSION" --status="$ret" --pipestatus="${pipeStatus[*]}" --execution-time="$omp_elapsed" --stack-count="$omp_stack_count" --no-status="$no_exit_code" --terminal-width="$omp_columns" | tr -d '\0')"
```
Bilge, line 73: Note that there was not really any need to introduce this new variable at all. We could have just [used a default value](https://flokoe.github.io/bash-hackers-wiki/syntax/pe/#use-a-default-value) in the case that `COLUMNS` was not set, i.e. `${COLUMNS-0}` would work here.

PR #5127, opened 2024-06-19
Update transient.mdx to include initialization tip

kenmorse:
### Prerequisites

- [x ] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [x ] The commit message follows the [conventional commits][cc] guidelines.

### Description

Added tip about when to load configuration file if using transient prompt. If done too early, transient prompt won't work, at least with some shells.


JanDeDobbeleer:
I would add this context to the prompt configuration guide as it also affects tooltips etc.

PR #5132, opened 2024-06-20
fix(shell): prioritize `$?` for consistent status in Bash/Zsh

lewis-yeung:
### Prerequisites

### Description

In Bash (before v5.1) or Zsh, the last error code (`$?`) might not be consistent with the pipeline status. For example, a SIGINT (typically sent by a <kbd>Ctrl</kbd> + <kbd>C</kbd> input) sets `$?` to `130` which means a death by SIGINT, while the previous value of `$PIPESTATUS` (in Bash) or `$pipestatus` (in Zsh) persists. In Bash, this is [a bug which was fixed in v5.1][bash-changelog]. In Zsh, this still happens in the latest version (v5.9).

We can handle such cases consistently. When `$?` is not consistent with the last pipeline status code, we should use the value of `$?` as the argument for both `--status` and `--pipestatus` in an CLI call.

[bash-changelog]: https://git.savannah.gnu.org/cgit/bash.git/tree/CHANGES#n836

### DIFF
--- a/src/shell/scripts/omp.zsh
+++ b/src/shell/scripts/omp.zsh
@@ -13,7 +13,7 @@ function _set_posh_cursor_position() {
   # not supported in Midnight Commander
   # see https://github.com/JanDeDobbeleer/oh-my-posh/issues/3415
   if [[ "::CURSOR::" != "true" ]] || [[ -v MC_SID ]]; then
-      return
+    return
   fi
 
   local oldstty=$(stty -g)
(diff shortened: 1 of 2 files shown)


src/shell/scripts/omp.zsh
```diff
@@ -107,30 +111,30 @@ function _posh-tooltip() {
...
-    # Ctrl-C
-    if (( ret )); then
-        zle .send-break
-    else
-        # Enter
-        zle .accept-line
-    fi
-    return ret
+  if ((ret)); then
+    # TODO (fix): this is not equal to sending a SIGINT, since the status code ($?) is set to 1 instead of 130.
```
lewis-yeung, line 131: I left a comment here because the transient prompt support in Zsh overwrites the built-in `zle-line-init` function, which causes a SIGINT to set `$?` to `1` instead of `130`. I haven't found a fix yet since it is not a fatal issue, IMO. Hope we can resolve this in the future.

JanDeDobbeleer (approved):

PR #5201, opened 2024-07-04
feat(install): validate file signature before installation

JanDeDobbeleer:
### Prerequisites


src/upgrade/verify.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+
+	rsaPublicKey, err := loadPublicKey()
+	if err != nil {
+		return err
+	}
+
+	hashed := sha256.Sum256(binaryData)
+	err = rsa.VerifyPKCS1v15(rsaPublicKey, crypto.SHA256, hashed[:], signature)
```
kapsiR: Any reason why not using an ECC algorithm? 

When using RSA, the recommended scheme is RSA-PSS.  
[From the docs](https://pkg.go.dev/crypto/rsa):
> The original specification for encryption and signatures with RSA is PKCS #1 and the terms "RSA encryption" and "RSA signatures" by default refer to PKCS #1 version 1.5. **However, that specification has flaws and new designs should use version 2, usually called by just OAEP and PSS, where possible.**

kapsiR (approved):
LGTM 👍

Thanks for your efforts!

src/upgrade/verify.go
JanDeDobbeleer: Because I'm not at home in these things, so I need people like you help me out on this one :-) I'll have a look.

JanDeDobbeleer: @kapsiR seems like PSS and `golang` don't play together nicely. I'm out of options as it can't parse an RSA-PSS public key correctly.

kapsiR: What about using an ECC algorithm like [Ed25519](https://pkg.go.dev/crypto/ed25519) or [ECDSA](https://pkg.go.dev/crypto/ecdsa)?  

Does it have to be compatible with the OpenSSL CLI?
Do you need a PoC?

kapsiR: Since Go isn't my main language, I'd give it a try to provide a PoC implementation to gain some experience 😉

JanDeDobbeleer: Updated to `ed25519`

src/upgrade/verify.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	pubKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("parse public key: %v", err)
+	}
+
+	ed25519PubKey, ok := pubKey.(ed25519.PublicKey)
+	if !ok {
+		return nil, fmt.Errorf("public key is not RSA")
```
kapsiR: _nit:_ we should use another error message here: "invalid public key"

kapsiR (approved):

src/upgrade/verify.go
JanDeDobbeleer: Check. That's still an old message. Good catch.

PR #5214, opened 2024-07-05
feat: add support for fish mode

jcdickinson:
### Prerequisites

### Description

Adds a new mechanism for shells to pass variables to omp through a `--var key=value` argument. An attempt is made to parse these as an integer, a boolean, and falling back to string. The result is added to `ShellVar` for use in templates.

This is then used to create the `ShellVar.FISH__MODE` variable, which is set to the value of `$fish_bind_mode`.

### DIFF
--- a/src/cache/template.go
+++ b/src/cache/template.go
@@ -18,6 +18,7 @@ type Template struct {
 	Code          int
 	Env           map[string]string
 	Var           maps.Simple
+	ShellVar      maps.Simple
 	OS            string
 	WSL           bool
 	PromptCount   int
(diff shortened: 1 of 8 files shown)


JanDeDobbeleer:
@jcdickinson why not use the `set_poshcontext` [function](https://ohmyposh.dev/docs/configuration/templates#environment-variables) for this purpose? I feel this will become a never ending addition of shell variables people want to leverage, which will mostly be a specific, personal use-case.

jcdickinson:
Yeah I wasn't 100% sure this would be in line with the the project style - having a never-ending list of variables was basically  my intent.

PR #5218, opened 2024-07-06
chore:Devcontainer compatibility with podman

po1o:
### Prerequisites

### Description
Podman is stricter than Docker.
Modify devcontainer.json so that it is possible to start the devcontainer with podman. Also make sure that SELinux label are dropped within the container otherwise the worksapce is unusable. These changes remain compatible with Docker.

Note: podman needs the option --userns=keep-id. However this would make Docker break as it is an unsupported user mode. The solution is to add this option to /etc/containers/container.conf.


po1o:
I tested this with Docker on Fedora Linux. I don't have access to a windows machine to test on Windows.

JanDeDobbeleer:
@po1o tested on Windows for you.

Issue #5193, opened 2024-07-01
Incompatibility with zsh-autocomplete and a transient prompt


D3SOX:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

After installing Oh my Posh I can no longer navigate [zsh-autocomplete](https://github.com/marlonrichert/zsh-autocomplete) for example with the arrow keys

### Theme

```toml
#:schema https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json

version = 2
final_space = true
console_title_template = '{{ .Shell }} in {{ .Folder }}'

[[blocks]]
  type = 'prompt'
  alignment = 'left'
  newline = true

  [[blocks.segments]]
    type = 'path'
    style = 'plain'
    background = 'transparent'
    foreground = 'blue'
    template = '{{ .Path }}'

    [blocks.segments.properties]
      style = 'full'

  [[blocks.segments]]
    type = 'git'
    style = 'plain'
    foreground = 'p:grey'
    background = 'transparent'
    template = ' {{ .HEAD }}{{ if or (.Working.Changed) (.Staging.Changed) }}*{{ end }} <cyan>{{ if gt .Behind 0 }}⇣{{ end }}{{ if gt .Ahead 0 }}⇡{{ end }}</>'

    [blocks.segments.properties]
      branch_icon = ''
      commit_icon = '@'
      fetch_status = true

[[blocks]]
  type = 'rprompt'
  overflow = 'hidden'

  [[blocks.segments]]
    type = 'executiontime'
    style = 'plain'
    foreground = 'yellow'
    background = 'transparent'
    template = '{{ .FormattedMs }}'

    [blocks.segments.properties]
      threshold = 5000

[[blocks]]
  type = 'prompt'
  alignment = 'left'
  newline = true

  [[blocks.segments]]
    type = 'text'
    style = 'plain'
    foreground_templates = [
      "{{if gt .Code 0}}red{{end}}",
      "{{if eq .Code 0}}magenta{{end}}",
    ]
    background = 'transparent'
    template = '❯'

[transient_prompt]
  foreground_templates = [
    "{{if gt .Code 0}}red{{end}}",
    "{{if eq .Code 0}}magenta{{end}}",
  ]
  background = 'transparent'
  template = '❯ '

[secondary_prompt]
  foreground = 'magenta'
  background = 'transparent'
  template = '❯❯ '
```

[...]

Issue #5079, opened 2024-06-14
Space stops back-search with zsh


Salamandar:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

I've been trying out omp with zsh and there is one weird issue: when I trigger back-search (reverse search) with Ctrl-R and type some words, the search is "stopped" when I press the space bar.

### Theme

100% default configuration.

### What OS are you seeing the problem on?

Linux

### Which shell are you using?

zsh

### Log output

```Shell
Version: 21.5.0

Shell: zsh

Prompt:

 felix   ~    

Segments:

ConsoleTitle(true)                        -  60 ms
Session(true)                             -   6 ms
Path(true)                                -   6 ms
Git(false)                                -   2 ms
Root(false)                               -   0 ms
Status(true)                              -   7 ms
Node(false)                               -   0 ms
Go(false)                                 -   0 ms
Python(false)                             -   0 ms
Shell(true)                               -   0 ms
Time(true)                                -   0 ms

Run duration: 77.924155ms

Cache path: /home/felix/.cache/oh-my-posh

Config path: no --config set, using default built-in configuration

Logs:

[DEBUG] 14:07:22.651 shell.go:Getenv:326 → NO DATA
[TRACE] 14:07:22.651 shell.go:Getenv(TERM_PROGRAM) - 40.679µs
[DEBUG] 14:07:22.651 shell.go:Getenv:326 → NO DATA
[TRACE] 14:07:22.651 shell.go:Getenv(OMP_CACHE_DIR) - 2.444µs
[DEBUG] 14:07:22.651 shell.go:Getenv:326 → NO DATA
[TRACE] 14:07:22.651 shell.go:Getenv(XDG_CACHE_HOME) - 1.814µs
[TRACE] 14:07:22.651 shell_unix.go:CachePath() - 19.867µs
[DEBUG] 14:07:22.653 shell.go:Getenv:326 → NO DATA
[TRACE] 14:07:22.653 shell.go:Getenv(POSH_THEME) - 6.038µs
[DEBUG] 14:07:22.653 shell.go:resolveConfigPath:246 → No config set, fallback to default config
[TRACE] 14:07:22.653 shell.go:resolveConfigPath() - 53.202µs
[TRACE] 14:07:22.653 shell.go:Init() - 1.416656ms
[TRACE] 14:07:22.653 shell.go:Flags() - 38ns
[...]

PR #5219, opened 2024-07-06
fix: #5193 - decorate user-defined zsh widget

po1o:
### Prerequisites

### Description

Posh redefines zle-line-init instead of decorating it. This makes some other zsh plugins fail if they also rely on redefining it. This commit adds a helper function to decorate any user-defined widget and use it to decorate zle-line-init.

The logic in this commit is largely inspired by how the zsh-highlighting plugin does the widget wrapping.

This fixes issue #5193.

I tested manually in the dev container by installing zsh-autocomplete and trying with and without this change but I don't know how I could test this automatically. Let me know if you think of a way.

### DIFF
--- a/src/shell/scripts/omp.zsh
+++ b/src/shell/scripts/omp.zsh
@@ -40,6 +40,7 @@ function prompt_ohmyposh_preexec() {
   if [[ "::FTCS_MARKS::" = "true" ]]; then
     printf "\033]133;C\007"
   fi
+
   omp_start_time=$(::OMP:: get millis)
 }
 
@@ -49,18 +50,23 @@ function prompt_ohmyposh_precmd() {
   omp_stack_count=${#dirstack[@]}
   omp_elapsed=-1
   omp_no_exit_code="true"
+
   if [ $omp_start_time ]; then
     local omp_now=$(::OMP:: get millis --shell=zsh)
     omp_elapsed=$(($omp_now - $omp_start_time))
     omp_no_exit_code="false"
   fi
+
   if [[ "${omp_pipestatus_cache[-1]}" != "$omp_status_cache" ]]; then
     omp_pipestatus_cache=("$omp_status_cache")
   fi
+
   count=$((POSH_PROMPT_COUNT + 1))
   export POSH_PROMPT_COUNT=$count
+
   set_poshcontext
   _set_posh_cursor_position
+
   eval "$(::OMP:: print primary --config="$POSH_THEME" --status="$omp_status_cache" --pipestatus="${omp_pipestatus_cache[*]}" --execution-time="$omp_elapsed" --stack-count="$omp_stack_count" --eval --shell=zsh --shell-version="$ZSH_VERSION" --no-status="$omp_no_exit_code")"
   unset omp_start_time
 }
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@po1o that's amazing. Does this also mean we no longer need to fix the broken key bindings below?

po1o:
Oh yes, I hadn't realized there was this workaround.
I believe it's no longer needed indeed.

I'll test when time permits and if it works I will amend my PR

JanDeDobbeleer:
@all-contributors please add @po1o for code

This is pure magic 🪄! @holopin-bot @po1o wizard

lewis-yeung:
@JanDeDobbeleer After this change, when there is no user-defined `zle-line-init` before OMP, the transient prompt in Zsh will not work.

JanDeDobbeleer:
@aha. So the decorate function should check for that I suppose? I'm no expert in zsh (and that documentation is crap).

src/shell/scripts/omp.zsh
```diff
@@ -143,20 +154,43 @@ function enable_poshtooltips() {
...
+  case ${widgets[$cur_widget]:-""} in
+    # Already decorated: do nothing.
+    user:_posh-decorated-*);;
+
+    user:*)
+      zle -N $prefix-$cur_widget ${widgets[$cur_widget]#*:}
+      eval "_posh-decorated-${(q)prefix}-${(q)cur_widget}() { posh::call_widget ${(q)prefix}-${(q)cur_widget} ${(q)posh_widget} -- \"\$@\" }"
+      zle -N $cur_widget _posh-decorated-$prefix-$cur_widget;;
+
+    # For now, do not decorate if it's not a user:*
```
JanDeDobbeleer, line 187: @lewis-yeung so this should in fact just add it, right?

po1o:
I'll have a look at the comments tonight. I did not have the time to get back to this yet.

src/shell/scripts/omp.zsh
po1o: Yes - I'll add it.
Sending PR tonight. I had a zle-line-init left in my config when I tested it. I'll use zsh --no-rcs to test. That's my bad

po1o: I need to check with ${+widgets[$cur_widget]} if the original widget exists or not.
I will carve out some time tonight for this + testing the bindkey removal.

lewis-yeung: > so this should in fact just add it, right?

@JanDeDobbeleer Yes. And I wonder if we can utilize the decorating logic to resolve #5079. Will try a fix when I'm available.

lewis-yeung: > I wonder if we can utilize the decorating logic to resolve #5079.

Seems unrelated. 🤷🏻‍♂️

JanDeDobbeleer: Yes, that's about key bindings. More tricky that one I guess (or meeds a different approach).

lewis-yeung: > I wonder if we can utilize the decorating logic to resolve #5079.

@JanDeDobbeleer Exciting news: this is practical. I've just tested out a fix and it resolved #5079. I'll double check and make a PR in a few days.

PR #5257, opened 2024-07-11
fix(zsh): widget decoration when non-existent

po1o:
### Prerequisites

- [] Tests for the changes have been added (for bug fixes / features).

### Description

This commit fixes commit f72c1464abbf7e78d98e5a8c09e53c053d3cabbe (PR #5219). When the widget is undefined, we need to define it instead of decorating it.
This PR also makes the logic more robust and in case the decoration function is used in other cases, expand its usage.

### Tests

This has been tested manually by starting zsh with zsh -f to ensure there is not leftover in the config.

### DIFF
--- a/src/shell/scripts/omp.zsh
+++ b/src/shell/scripts/omp.zsh
@@ -156,12 +156,13 @@ function enable_poshtooltips() {
 
 # Helper function for posh::decorate_widget
 # It calls the posh function right after the original definition of the widget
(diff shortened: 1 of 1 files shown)


po1o:
Since in #5219, there were some discussions to use the decoration for more uses, I expanded it also to allow decoration of builtins.

JanDeDobbeleer (approved):

PR #5295, opened 2024-07-17
feat(pwsh): cache prompt for repainting

lewis-yeung:
### Prerequisites

### Description

A follow-up of #5070, making prompt behaviors in PowerShell consistent with other shells.


src/prompt/engine.go
```diff
@@ -500,3 +509,80 @@ func (e *Engine) adjustTrailingDiamondColorOverrides() {
...
+
+func (e *Engine) updatePromptCache(value *promptCache) {
+	cacheJSON, err := json.Marshal(value)
+	if err != nil {
+		return
+	}
+	e.Env.Cache().Set(cache.PROMPTCACHE, string(cacheJSON), 1440)
+}
+
+// New returns a prompt engine initialized with the
```
JanDeDobbeleer, line 541: why did you move this one?

src/prompt/primary.go
```diff
@@ -9,71 +9,71 @@ import (
...
-	var cancelNewline, didRender bool
-
 	needsPrimaryRPrompt := e.needsPrimaryRPrompt()
 
-	for i, block := range e.Config.Blocks {
-		// do not print a leading newline when we're at the first row and the prompt is cleared
-		if i == 0 {
-			row, _ := e.Env.CursorPosition()
-			cancelNewline = e.Env.Flags().Cleared || e.Env.Flags().PromptCount == 1 || row == 1
+	switch e.Env.Shell() {
```
JanDeDobbeleer: no convinced that a switch is the best use-case for this one, why did you do in that direction?

src/prompt/tooltip.go
```diff
@@ -56,6 +43,15 @@ func (e *Engine) Tooltip(tip string) string {
 
 	switch e.Env.Shell() {
 	case shell.PWSH, shell.PWSH5:
+		defer func() {
+			if e.fetchPromptCache() {
```
JanDeDobbeleer: I would rename that `CachePrompt()` as this isn't about fetching.

JanDeDobbeleer (changes requested):
A few comments

src/prompt/engine.go
lewis-yeung: Just an advice on the code organization. The function returns a new prompt engine and it's all about the "engine" logic, so there is no need to put it in a standalone source file, IMO. It's okay to revert this if you prefer the original way. :)

JanDeDobbeleer: @lewis-yeung I think we can argue either way. This is fine, I'll allow it 😃

src/prompt/primary.go
lewis-yeung: I'll reorganize the part of logic that adapts to PowerShell only.

src/prompt/tooltip.go
lewis-yeung: Not sure if I get your point. Do you mean this is not about "fetching cache from file"? Here it tries to check any existing prompt cache and fill corresponding fields. The return value indicates whether a prompt cache is available.

lewis-yeung: Renamed to `checkPromptCache` now.

PR #5305, opened 2024-07-18
feat(kubectl): add cache to kubectl

EladLeev:
### Prerequisites

### Description
Hey 👋 
The `kubectl` segment is a bit slow - both in command and file parsing mode. With a `kubeconfig` file consisting of 50+ clusters, each operation takes ~400ms (according to `oh-my-posh debug`).

Under the assumption that usually, you are making a few operations under the same Kube context before moving to a different one, I thought it would be nice to leverage the caching.
This PR will add a default cache of 1 minute, before checking/parsing the context again.

Users can disable it using `cache_timeout` - same as other segments.

Thank you for all the hard work! 💪 

### DIFF
--- a/src/segments/kubectl.go
+++ b/src/segments/kubectl.go
@@ -1,6 +1,8 @@
 package segments
 
 import (
+	"encoding/json"
+	"fmt"
 	"path/filepath"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/properties"
@@ -13,11 +15,14 @@ import (
 const (
 	ParseKubeConfig properties.Property = "parse_kubeconfig"
 	ContextAliases  properties.Property = "context_aliases"
+	kubectlCacheKey                     = "kubectl"
 )
 
 type Kubectl struct {
 	props properties.Properties
 	env   runtime.Environment
+
+	dirty bool
 
 	Context string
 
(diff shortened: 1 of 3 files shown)


EladLeev:
Fixed the golangci errors -- @JanDeDobbeleer wdyt?

JanDeDobbeleer:
@EladLeev just a question, why cache the CLI call when you can parse the cubeconfig and get the same data faster?

EladLeev:
Per my comment, it's not faster unfortunately 😞 
On big Kubeconfig files, it takes around 400ms per event

JanDeDobbeleer:
@all-contributors please add @EladLeev for code,doc

This is pure magic 🪄! @holopin-bot @EladLeev wizard

ahmetb:
@EladLeev @JanDeDobbeleer we should revert this. This started to break the behavior in version `23.3.2`.

Caching the kubeconfig file is _not_ a good idea. It should be parsed every time. There are tools like `kubectx` that are used by tens of thousands of people actively, and they rely on ctx changing in the prompt.

❗  This behavior is extremely dangerous as the stale prompt may cause people to run a command in the wrong context (e.g. run a destructive action in a production environment instead of a test environment).

EladLeev:
Hey @ahmetb - sorry for that, and thanks for the comment on the PR. 

I've opened a revert PR:
https://github.com/JanDeDobbeleer/oh-my-posh/pull/5389

JanDeDobbeleer:
@EladLeev @ahmetb hold on. The default just needs to be added in case it's enabled due to that. I'm working on adding caching on segment level anyways, which should be a conscious user defined decision.

ahmetb:
Thanks for the quick action. I think I'd make the caching opt-in (if someone has too many contexts in their ~/.kube/config), and not opt-out.

JanDeDobbeleer:
@ahmetb check, that's what I meant.

PR #5314, opened 2024-07-20
Multiple issues at the same time

JanDeDobbeleer:
### Prerequisites

### Description

- **fix(bash): correctly align multiline prompt**
- **refactor: simpler prompt caching logic**

### DIFF
--- a/src/cache/cache.go
+++ b/src/cache/cache.go
@@ -34,6 +34,13 @@ var (
 	TOGGLECACHE      = fmt.Sprintf("toggle_cache_%s", pid())
 	PROMPTCOUNTCACHE = fmt.Sprintf("prompt_count_cache_%s", pid())
 	ENGINECACHE      = fmt.Sprintf("engine_cache_%s", pid())
+	FONTLISTCACHE    = "font_list_cache"
+)
+
+const (
+	ONEDAY   = 1440
+	ONEWEEK  = 10080
+	ONEMONTH = 43200
 )
 
 type Entry struct {
--- a/src/cache/cache.go
+++ b/src/cache/cache.go
@@ -33,7 +33,7 @@ var (
 	TEMPLATECACHE    = fmt.Sprintf("template_cache_%s", pid())
 	TOGGLECACHE      = fmt.Sprintf("toggle_cache_%s", pid())
 	PROMPTCOUNTCACHE = fmt.Sprintf("prompt_count_cache_%s", pid())
-	PROMPTCACHE      = fmt.Sprintf("prompt_cache_%s", pid())
+	ENGINECACHE      = fmt.Sprintf("engine_cache_%s", pid())
 )
 
 type Entry struct {
--- a/src/prompt/engine.go
+++ b/src/prompt/engine.go
@@ -16,7 +16,7 @@ import (
 
 var cycle *color.Cycle = &color.Cycle{}
 
-type promptCache struct {
+type engineCache struct {
 	Prompt            string
 	CurrentLineLength int
 	RPrompt           string
@@ -36,7 +36,9 @@ type Engine struct {
 	activeSegment         *config.Segment
 	previousActiveSegment *config.Segment
 
-	promptCache *promptCache
+	engineCache *engineCache
+
+	cached bool
 }
 
 func (e *Engine) write(text string) {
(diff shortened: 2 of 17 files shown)


JanDeDobbeleer:
@lewis-yeung if you find a moment, care to review the refactor commit?

src/prompt/engine.go
```diff
@@ -510,32 +512,38 @@ func (e *Engine) adjustTrailingDiamondColorOverrides() {
...
+	e.write(e.engineCache.Prompt)
+	e.currentLineLength = e.engineCache.CurrentLineLength
+	e.rprompt = e.engineCache.RPrompt
+	e.rpromptLength = e.engineCache.RPromptLength
 
 	return true
 }
 
-func (e *Engine) updatePromptCache(value *promptCache) {
+func (e *Engine) updatePromptCache(value *engineCache) {
```
lewis-yeung: Should it also be renamed to `updateEngineCache` correspondingly?

src/prompt/primary.go
```diff
@@ -139,4 +119,15 @@ func (e *Engine) writePrimaryRPrompt() {
 	e.write(strings.Repeat(" ", space))
 	e.write(e.rprompt)
 	e.write(terminal.RestoreCursorPosition())
+
+	if e.cached {
```
lewis-yeung: This will never be `true`, since an assignment is missing after the engine is restored from cache.

src/prompt/engine.go
```diff
@@ -510,32 +512,38 @@ func (e *Engine) adjustTrailingDiamondColorOverrides() {
 	}
 }
 
-func (e *Engine) checkPromptCache() bool {
-	data, ok := e.Env.Cache().Get(cache.PROMPTCACHE)
+func (e *Engine) restoreEngineFromCache() bool {
+	if !e.Env.Flags().Cached {
```
lewis-yeung, line 516: This flag is for the primary prompt only. Once a tooltip is rendered, we always cache it for reuse, regardless of this flag. Since we don't pass a `true` to this flag when requesting a tooltip from OMP CLI, if you check it here, a tooltip will never be cached.

src/prompt/primary.go
```diff
@@ -139,4 +119,15 @@ func (e *Engine) writePrimaryRPrompt() {
...
+		return
+	}
+
+	e.updatePromptCache(&engineCache{
+		Prompt:            e.prompt.String(),
+		CurrentLineLength: e.currentLineLength,
+		RPrompt:           e.rprompt,
+		RPromptLength:     e.rpromptLength,
+	})
```
lewis-yeung: How about leaving this part in the `Primary` method, and before its return?

```go
if !useCache {
  e.updatePromptCache(&engineCache{
    Prompt:            e.prompt.String(),
    CurrentLineLength: e.currentLineLength,
    RPrompt:           e.rprompt,
    RPromptLength:     e.rpromptLength,
  })
}

return e.string()
```

src/cache/cache.go
```diff
@@ -33,7 +33,7 @@ var (
 	TEMPLATECACHE    = fmt.Sprintf("template_cache_%s", pid())
 	TOGGLECACHE      = fmt.Sprintf("toggle_cache_%s", pid())
 	PROMPTCOUNTCACHE = fmt.Sprintf("prompt_count_cache_%s", pid())
-	PROMPTCACHE      = fmt.Sprintf("prompt_cache_%s", pid())
+	ENGINECACHE      = fmt.Sprintf("prompt_cache_%s", pid())
```
lewis-yeung:
```suggestion
	ENGINECACHE      = fmt.Sprintf("engine_cache_%s", pid())
```

src/prompt/primary.go
JanDeDobbeleer: Good catch, it was there but I missed it after another refactor.

src/prompt/engine.go
JanDeDobbeleer: But should we do that for PowerShell then?

lewis-yeung: For PowerShell only? Yes, at least currently this approach of prompt caching is used in PowerShell only.

JanDeDobbeleer: I would let the CLI determine this rather than a check in the code. That way it's reusable whenever.

src/prompt/primary.go
JanDeDobbeleer: Agreed, it's more explicit that way.

src/shell/scripts/omp.bash
```diff
@@ -78,14 +78,8 @@ function _omp_hook() {
     PS1="$(::OMP:: print primary --config="$POSH_THEME" --shell=bash --shell-version="$BASH_VERSION" --status="$_omp_status_cache" --pipestatus="${_omp_pipestatus_cache[*]}" --execution-time="$_omp_elapsed" --stack-count="$_omp_stack_count" --no-status="$_omp_no_exit_code" --terminal-width="${COLUMNS-0}" | tr -d '\0')"
 }
 
-# rprompt
-function _omp_rprompt() {
-    ::OMP:: print right --config="$POSH_THEME" --shell=bash --shell-version="$BASH_VERSION" --status="$_omp_status_cache" --pipestatus="${_omp_pipestatus_cache[*]}" --execution-time="$_omp_elapsed" --stack-count="$_omp_stack_count" --no-status="$_omp_no_exit_code" --terminal-width="${COLUMNS-0}" | tr -d '\0'
-    return $_omp_status_cache
```
lewis-yeung, line 84: This line should be added to the bottom of the `_omp_hook` function (where it was originally located).

JanDeDobbeleer: @lewis-yeung I was unaware it was moved, again, good catch.

lewis-yeung:
Couldn't agree more. The maintenance can be much less painful after completely removing the right prompt support for Bash. 😆

src/shell/scripts/omp.bash
lewis-yeung: I remember this because it was moved by me. 😆

Issue #5332, opened 2024-07-23
Path segment always displays "~", regardless of the actual current directory


daanstolp:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

**Expected**: when I navigate to another directory, I expect the path segment to display the path to that directory.
**Actual**: the path segment always displays "~", regardless of the current working directory.

### Theme

I have tried this on the out-of-the-box **takuya** and **jdedobbeleer** themes. Both show the same issue.

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

```Shell
Version: 22.0.1

Shell: pwsh (7.4.3)

Prompt:

╭─ pwsh  D:\                                                                                                                                 11:04:38 
╰─❯

Segments:

ConsoleTitle(false)                       -   0 ms
Shell(true)                               -   1 ms
Path(true)                                -   2 ms
Time(true)                                -   0 ms
Text(true)                                -   0 ms
Status(true)                              -   0 ms

Run duration: 17.73ms

Cache path: C:\Users\daans\AppData\Local\oh-my-posh

Config path: C:\Users\daans\AppData\Local\Programs\oh-my-posh\themes\takuya.omp.json

Logs:

[DEBUG] 11:04:38.788 os.go:Getenv:309 → C:\Users\daans\AppData\Local
[TRACE] 11:04:38.788 os.go:Getenv(LOCALAPPDATA) - 507.8µs
[TRACE] 11:04:38.788 os_windows.go:CachePath() - 507.8µs
[DEBUG] 11:04:38.788 os.go:Getenv:309 → C:\Users\daans\AppData\Local\Programs\oh-my-posh\themes\takuya.omp.json
[TRACE] 11:04:38.788 os.go:Getenv(POSH_THEME) - 0s
[DEBUG] 11:04:38.788 os.go:Shell:584 → no shell name provided in flags, trying to detect it
[DEBUG] 11:04:38.794 os.go:Shell:592 → process name: pwsh.exe
[TRACE] 11:04:38.794 os.go:Shell() - 5.8922ms
[TRACE] 11:04:38.794 os.go:resolveConfigPath() - 5.8922ms
[TRACE] 11:04:38.794 os.go:Init() - 6.4ms
[TRACE] 11:04:38.794 os.go:Flags() - 0s
[TRACE] 11:04:38.795 load.go:loadConfig() - 679.5µs
[...]

PR #5333, opened 2024-07-23
fix(pwsh): remove prompt cache

JanDeDobbeleer:
### Prerequisites

### DIFF
--- a/src/cli/print.go
+++ b/src/cli/print.go
@@ -110,11 +110,13 @@ func init() {
 	printCmd.Flags().StringVar(&command, "command", "", "tooltip command")
 	printCmd.Flags().BoolVarP(&plain, "plain", "p", false, "plain text output (no ANSI)")
 	printCmd.Flags().BoolVar(&cleared, "cleared", false, "do we have a clear terminal or not")
-	printCmd.Flags().BoolVar(&cached, "cached", false, "use a cached prompt")
 	printCmd.Flags().BoolVar(&eval, "eval", false, "output the prompt for eval")
 	printCmd.Flags().IntVar(&column, "column", 0, "the column position of the cursor")
-	// Deprecated flags
+
+	// Deprecated flags, keep to not break CLI integration
 	printCmd.Flags().IntVarP(&status, "error", "e", 0, "last exit code")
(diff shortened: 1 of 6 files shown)


JanDeDobbeleer:
@lewis-yeung I debugged this from PowerShell and we never really use the case so I'm going back to the drawing board with this one before continuing. We did not restore, nor re-use anything as tooltips did not use the cache, we only spent time writing the context to file. 

What was the actual goal here, cache the prompt so we do not repaint? Because in PowerShell we're not repainting the prompt for tooltips, it insert the tooltip into the already printed prompt.

lewis-yeung:
@JanDeDobbeleer

> We did not restore, nor re-use anything as tooltips did not use the cache, we only spent time writing the context to file.

> What was the actual goal here, cache the prompt so we do not repaint? Because in PowerShell we're not repainting the prompt for tooltips, it insert the tooltip into the already printed prompt.

A tooltip should always be cached because an existing tooltip should be redrawn along with the whole prompt repaint (not rendering a new one), and the tooltip should be prioritized, regardless of whether a right prompt is there. We know that by default <kbd>Ctrl</kbd>+<kbd>L</kbd> clears the screen buffer and triggers a prompt repaint, however, a CLI call is always made to generate a new prompt in previous implementations, and without the prompt cache, an existing tooltip will disappear after the repaint. That's **NOT CONSISTENT** with the current behaviors in other shells.

I know the limitation that we cannot handle the case where the key handler of Enter is overridden (#5332), but I don't think caching prompts is useless or unnecessary. Since the basic state (working directory, command execution history, last status code, etc.) does not change (when merely clearing the screen buffer), redrawing the current prompt is more reasonable than rendering a new one, especially in some cases rendering a segment may be time-consuming (e.g., a `git` segment in a large sized Git repository). For <kbd>Ctrl</kbd>+<kbd>L</kbd> we don't want anything other than a clean screen buffer, right?

JanDeDobbeleer:
@lewis-yeung definitely. My question wasn't to challenge that we do it, more around why and how we do it so that it doesn't interfere with other functionality. I'll do a POC implementation on this if that's OK, one we feel comfortable with that doesn't introduce additional side-effects. There's value for sure.

PR #5342, opened 2024-07-25
feat(spotify): add cache

DeepSpace2:
### Prerequisites

- [x] Docs have been added/updated (for bug fixes / features) **(I will after the community agrees to add this functionality)** .

### Description

This is a WIP PR, used as a POC only for MacOS. Of course I plan to implement it for Windows and WSL as well, I just wanted to check if it's even something the community would like to have and get some feedback. 

Since adding the spotify segment slows down the shell I decided to implement a caching mechanism.
Acknowledging that most songs are longer than 2-3 minutes, a 1 minute cache can greatly improve the UX.
If someone is using spotify with podcasts then the benefit is even greater.

I tried to make as few changes as possible in the platform-specific implementation.

example config:

```yaml
      - type: spotify
        style: powerline
        powerline_symbol: ...
        foreground: "#000000"
        background: "#1DB954"
        template: "{{ .Icon }}{{ if ne .Status \"stopped\" }}{{ .Artist }} - {{ .Track }}{{ end }}"
        properties:
          cache_timeout: 1
```

### DIFF
--- a/src/segments/spotify_darwin.go
+++ b/src/segments/spotify_darwin.go
@@ -2,7 +2,20 @@
 
 package segments
 
+import (
+	"encoding/json"
+
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+)
+
+const spotifyCacheKey = "spotify_music_player"
+
 func (s *Spotify) Enabled() bool {
+	cacheTimeout := s.props.GetInt(properties.CacheTimeout, 0)
+	if cacheTimeout > 0 && s.getFromCache() {
+		return true
+	}
+
 	// Check if running
 	running := s.runAppleScriptCommand("application \"Spotify\" is running")
 	if running == "false" || running == "" {
@@ -23,7 +36,12 @@ func (s *Spotify) Enabled() bool {
 
 	s.Artist = s.runAppleScriptCommand("tell application \"Spotify\" to artist of current track as string")
 	s.Track = s.runAppleScriptCommand("tell application \"Spotify\" to name of current track as string")
+
 	s.resolveIcon()
+
+	if cacheTimeout > 0 {
+		s.setCache(cacheTimeout)
+	}
 
 	return true
 }
@@ -31,4 +49,29 @@ func (s *Spotify) Enabled() bool {
 func (s *Spotify) runAppleScriptCommand(command string) string {
 	val, _ := s.env.RunCommand("osascript", "-e", command)
 	return val
+}
+
+func (s *Spotify) getFromCache() bool {
+	str, found := s.env.Cache().Get(spotifyCacheKey)
+	if !found {
+		return false
+	}
+
+	var cachedMusicPlayer MusicPlayer
+	err := json.Unmarshal([]byte(str), &cachedMusicPlayer)
+	if err != nil {
+		return false
+	}
+
+	s.MusicPlayer = cachedMusicPlayer
+	return true
+}
+
+func (s *Spotify) setCache(cacheTimeout int) {
+	cache, err := json.Marshal(s.MusicPlayer)
+	if err != nil {
+		return
+	}
+
+	s.env.Cache().Set(spotifyCacheKey, string(cache), cacheTimeout)
 }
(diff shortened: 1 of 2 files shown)


DeepSpace2:
@JanDeDobbeleer @lewis-yeung

src/segments/spotify_darwin.go
```diff
@@ -2,7 +2,24 @@
 
 package segments
 
+import (
+	"github.com/jandedobbeleer/oh-my-posh/src/properties"
+)
+
 func (s *Spotify) Enabled() bool {
+	artist, track, status := "", "", ""
+	cacheTimeout := s.props.GetInt(properties.CacheTimeout, 0)
```
DeepSpace2, line 14: - Using 0 to disable cache by default, effectively making the spotify cache opt-in.
- Ultimately this piece of code should be in the generic `spotify.go`. Maybe in the `Init` function?

JanDeDobbeleer:
@DeepSpace2 windows loops window titles and doesn't need to leverage the utterly slow implementation we have on macOS. So that's not really needed there. WSL uses a similar tactic.

src/segments/spotify.go
```diff
@@ -30,6 +30,10 @@ const (
 	playing = "playing"
 	stopped = "stopped"
 	paused  = "paused"
+
+	spotifyArtistCacheKey = "spotify_artist_cache_key"
```
JanDeDobbeleer: I would cache the `MusicPlayer` struct instead of every individual field.

src/segments/spotify_darwin.go
JanDeDobbeleer: I would keep it here as per my general statement

src/segments/spotify.go
```diff
@@ -48,6 +52,19 @@ func (s *Spotify) resolveIcon() {
 	}
 }
 
+func (s *Spotify) getFromCache() (artist, track, status string) {
```
JanDeDobbeleer: this should return a boolean so you don't have to check every field downstream

src/segments/spotify_darwin.go
```diff
@@ -21,10 +38,18 @@ func (s *Spotify) Enabled() bool {
 		return false
 	}
 
-	s.Artist = s.runAppleScriptCommand("tell application \"Spotify\" to artist of current track as string")
-	s.Track = s.runAppleScriptCommand("tell application \"Spotify\" to name of current track as string")
-	s.resolveIcon()
+	if artist == "" || track == "" || status == "" {
```
JanDeDobbeleer: no need to use `if/else` if you restore from cache inside `getFromCache`

JanDeDobbeleer (changes requested):

DeepSpace2:
> @DeepSpace2 windows loops window titles and doesn't need to leverage the utterly slow implementation we have on macOS. So that's not really needed there. WSL uses a similar tactic.

@JanDeDobbeleer 
Cool, thanks for the input. So in that case I'll just implement everything in the macos implementation so it is clear it only applies there.

Do you have a preferred way for me to document that the cache property is only relevant on macos (or an existing case I can reference)?

src/segments/spotify_darwin.go
```diff
@@ -3,18 +3,17 @@
...
 func (s *Spotify) Enabled() bool {
-	artist, track, status := "", "", ""
 	cacheTimeout := s.props.GetInt(properties.CacheTimeout, 0)
 	if cacheTimeout > 0 {
-		artist, track, status = s.getFromCache()
-		if artist != "" && track != "" && status != "" {
-			s.Artist = artist
-			s.Track = track
-			s.Status = status
+		if loadedFromCache := s.getFromCache(); loadedFromCache {
```
JanDeDobbeleer:
```suggestion
		if s.getFromCache() {
```

JanDeDobbeleer:
@DeepSpace2 I made the changes, have a look.

DeepSpace2:
@JanDeDobbeleer Looks great, thanks.

JanDeDobbeleer:
@all-contributors please add @DeepSpace2 for code,doc

This is pure magic 🪄! @holopin-bot @DeepSpace2 wizard

PR #5365, opened 2024-07-27
feat(pwsh): pass original last execution status to Set-PoshContext

Soyvolon:
…r use in pwsh alias.

### Prerequisites

### Description

Updates the `Set-PoshContext` in `src/shell/scripts/omp.ps1` to include a parameter: `[bool]$originalStatus`. This allows for the original execution status to be passed into the `Set-PoshContext` function so it can be used during the execution of a configured alias (if applicable). The parameter is needed because `$?` will not show the proper output and `$script:` variables can't be accessed from the alias. Therefore, if a user wanted to do logic based on the execution status they need a different method of accessing the correct status value, which the parameter provides.

Notes:
- I don't know if any tests are needed - from what I could tell, the existing powershell tests cover this but I may be mistaken.
- This should not cause any problems with existing installs. Having an alias without the parameter present will not cause any errors.
- Rebased and fixed errors with the commit messages.

### DIFF
--- a/src/shell/scripts/omp.ps1
+++ b/src/shell/scripts/omp.ps1
@@ -113,7 +113,7 @@ New-Module -Name "oh-my-posh-core" -ScriptBlock {
         $stdoutTask.Result
     }
 
-    function Set-PoshContext {}
+    function Set-PoshContext([bool]$originalStatus) {}
 
     function Get-CleanPSWD {
         $pswd = $PWD.ToString()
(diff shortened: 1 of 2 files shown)


Soyvolon:
Fixed some problems with my rebase and the bad commit message. Without making a mess of the branches history this time - which is why it's in a new PR.

JanDeDobbeleer:
@all-contributors please add @Soyvolon for code,doc

This is pure magic 🪄! @holopin-bot @Soyvolon wizard

Issue #5362, opened 2024-07-27
wrong priory of config sources


harrhp:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

on latest version when POSH_THEME env variable is set it takes precedence over --config arg
```
$env:POSH_THEME="config_from_env"
oh-my-posh init -c "/config_from_arg" pwsh
```
output
```
(@(& 'C:/Users/Skobka/AppData/Local/Programs/oh-my-posh/bin/oh-my-posh.exe' init pwsh --config='config_from_env' --print) -join "`n") | Invoke-Expression
```

works on 22.1.0

### Theme

any

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

powershell

### Log output

```Shell
Version: 23.0.2

Shell: pwsh

Prompt:

εé╢ User εé░εé░ ε¬â ]8;;file:C:/Users/User\~]8;;\ εé░εé░ CONFIG ERROR εé┤ 7                                                                   in pwsh at 07:49:358

Segments:

ConsoleTitle(true)                        -   0 ms
Session(true)                             -   0 ms
Path(true)                                -   2 ms
Status(true)                              -   1 ms
Shell(true)                               -   0 ms
Time(true)                                -   0 ms

Run duration: 13.3601ms

Cache path: C:\Users\User\AppData\Local\oh-my-posh

Config path: config_from_env

Logs:

[38;2;191;207;240m[DEBUG] [38;2;156;231;201m07:49:35.596 [0mdebug.go:42[38;2;156;231;201m ΓåÆ[0m debug mode enabled
[DEBUG] 07:49:35.605 debug.go:42 ΓåÆ dlain mode enabled
[DEBUG] 07:49:35.605 os.go:Getenv:333 ΓåÆ C:\Users\User\AppData\Local
[TRACE] 07:49:35.605 os.go:Getenv(LOCALAPPDATA) - 0s
[TRACE] 07:49:35.605 os_windows.go:CachePath() - 0s
[DEBUG] 07:49:35.605 os.go:238 ΓåÆ loading cache file: C:\Users\User\AppData\Local\oh-my-posh\omp.cache
[TRACE] 07:49:35.605 os.go(C:\Users\User\AppData\Local\oh-my-posh\omp.cache) - 0s
[DEBUG] 07:49:35.605 os.go:Getenv:333 ΓåÆ C:\Users\User\AppData\Local
[...]

PR #5371, opened 2024-07-28
Resolve config path for `POSH_THEME` environment variable

thanhph111:
### Prerequisites

### Description

I have a specific use case where I need to set `POSH_THEME` from the client when SSHing into a server (using `SendEnv`). Due to differences in operating systems, the path for `POSH_THEME` needs to utilize `~` as HOME. I noticed that a recent update broke this setup.

While it is not recommended to overwrite `POSH_THEME` as mentioned in [this discussion](https://github.com/JanDeDobbeleer/oh-my-posh/issues/5362#issuecomment-2254157084), it is currently the only method I have to change it on the fly. Therefore, I would like to attempt resolving the path in `POSH_THEME` similar to how it is handled with the `--config` flag.

By the way, thank you for creating such amazing software and for considering this change.

### DIFF
--- a/src/runtime/os.go
+++ b/src/runtime/os.go
@@ -257,9 +257,44 @@ func (term *Terminal) Init() {
 func (term *Terminal) resolveConfigPath() {
 	defer term.Trace(time.Now())
 
+	processConfigPath := func(configString string) string {
+		if strings.HasPrefix(configString, "https://") {
+			filePath, err := config.Download(term.CachePath(), configString)
+			if err != nil {
+				term.Error(err)
+				return ""
+			}
+			return filePath
+		}
+
+		isCygwin := func() bool {
+			return term.Platform() == WINDOWS && len(term.Getenv("OSTYPE")) > 0
+		}
+
+		// Cygwin path always needs the full path as we're on Windows but not really.
+		// Doing filepath actions will convert it to a Windows path and break the init script.
+		if isCygwin() {
+			term.Debug("cygwin detected, using full path for config")
+			return configString
+		}
+
+		if strings.HasPrefix(configString, "~") {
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@thanhph111 this isn't how we can solve this. `POSH_THEME` isn't to be accessed, it's currently the only way we can avoid doing more than we need to as `init` will resolve it so we can reuse the actual path for every prompt generation. Why not provide your own environment variable you add to `init`?

EDIT: this is what we did before, and it saves a few milliseconds off the execution. And I don't want to sacrifice that again.

JanDeDobbeleer:
@thanhph111 I realized this could be trickier than originally designed. What if I adjust the logic so `POSH_THEME` stays internal, but another environment variable can be leveraged for this use-case? As in, that one would always follow the same logic we have for the `--config` flag (and is documented).

I wanted to rename `POSH_THEME` to `POSH_CONFIG_PATH` anyways, and if we have a `POSH_CONFIG_PATH_OVERRIDE` that will solve your challenge (and others').

thanhph111:
@JanDeDobbeleer That would be great. I think we can design the command to reflect changes in `POSH_CONFIG_PATH_OVERRIDE` with minimal updates, making it more efficient than rerunning `init --config`. The priority order could be `--config` > `POSH_CONFIG_PATH_OVERRIDE` > `POSH_CONFIG_PATH`.

However, if `POSH_CONFIG_PATH` is intended for internal use, I think it might confuse users who try to override it. To avoid this, we could consider resetting `POSH_CONFIG_PATH` in every iteration.

JanDeDobbeleer:
@thanhph111 if we rename it to _POSH_CONFIG_PATH, we're good to go. That one implies internal.

PR #5385, opened 2024-07-30
Nix Shell Segment

yash-garg:
### Prerequisites

### Description

This adds a segment for [Nix Shell](https://nixos.org/guides/nix-pills/developing-with-nix-shell.html).

Please review and let me know if there are any improvements needed in the code or documentation.

### DIFF
--- a/src/config/segment_types.go
+++ b/src/config/segment_types.go
@@ -27,7 +27,6 @@ const (
 	Accordion SegmentStyle = "accordion"
 	// Diamond writes the prompt shaped with a leading and trailing symbol
 	Diamond SegmentStyle = "diamond"
-
 	// ANGULAR writes which angular cli version us currently active
 	ANGULAR SegmentType = "angular"
 	// ARGOCD writes the current argocd context
@@ -128,6 +127,8 @@ const (
 	NPM SegmentType = "npm"
 	// NX writes which Nx version us currently active
 	NX SegmentType = "nx"
+	// NIXSHELL writes the active nix shell details
+	NIXSHELL SegmentType = "nix-shell"
 	// OCAML writes the active Ocaml version
 	OCAML SegmentType = "ocaml"
 	// OS write os specific icon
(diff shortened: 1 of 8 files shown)


src/config/segment_types.go
```diff
@@ -128,6 +128,8 @@ const (
 	NPM SegmentType = "npm"
 	// NX writes which Nx version us currently active
 	NX SegmentType = "nx"
+	// NIX writes the active nix shell details
+	NixShell SegmentType = "nix_shell"
```
JanDeDobbeleer:
```suggestion
	NIX SegmentType = "nix"
```

src/segments/nixshell.go
```diff
@@ -0,0 +1,67 @@
...
+	default:
+		if n.InNewNixShell() {
+			return UNKNOWN
+		}
+		return NONE
+	}
+}
+
+// Hack to detect if we're in a `nix shell` (in contrast to a `nix-shell`).
+func (n *NixShell) InNewNixShell() bool {
```
JanDeDobbeleer, line 44: I don't think this offer any guarantees, it assumes the availability of one `PATH` entry implies we're in a nix shell whereas you could still have, and be running, `nix-shell` if that one's also on your `PATH`. Also, why is this relevant? If the shell type is there (`IN_NIX_SHELL`), do you need to do this validation?

yash-garg: @JanDeDobbeleer so there are two types of shells by design, `nix shell` and `nix-shell`, some discussion can be found [here](https://discourse.nixos.org/t/nix-shell-nix-shell-and-nix-develop/25964).

When we use `nix shell` the `IN_NIX_SHELL` variable isn't set, see https://github.com/NixOS/nix/issues/6677 so we check the `PATH` for it to verify.

https://starship.rs/ also uses the same approach for detecting it.

JanDeDobbeleer: @Yash-Garg can you add that issue to the inline comments as well. I don't understand how validating a PATH entry even closely qualifies as implying being in a certain shell environment, but I guess that's just me.

yash-garg: @JanDeDobbeleer sure, i've added the comments for it.

JanDeDobbeleer:
@all-contributors please add @Yash-Garg for code,design,doc

This is pure magic 🪄! @holopin-bot @Yash-Garg wizard

PR #5389, opened 2024-07-30
Revert "feat(kubectl): add cache to kubectl"

EladLeev:
Reverts JanDeDobbeleer/oh-my-posh#5305

### DIFF
--- a/src/segments/kubectl.go
+++ b/src/segments/kubectl.go
@@ -1,8 +1,6 @@
 package segments
 
 import (
-	"encoding/json"
-	"fmt"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


JanDeDobbeleer:
@EladLeev set the default cache timeout to 0 and we're good to go. That will make this behave opt-in rather than opt-out. Should've realized this when reviewing.

EladLeev:
fair enough, opened this one for a quick revert in case you want to do that.
Here's the correct one - 
https://github.com/JanDeDobbeleer/oh-my-posh/pull/5390

PR #5457, opened 2024-08-16
feat: Add 'random' template

oliviaBahr:
### Prerequisites

### Description
Add a segment called Random which selects a random string from a list to display.

I wasn't sure which category to put this segment in, so I made a new one called `Miscellaneous` (We could also call it `Other`). We could also be more specific and call it `Custom` or something but I'm not sure.
If we do add this category, I would probably move the `Text` segment there as well. Happy to add that to this pr or make a new one.

I couldn't figure out how to get the test config to persist in my env, so I've only actually seen it work when calling
`oh-my-posh print primary --config`
But so long as .Enabled() is called every time the prompt is rendered then it will work. I assume .Enabled() is called every time, because that's where the `Time` segment updates its info.


JanDeDobbeleer (changes requested):
@oliviaBahr thank you so much for this PR and your patience in me getting back to you. I feel like this would be a perfect [template function](https://github.com/JanDeDobbeleer/oh-my-posh/blob/ebe2cca2329243db26dca884e7eca0a8886af062/src/template/func_map.go#L10) instead of a segment as that's easy to use and leverage cross segment. We could even make the signature work with any type (if that's possible).

JanDeDobbeleer:
@oliviaBahr this is absolutely perfect. I have one remark, but overall 👍

src/template/random.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+	"math/rand"
+	"reflect"
+)
+
+func random(list interface{}) (string, error) {
+	v := reflect.ValueOf(list)
+	if v.Kind() != reflect.Slice && v.Kind() != reflect.Array {
+		return "", errors.New("input must be a slice or array")
+	}
```
JanDeDobbeleer, line 15: Can you add a newline here

src/template/random.go
```diff
@@ -0,0 +1,19 @@
...
+	if v.Len() == 0 {
+		return "", errors.New("input slice or array is empty")
+	}
```
JanDeDobbeleer, line 19: Can you add a newline here?

JanDeDobbeleer (changes requested):

oliviaBahr:
#5644

PR #5466, opened 2024-08-18
fix(bash): respect `PROMPT_COMMAND` as an array

lewis-yeung:
### Prerequisites

### Description

According to [descriptions about the `PROMPT_COMMAND` variable][bash-manual-prompt-command] in *Bash Reference Manual*:

> If this variable is set, and is an array, the value of each set element is interpreted as a command to execute before printing the primary prompt (`$PS1`). If this is set but not an array variable, its value is used as a command to execute instead.

the variable can be an array, so we should always respect `PROMPT_COMMAND` as an array when installing the OMP hook.

[bash-manual-prompt-command]: https://www.gnu.org/software/bash/manual/bash.html#index-PROMPT_005fCOMMAND

### DIFF
--- a/src/shell/scripts/omp.bash
+++ b/src/shell/scripts/omp.bash
@@ -35,7 +35,7 @@ function _omp_set_cursor_position() {
 
     local COL
     local ROW
-    IFS=';' read -sdR -p $'\E[6n' ROW COL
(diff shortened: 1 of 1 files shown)


lewis-yeung:
~Needs double check...~ We can move on now.

PR #5473, opened 2024-08-19
Revert broken commits

JanDeDobbeleer:
### Prerequisites

- [x] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [x] The commit message follows the [conventional commits][cc] guidelines.
- [x] Tests for the changes have been added (for bug fixes / features).
- [x] Docs have been added/updated (for bug fixes / features).

### Description

@lewis-yeung as I have no idea which commit really introduces the issue (or due to both), I’m reverting both of them as I can’t test.

[CONTRIBUTING.md]: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/CONTRIBUTING.md
[cc]: https://www.conventionalcommits.org/en/v1.0.0/#summary

PR #5474, opened 2024-08-19
Reintroduce changes for Bash with rework

lewis-yeung:
### Prerequisites

### Description

~Sadly, I failed to find a `preexec`-like solution for Bash other than the original `PS0` trick. It's worth noting that overriding the `DEBUG` trap in Bash is more hacky and will break something like [bash-preexec][bash-preexec]. So we should always enable the `promptvars` shell option as opposed to the previous buggy implementation, and escape special characters in a prompt string.~

Finally, with some rework and careful testing, we can safely reintroduce two previously reverted changes (in #5473).

[bash-preexec]: https://github.com/rcaloras/bash-preexec

### DIFF
--- a/src/shell/formats.go
+++ b/src/shell/formats.go
@@ -16,7 +16,7 @@ type Formats struct {
 	Osc7  string
 	Osc51 string
 
-	EscapeSequences map[rune]rune
+	EscapeSequences map[rune]string
 
 	HyperlinkStart  string
 	HyperlinkCenter string
@@ -50,9 +50,8 @@ func GetFormats(shell string) *Formats {
 			ITermPromptMark:       "\\[$(iterm2_prompt_mark)\\]",
 			ITermCurrentDir:       "\\[\x1b]1337;CurrentDir=%s\x07\\]",
 			ITermRemoteHost:       "\\[\x1b]1337;RemoteHost=%s@%s\x07\\]",
-			EscapeSequences: map[rune]rune{
-				96: 92, // backtick
-				92: 92, // backslash
+			EscapeSequences: map[rune]string{
+				'\\': `\\`,
 			},
 		}
 	case ZSH, TCSH:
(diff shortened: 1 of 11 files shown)


lewis-yeung:
This time, to avoid introducing other bugs, I'll further review the whole prompt logic before the changes are merged.

src/terminal/writer.go
```diff
@@ -392,17 +400,18 @@ func write(s rune) {
...
-	if !Interactive {
-		for special, escape := range formats.EscapeSequences {
-			if s == special && lastRune != escape {
-				builder.WriteRune(escape)
-			}
+	// UNSOLVABLE: When "Interactive" is true, the prompt length calculation in Bash/Zsh can be wrong, since the final string expansion is done by shells.
+	length += runewidth.RuneWidth(s)
+	// length += utf8.RuneCountInString(string(s))
+
+	if !Interactive && !Plain {
```
lewis-yeung, line 407: A problem here: when `Interactive` is `true`, the escape will not be done by OMP here, however, a series of backslashes will be half eaten by the shell escape logic.

src/shell/formats.go
```diff
@@ -50,9 +50,10 @@ func GetFormats(shell string) *Formats {
 			ITermPromptMark:       "\\[$(iterm2_prompt_mark)\\]",
 			ITermCurrentDir:       "\\[\x1b]1337;CurrentDir=%s\x07\\]",
 			ITermRemoteHost:       "\\[\x1b]1337;RemoteHost=%s@%s\x07\\]",
-			EscapeSequences: map[rune]rune{
-				96: 92, // backtick
-				92: 92, // backslash
+			EscapeSequences: map[rune]string{
+				'\\': `\\\\`,
+				'`':  "\\`",
+				'$':  `\$`,
```
lewis-yeung: These two characters should be escaped regardless of `terminal.Interactive`. 🤦🏻

lewis-yeung:
Things always go tricky when falling on Bash...

lewis-yeung:
GOOD NEWS! Found a solution where we need only a simple trick on the shell side, without handling the annoying escape in OMP. Will update the PR later.

lewis-yeung:
Ready for review now.

JanDeDobbeleer:
@lewis-yeung looking at this one when I get back from holidays.

PR #5478, opened 2024-08-20
Feature (OWM) - Add ability to set the location parameter by an Environmental Variable

JamesAndrewJackson13:
### Prerequisites

### Description

Within the OWM segment, add the ability to set the location property via an environmental variable (just like how the API key can be set that way). 

This allows for the location to be dynamically set, which allows for the location to be updated depending on where the user is. Useful for anyone running oh-my-posh on a laptop and changes location often. Without this change, if you want to do something like this you have to manually insert the location into the config file right before passing it to oh-my-posh (at least that's how I'm doing it; may not be the best method). 

### Note

For anyone curious about how you can get your location within PowerShell (on Windows), [this stackoverflow post](https://stackoverflow.com/questions/46287792/powershell-getting-gps-coordinates-in-windows-10-using-windows-location-api) shows how to use the location-api. That gets your lat-long position. You'll then need to send that to a reverseGeocode api of some kind to get the city name, state code and country code (I was using Bing Maps API to do it, since it was free, but since that is getting shut down next year I've moved to Azure Maps). 

### DIFF
--- a/src/segments/owm.go
+++ b/src/segments/owm.go
@@ -32,8 +32,10 @@ const (
 	CacheKeyResponse string = "owm_response"
 	// CacheKeyURL key used when caching the url responsible for the response
 	CacheKeyURL string = "owm_url"
-
-	PoshOWMAPIKey = "POSH_OWM_API_KEY"
+	// Environmental variable to dynamically set the Open Map API key
+	OWMAPIKey string = "POSH_OWM_API_KEY"
+	// Environmental variable to dynamically set the location string
+	OWMLocationKey string = "POSH_OWM_LOCATION"
 )
 
 type weather struct {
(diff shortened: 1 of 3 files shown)


src/segments/owm.go
```diff
@@ -88,14 +90,21 @@ func (d *Owm) getResult() (*owmDataResponse, error) {
...
 	}
 
-	location := d.props.GetString(Location, "De Bilt,NL")
+	if len(apikey) == 0 {
+		return nil, errors.New("no api key found")
+	}
 
-	location = url.QueryEscape(location)
+	location := d.props.GetString(Location, "De Bilt,NL")
+	if len(location) == 0 {
```
JanDeDobbeleer: This will never be true, right? We default to a value when fetching the `location` from the `properties` which implies `location` will always have a value.

JanDeDobbeleer: @JamesAndrewJackson13 any update on this remark?

JanDeDobbeleer:
@all-contributors please add @JamesAndrewJackson13 for code,doc

This is pure magic 🪄! @holopin-bot @JamesAndrewJackson13 wizard

PR #5498, opened 2024-08-22
feat(runtime/http): Use standard library proxy resolution in HTTP client

thep0y:
Replace custom proxy resolution logic with the built-in http.ProxyFromEnvironment. This leverages the standard library to handle proxy configuration, simplifying our codebase and ensuring compatibility with environment variable-based proxy configurations.

The original code did not include test cases for the proxy functionality, and there is no related content in the documentation either. Therefore, the documentation regarding proxy features may need to be updated primarily by the main contributors.

### DIFF
--- a/src/runtime/http/http.go
+++ b/src/runtime/http/http.go
@@ -3,8 +3,6 @@ package http
 import (
 	"net"
 	"net/http"
-	"net/url"
-	"os"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@thep0y we initially had this, which then caused issues. But I'm in favor of reverting and seeing if this hot improved in later go versions.

JanDeDobbeleer:
@all-contributors please add @thep0y for code

This is pure magic 🪄! @holopin-bot @thep0y wizard

PR #5547, opened 2024-09-02
fix(elvish): prevent cutting off a right-aligned block

lewis-yeung:
### Prerequisites

### Description

In Elvish, a prompt line will always wrap when the cursor reaches the right-most cell on the terminal screen. We have to reduce the terminal width by 1 (just like the patch for the bleed bug in PowerShell) and use ANSI control sequences whenever we need a line feed (`\n`), so a right-aligned block will not be broken. BTW, the right prompt always works correctly because it's natively supported by the shell.

### DIFF
--- a/src/prompt/engine.go
+++ b/src/prompt/engine.go
@@ -108,8 +108,9 @@ func (e *Engine) pwd() {
 }
 
 func (e *Engine) getNewline() string {
-	// WARP terminal will remove \n from the prompt, so we hack a newline in
-	if e.isWarp() {
+	// WARP terminal will remove \n from the prompt, so we hack a newline in.
+	// For Elvish, we do this to prevent cutting off a right-aligned block.
+	if e.isWarp() || e.Env.Shell() == shell.ELVISH {
 		return terminal.LineBreak()
 	}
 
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer (approved):

JanDeDobbeleer:
Thanks for submitting a PR to the project!

In order to review and merge PRs most efficiently, we require that all PRs grant maintainer edit access before we review them. For information on how to do this, [see the documentation](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

JanDeDobbeleer:
Thanks for submitting a PR to the project!

In order to review and merge PRs most efficiently, we require that all PRs grant maintainer edit access before we review them. For information on how to do this, [see the documentation](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

lewis-yeung:
I have to rework this. The behavior of Elvish wrapping at the end of a prompt line is inconsistent across different platforms (tested on Windows and WSL). On WSL (maybe also regular Linux), the newline rendering of prompts in Elvish seems broken because of this change, but I haven't found the root cause yet. Further debugging/testing is needed. The shell-specific is somewhat tricky, but no worse than Tcsh which is weird in so many ways. 😅

JanDeDobbeleer:
@lewis-yeung I wonder if we should keep supporting these shells though. They're nice and not really feature rich. I only added them for feature parity with starship, but they're actually terrible.

lewis-yeung:
@JanDeDobbeleer We just keep basic support (at least for a primary prompt), which is not guaranteed to be full-featured. After all, the update frequency of feature support for a shell is positively correlated with its user base.

lewis-yeung:
Follow-up work has been done in #5614.

PR #5592, opened 2024-09-09
feat(git): add lists of commit's head, tag and remote references

kizivat:
### Prerequisites

### Description

Add `.Commit.Refs` which contain lists of head, remote and tag references of the given commit.

I updated the [`log` command](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segments/git.go#L214) at  to include the references.

It adds a new line to the output, starting with `rf:`.
Using  the [`%D` placeholder](https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-emDem) and [`--decorate=full` option](https://git-scm.com/docs/git-log#Documentation/git-log.txt---decorateshortfullautono) to print the references.

Those are then parsed based on the `/refs/<type>` prefix to get the

```go
type Refs struct {
	Heads   []string
	Tags    []string
	Remotes []string
}
```

references.

I'm not sure if we're ok with an array of strings being an output of these. But I found it useful to be able to `join` those in the template myself.

Feel free to suggest different approach.

### DIFF
--- a/src/segments/git.go
+++ b/src/segments/git.go
@@ -22,6 +22,13 @@ type Commit struct {
 	Subject   string
 	Timestamp time.Time
 	Sha       string
+	Refs      *Refs
+}
+
+type Refs struct {
+	Heads   []string
+	Tags    []string
+	Remotes []string
 }
 
 type User struct {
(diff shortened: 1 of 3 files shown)


src/segments/git.go
```diff
@@ -237,6 +245,25 @@ func (g *Git) Commit() *Commit {
 			g.commit.Subject = line
 		case "ha:":
 			g.commit.Sha = line
+		case "rf:":
+			refs := strings.Split(line, ", ")
+			for _, ref := range refs {
+				ref = strings.TrimSpace(ref)
+				switch {
+				case ref == "HEAD":
+					// skip "HEAD"
```
JanDeDobbeleer: maybe add a continue here rather than empty case

JanDeDobbeleer (changes requested):

src/segments/git.go
kizivat: pretty new to Go so just checking - won't `continue` unnecessarily check the the next case condition?

kizivat: oh ok forget that, of course I could check myself and I did - it won't

kizivat: Did so, also changed the case's condition to ignore remotes' `HEAD` references as well (not just local `HEAD`)

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @kizivat for code,doc

This is pure magic 🪄! @holopin-bot @kizivat wizard

PR #5623, opened 2024-09-15
chore(lint): enhance commitlint rules

ClxUne09:
### Prerequisites

### Description

- Updated commitlint rules to adjust allowed types and line length.
- Added 'style' to the list of commit types.
- Sorted commit types alphabetically.


.commitlintrc.yml
```diff
@@ -2,32 +2,79 @@
...
+              **Example**: `test: add unit tests for user service`
+
   `theme`:    Work on or add a theme.
-  `refactor`: Rewrites code without feature, performance or bug changes.
-  `perf`:     Improves performance.
-  `test`:     Adds or modifies tests.
-  `revert`:   Changes that reverting other changes
+  
+              **Usage**: Use for theme-related changes.
+              **Example**: `theme: add new theme`
```
JanDeDobbeleer: adding a new theme isn't allowed, you can only edit existing ones. SO best to reflect that in the help as well.

ClxUne09: > @JanDeDobbeleer
adding a new theme isn't allowed, you can only edit existing ones. SO best to reflect that in the help as well.

I’m not quite sure what you mean. Are you suggesting I should completely change the help section, including the examples? Or are you referring to modifying the `theme` type and its help description? Or is it about removing the new `style` type I added? I believe the `style` type is somewhat necessary, as it focuses on code style changes without altering functionality, which differs slightly from `refactor`.

JanDeDobbeleer: The example states `theme: add new theme` but that isn't allowed. So it should be something like `theme(jandedobbeleer): update git branch icon`.

JanDeDobbeleer:
@ClxUne09 how can I print the help text? I tried to find where this information is displayed to the user, but that doesn't seem to render anywhere.

ClxUne09:
> @JanDeDobbeleer
> 
> \@ClxUne09 how can I print the help text? I tried to find where this information is displayed to the user, but that doesn't seem to render anywhere.

I believe the `help` field in the `commitlint` configuration doesn’t print because it’s just a comment for reference. It won’t show up for users during commit linting. They’ll only see errors or warnings about their commit messages, not the help text. To make this info accessible, we’d need to document it elsewhere or whip up a separate script.

JanDeDobbeleer:
@ClxUne09 In that case it makes no sense to have this added here. I would create a new page in the contribution guide on the docs. We can then use that page as a reference in the PR template as well.

JanDeDobbeleer (approved):

Issue #5813, opened 2024-11-05
Inconsistent/random .Segments.Path.Writable result with v24+


stanio:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

After upgrading from v23.20.3 → v24.0.2, and currently on v24.0.4, I've noticed a `.Segments.Path.Writable` template condition:

```json
        {
          "style": "diamond",
          "leading_diamond": "\ue0b6",
          "trailing_diamond": "\ue0b0",
          "template": "{{ if .WSL }}WSL at {{ end }}{{.Icon}} ",
          "foreground": "black",
          "background": "magenta",
          "type": "os",
          "background_templates": [
            "{{ if .Root }}red{{ end }}",
            "{{ if not .Segments.Path.Writable }}white{{ end }}"
          ]
        },
```

stopped working.  More accurately it fails to indicate a writable directory more often than it works as expected (white indicates non-writable directory, while purple is normal/writable):

Cmd | Bash
---|---
![cmd](https://github.com/user-attachments/assets/2de49002-5b94-46d9-aa5a-34dbb890eeeb) | ![bash](https://github.com/user-attachments/assets/95045a43-1366-493e-8581-07de33d7297d)

### Theme

<details>
<summary>Custom</summary>

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json",
  "blocks": [
    {
      "type": "prompt",
      "alignment": "right",
      "segments": [
        {
          "properties": {
            "style": "round"
          },
          "style": "plain",
          "template": "{{ .FormattedMs }} | ",
          "foreground": "darkGray",
          "background": "",
          "type": "executiontime"
        },
        {
          "style": "plain",
          "template": "{{ if .SSHSession }}\uf817 {{ end }}{{ .UserName }}@{{ .HostName }}",
          "foreground": "green",
          "background": "",
          "type": "session"
        },
        {
          "properties": {
            "root_icon": "root"
          },
          "style": "plain",
          "template": " as \uf0e7",
          "foreground": "red",
          "background": "",
          "type": "root"
        }
      ]
    },
    {
      "type": "prompt",
      "alignment": "left",
      "segments": [
        {
          "style": "diamond",
          "leading_diamond": "\ue0b6",
          "trailing_diamond": "\ue0b0",
[...]

PR #5859, opened 2024-11-08
feat(language): allow tweaking version cache

JanDeDobbeleer:
### Prerequisites


website/docs/segments/languages/zig.mdx
```diff
@@ -19,7 +19,8 @@ import Config from "@site/src/components/Config.js";
     powerline_symbol: "\ue0b0",
     foreground: "#342311",
     background: "#ffad55",
-    template: " \ue6a9 {{ if .Error }}{{ .Error }}{{ else }}{{ .Full }}{{ end }} ",
+    template:
+      " \ue6a9 {{ if .Error }}{{ .Error }}{{ else }}{{ .Full }}{{ end }} ",
```
lewis-yeung, line 23: @JanDeDobbeleer Is this newline added by mistake?

JanDeDobbeleer: Seems so, thanks!

JanDeDobbeleer: ah no, it's the formatter that does this.

JanDeDobbeleer:
@lewis-yeung `.Text` can never be taken into account for cross template properties for primary and right, only for extra prompts. The reason is that the execution happens in parallel and writing sequentially. The check looks at if the segment has been executed in case there's a need in a template, but `.Text` is only available after writing, and thus after that check happens, as needs is a precondition for resolving `.Text`. Theoretically a check could be built-in that validates the presence of `.Text` and its value, but that would only work sequentially and in my opinion, that logic isn't worth my while as I don't really see the use case in primary and right.

lewis-yeung:
@JanDeDobbeleer Got it. I was taking that as an example to reproduce the issue, but not using it in my own config.

JanDeDobbeleer:
@lewis-yeung It was a really good catch, but that should be the only case where this behavior occurs, which I can live with as "by design". I want to keep that part as simple as possible. If you notice any other behavior, let me know!

lewis-yeung:
@JanDeDobbeleer Unfortunately, reverting the fix brought back the issue of #2885. It occurs not only when cross-referencing the `.Text` of a segment, but **any property**.

JanDeDobbeleer:
@lewis-yeung I made sure to test this with cross segment properties. I can double check, but this shouldn't be the case though.

lewis-yeung:
@JanDeDobbeleer It should work in an extra prompt, right? However, I always see the issue when using a cross segment property in the transient prompt, but not with the previous fix of ce027593b1bc397b0b9c65749c0020c900466c00.

JanDeDobbeleer:
@lewis-yeung ah, that's a good one. I didn't explicitly check that. It definitely works in primary and right as otherwise #5813 would be broken again. That's the one that broke when fixing #2885 previously.

JanDeDobbeleer:
@lewis-yeung I can't reproduce this though, my transient prompt references the `git` segment successfully (text and properties).

lewis-yeung:
@JanDeDobbeleer Have you checked on Windows? It happens more often on Windows.

JanDeDobbeleer:
@lewis-yeung I only test on Windows. Can't reproduce it at all.

JanDeDobbeleer:
@lewis-yeung care to share the config?

lewis-yeung:
A minimal example:

```yaml
version: 3
patch_pwsh_bleed: true
final_space: true
blocks:
  - type: prompt
    alignment: left
    segments:
      - type: os
        style: plain
        properties:
          windows: ' '
        template: '{{.Icon}}'
      - type: text
        style: plain
        template: foo|
      - type: text
        style: plain
        template: bar|
      - type: text
        style: plain
        template: baz|
transient_prompt:
  template: '->{{.Segments.Os.Icon}}<-'
```

Result:

screenshot

JanDeDobbeleer:
@lewis-yeung I'll test after dinner.  Using this extensively on my setup as well, haven't noticed that.

JanDeDobbeleer:
@lewis-yeung I can reproduce this, but it seems to be linked to extra only, non-extra isn't impacted.

JanDeDobbeleer:
@lewis-yeung so, the issue is with the key value map that's behind the concurrent map. Everything gets added, but not always correctly. Sometimes there seems to be issues in concurrency, even when that should not happen at all. And when it happens, the cache writes an empty map, my assumption is that the memory layout somehow breaks.

PR #5863, opened 2024-11-08
fix(template): store segment data synchronous

JanDeDobbeleer:
### Prerequisites

### DIFF
--- a/src/config/segment.go
+++ b/src/config/segment.go
@@ -138,8 +138,9 @@ func (segment *Segment) Render() {
 		return
 	}
 
-	segment.writer.SetText(text)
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer:
@lewis-yeung this should solve the issue in #5859 but the only line that really matters is the change in `segment.go`. The map rewrite is only to make it more simple. In the end it turns out the logic is correct but it writes so fast that the map breaks in parallel. Adding it again after write makes it behave correctly in all cases apparently.

Can you do a sanity check here?

lewis-yeung:
@JanDeDobbeleer Seems fixed.

PR #5865, opened 2024-11-10
feat(cache): make template cache available to all prompts

lewis-yeung:
### Prerequisites

- [x] I have read and understood the [Contributing Guide][contributing-guide].
- [x] The commit message follows the [Conventional Commits][conventional-commits].
- [x] Tests for the changes have been added (for bug fixes or features).
- [x] The documentation has been added or updated (for bug fixes or features).

### Description

The template cache is frequently used during prompt rendering, so we don't need a lazy population. BTW, this will also thoroughly resolve the issue mentioned in #2885, which is not truly fixed in the PR #5863, since the [culprit][comment] is still the one that I found about two years ago. This time, I've left comments in code to prevent negligence in future maintenance.

[contributing-guide]: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/CONTRIBUTING.md
[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/#summary
[comment]: https://github.com/JanDeDobbeleer/oh-my-posh/discussions/2885#discussioncomment-4497439


JanDeDobbeleer:
@lewis-yeung you can install `fieldalignment` and fix the error using `fieldalignment --fix "./..."` on Windows.

I missed the comment from 2 years ago, otherwise I would've caught that as well indeed. The map changes don't necessarily need to be reverted, not sure if sync.Map is more performant, especially with it's weird way to convert to a simple map. But that could be gut feeling.

src/prompt/engine.go
```diff
@@ -468,21 +468,19 @@ func New(flags *runtime.Flags) *Engine {
...
-	if flags.Type == DEBUG ||
-		flags.Type == SECONDARY ||
-		flags.Type == TRANSIENT ||
-		flags.Type == VALID ||
-		flags.Type == ERROR {
-		env.LoadTemplateCache()
-	}
+	// To prevent cross-segment template referencing issues, this should not be moved elsewhere.
+	// Related: https://github.com/JanDeDobbeleer/oh-my-posh/discussions/2885#discussioncomment-4497439
+	env.PopulateTemplateCache()
```
JanDeDobbeleer: why don't we make this part of `env.Init()` in that case?

src/runtime/terminal.go
```diff
@@ -697,9 +672,33 @@ func (term *Terminal) TemplateCache() *cache.Template {
...
+	err := json.Unmarshal([]byte(val), &tmplCache)
+	if err != nil {
+		term.Error(err)
+		return
+	}
+
+	tmplCache.Segments = tmplCache.SegmentsCache.ToConcurrent()
+}
+
+func (term *Terminal) Logs() string {
```
JanDeDobbeleer, line 549: where is this used?

src/prompt/engine.go
lewis-yeung: @JanDeDobbeleer I tried that, but it causes a problem. The order of these calls should not be changed:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/388fa2e128def345a073990e239815c33637c4ea/src/prompt/engine.go#L469-L471

The config have to be loaded after `env.Init()` so that `env.Var` can be properly assigned and then be passed to the template cache:

https://github.com/JanDeDobbeleer/oh-my-posh/blob/388fa2e128def345a073990e239815c33637c4ea/src/runtime/terminal.go#L641-L643

src/runtime/terminal.go
lewis-yeung: It's used in debugging. Actually, this is part of the existing code and I didn't touch it. Something raised by the Git diff algorithm. :)

src/prompt/engine.go
JanDeDobbeleer: I'm refactoring on top of your HEAD as it occurred to me `TemplateCache` should not be part of `runtime`, but `template`. This also solves this remark, as it can be part of `template.Init` instead. It shouldn't have any side effects but it's quite a refactor so give me tomorrow as well to get it right.

JanDeDobbeleer: Still on it, almost there. I have one issue I need to fix and then I can push this change on top.

JanDeDobbeleer: I'm so sorry, this went in a whole new direction 😓 But I do feel this is a lot cleaner.

lewis-yeung: Never mind. I'll find a moment and have a look.

lewis-yeung: Seems good.

JanDeDobbeleer: I'm finishing this today. I'm going to fix the linter issues and do a final validation.

PR #5922, opened 2024-11-23
feat(install): add msi

JanDeDobbeleer:
### Prerequisites


packages/msi/README.md
```diff
@@ -0,0 +1,40 @@
...
+```powershell
+wix build -arch arm64 -out install-arm64.msi
+```
+
+## Install the package
+
+### For the current user
+
+```powershell
+install-arm64.msi MSIINSTALLPERUSER=1 ALLUSERS=2
```
heaths: `ALLUSERS=2` is what you'd typically author into the `Properties` table. It can be passed via the command line, yes, but it might be worth informing users that "2" is context-sensitive. If they are in an elevated process, the MSI will install per-machine. Upgrades *must* happen in the same context in which the MSI was installed, so if they later try to do a per-user upgrade it'll fail with a rather unintuitive user error message; and the error message buried in a log - none by default - is only slightly more intuitive.

Typically, the default `ALLUSERS` = unset, equivalent to an empty string, will install per-user. If a user wants to do a per-machine install - likely rare given your current acquisition docs - they can pass `ALLUSERS=1`.

See <https://learn.microsoft.com/windows/win32/msi/allusers> for docs, and <https://github.com/Azure/azure-dev/blob/6cd82ae9ff02e156a0056e994215191469442ab3/cli/installer/windows/azd.wxs#L32-L34> for an example I did for the Azure Developer CLI (`azd`).

You should author those into the MSI so that the default install just works as a per-user install, which requires no elevation just like your `brew` installation, etc.

packages/msi/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+### For all users
+
+```powershell
+install-arm64.msi
```
heaths: Building on my previous comment, you'd have them pass just `ALLUSERS=1` here.

packages/msi/oh-my-posh.wxs
```diff
@@ -0,0 +1,39 @@
+<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
+	<Package Language="1033" Manufacturer="Jan De Dobbeleer" Name="Oh My Posh" UpgradeCode="b2a6bcaa-bda3-4c7f-8dec-d8665d7a9b69" Version="$(env.VERSION)" UpgradeStrategy="none">
+		<SummaryInformation Description="https://ohmyposh.dev" />
+		<MediaTemplate CabinetTemplate="tmpl{0}.cab" EmbedCab="yes" />
+		<SetProperty Id="ProgramFiles64Folder" Value="[LocalAppDataFolder]" Sequence="first" Before="CostFinalize" Condition="MSIINSTALLPERUSER=1" />
```
heaths: You shouldn't need this. Windows Installer will virtualize the directory based on the context: <https://learn.microsoft.com/windows/win32/msi/installation-context#folder-redirection>

packages/msi/oh-my-posh.wxs
```diff
@@ -0,0 +1,39 @@
...
+		</ComponentGroup>
+	</Fragment>
+	<Fragment>
+		<ComponentGroup Id="EnvironmentVariables">
+			<Component Id="SystemEnvironmentVariables" Condition="ALLUSERS=1" Directory="INSTALLDIR" Guid="ee903196-9032-4592-b461-25c96823bed7">
+				<Environment Id="SystemPath" Action="set" Name="Path" Part="first" Permanent="no" System="yes" Value="[INSTALLDIR]" />
+				<Environment Id="SystemThemesDir" Action="set" Name="POSH_THEMES_PATH" Part="all" Permanent="no" System="yes" Value="[THEMESDIR]" />
+				<Environment Id="SystemInstaller" Action="set" Name="POSH_INSTALLER" Part="all" Permanent="no" System="yes" Value="$(env.INSTALLER)" />
+			</Component>
+			<Component Id="UserEnvironmentVariables" Condition="MSIINSTALLPERUSER=1" Directory="INSTALLDIR" Guid="212dc904-b134-45b6-a647-a05ce417684a">
```
heaths: Since `MSIINSTALLPERUSER=1` should be in the `Property` table by default (you'll need to add it), condition per-machine components on `ALLUSERS=1` and per-user components as `NOT ALLUSERS`. Windows Installer treats empty strings as unset so `NOT ALLUSERS` is safest. See <https://github.com/Azure/azure-dev/blob/6cd82ae9ff02e156a0056e994215191469442ab3/cli/installer/windows/azd.wxs#L78-L90> for an example.

heaths (commented):
Pretty good overall, but I'd recommend defaulting to a per-user install, which is congruent with your current Windows install instructions as well as other platforms' install instructions e.g., using home-brew.

JanDeDobbeleer:
@heaths the current installer is used by Winget/Scoop/Store who add the correct switches by default. Is that still a relevant change in that case? If there's an easy way to set the default, all in favor though.

heaths:
> @heaths the current installer is used by Winget/Scoop/Store who add the correct switches by default. Is that still a relevant change in that case? If there's an easy way to set the default, all in favor though.

I'd still make the default per-user. One advantage of using an MSI - as long as you keep the same `UpgradeCode` and change the `ProductCode` and `ProductVersion` (any of the first 3 fields; the 4th is ignored) with each revision - no matter how a user gets it, it'll always be installed or upgraded. Don't depend on a single chained or package manager. Someone could install it via widget but upgrade it later via scoop. And different package managers require different managers or even default per-user or per-machine contexts differently. Let the MSI dictate the defaults.

heaths:
...though, more likely is someone install it via scoop or chocolatey and upgrade it via winget. The former two keep a manifest of what they installed and can upgrade, while winget will discover any MSI installed (uses the MSI APIs) and offer an upgrade based on its manifest if one is available regardless of how it's installed. That's why I've worked with partners on using MSI as the installer package rather than package manager-specific manifests: it's easy to get duplicative installs that way, but having a single registered package that dictates its default behavior works out regardless of how it's installed.

packages/msi/oh-my-posh.wxs
JanDeDobbeleer: I might need this if the actual installation path deviate from Inno, otherwise there are two installations. Validating this later today.

packages/msi/README.md
JanDeDobbeleer: I found that adding `Scope="perUserOrMachine"` to the `Package` element does the same. Wouldn't have found out without your comment though!

JanDeDobbeleer:
@heaths I think I got them all 😅

PR #5953, opened 2024-11-28
feat(block): added FillerTemplate to block

dusktreader:
### Prerequisites

### Description

This change introduces a new block configuration called "FillerTemplate". This allows the user to specify a template for filler so that different behavior can be configured when the block is handling an overflow.

An example configuration might be:

```yaml
- type: prompt
  alignment: right
  filler_template: "{{ if .Overflow }} {{ else }}<#3d59a1,transparent>━</>{{ end }}"
  overflow: break
```

This would draw filler text ("-") when there is no overflow and empty space when there is an overflow.

### DIFF
--- a/src/prompt/engine.go
+++ b/src/prompt/engine.go
@@ -20,9 +20,11 @@ type Engine struct {
 	activeSegment         *config.Segment
 	previousActiveSegment *config.Segment
 	rprompt               string
+	Overflow              config.Overflow
 	prompt                strings.Builder
 	currentLineLength     int
 	rpromptLength         int
+	Padding               int
 	Plain                 bool
 }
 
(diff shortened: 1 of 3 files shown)


dusktreader:
Example renderings:

Given config:
```yaml
- type: prompt
  alignment: right
  filler_template: "{{ if .Overflow }} {{ else }}<#3d59a1,transparent>━</>{{ end }}"
  overflow: break
```

No Overflow:
image

With Overflow:
image

JanDeDobbeleer:
@dusktreader you could also enable `filler` to be template-able, it doesn't need an additional property. I'll refactor this a bit so you can see how that would work.

JanDeDobbeleer:
@all-contributors please add @dusktreader for code,doc

This is pure magic 🪄! @holopin-bot @dusktreader wizard

dusktreader:
Thanks! Your refactor is definitely a better approach. I'm excited for the next release!

PR #6014, opened 2024-12-14
feat(cache): invalidate SCM caches when branch changes

### LINKED ISSUE #6012: Allow some segment-related command to refresh segment cache
### Code of Conduct

I'd love to have, for example, executing `git` to refresh the `Git` segment cache. This should likely apply to all source control segments and may for other segments as well.

I am a contributor on several massive repos that, even with some past git improvements to improve perf, still take a while to render the prompt. Caching would be great, but any amount practical for caching e.g., 5m, is often too long when I'm switching branches frequently - not uncommon when `rebasing --onto` topic branches against old, now-merged topic branches (squash merge we use more often).

I was thinking that since OMP already supports "hooking" commands to support tooltips, it may be able to support this as well. This way, if I run any `git` command - not necessary `git branch` or `git checkout` (there are so many ways to switch branches, so I don't think it's practical to try to support subcommands) - you can refresh the segment immediately.

heaths:
Resolves #6012

### DIFF
--- a/src/config/segment.go
+++ b/src/config/segment.go
@@ -223,9 +223,10 @@ func (segment *Segment) restoreCache() bool {
 		return false
 	}
 
-	data, OK := segment.env.Session().Get(segment.cacheKey())
+	cacheKey := segment.cacheKey()
+	data, OK := segment.env.Session().Get(cacheKey)
 	if !OK {
-		log.Debug("no cache found for segment: ", segment.Name())
+		log.Debugf("no cache found for segment: %s, key: %s", segment.Name(), cacheKey)
 		return false
 	}
 
--- a/src/segments/git.go
+++ b/src/segments/git.go
@@ -217,7 +217,9 @@ func (g *Git) CacheKey() (string, bool) {
 		return "", false
 	}
 
-	return dir.Path, true
+	ref := g.FileContents(dir.Path, "HEAD")
+	ref = strings.Replace(ref, "ref: refs/heads/", "", 1)
+	return fmt.Sprintf("%s@%s", dir.Path, ref), true
 }
 
 func (g *Git) Commit() *Commit {


heaths:
This is only a WIP, but wanted to get any initial feedback. I realized as I was about to submit this that I don't think we can take a cache "tag" (with HTTP etags as an influence) only into account but will have to consider that along with a cache duration because info like the number of modified, removed, and added files, etc., is also part of that cache and would never update. Still, it's a start, if you're interested in the feature and have any initial feedback.

heaths:
I updated the algorithm to always take duration into account. Testing with local edits, this works and makes much more sense. I set my duration to 10s and added/removed files. The prompt most often doesn't pick them up right away unless I switch branches or checkout HEAD~ as detached, etc.

If you're good with this, I can proceed with any feedback, add tests, and update other SCMs that could benefit. I'd have to research a few of them given either lack of familiarity or, like with mercurial, it's been a while since I've used and don't remember its housekeeping.

src/cache/cache.go
```diff
@@ -16,9 +16,18 @@ type Cache interface {
 	// Returns the value and a boolean indicating if the key was found.
 	// In case the duration expired, the function returns false.
 	Get(key string) (string, bool)
+	// Gets the value for a given key and tag.
+	// Returns the value and a boolean indicating if the key was found.
+	// In case the duration is expired or the tag does not match the cached tag,
+	// the function returns false.
+	GetTagged(key, tag string) (string, bool)
```
JanDeDobbeleer: I don't see why this needs to be part of this interface, a `SegmentWriter `can implement another interface for this to work and we can use the key in combination with a tag as the actual key value, right? It doesn't need to change the `Cache` interface.

heaths: I suppose that could work, though we wouldn't even need a new interface in that case: just return the current `CacheKey()` + the file content from `CacheKey()`, but wouldn't that mean your current session's cache file keeps growing with each branch change? I'll try it and see.

Alternatively, the value stored could be a wrapper such that the key doesn't change but we store and compare the tag along with the serialized `writer`.

src/cache/file.go
```diff
@@ -68,7 +68,22 @@ func (fc *File) Get(key string) (string, bool) {
 		return "", false
 	}
 
-	if co, ok := val.(*Entry); ok {
+	if co, ok := val.(*Entry); ok && !co.Expired() {
```
heaths: @JanDeDobbeleer is this change okay? It matches the description both here and in the `Context` interface i.e., that the value isn't returned if expired.

src/cache/cache.go
heaths: Yeah, just changing the key will work. I wasn't considering that the cache is rewritten each time so there's no risk of growth (apart from lots of cache files being created normally, but a cronjob could clean that up).

I'll add an interface for segments to implement who want to provide an extra tag to their key and we can keep the `Cache` interface as it is in main.

src/cache/file.go
JanDeDobbeleer: @heaths they are removed on initialization so you don't have to worry about that when calling the cache.

src/cache/cache.go
JanDeDobbeleer: You can actually reuse [this](https://github.com/JanDeDobbeleer/oh-my-posh/blob/a9a8dcd4f2c120deed189a865ce4c55de779a99e/src/config/segment.go#L274) logic. It's implemented [here](https://github.com/JanDeDobbeleer/oh-my-posh/blob/a9a8dcd4f2c120deed189a865ce4c55de779a99e/src/segments/git.go#L214) for Git for example, we can extend that using the logic you have in mind.

heaths: I have something working, but there's a lot of duplication in `git.go` e.g., setting `g.realDir`, `g.Ref`, etc., I'm trying to clean up with introducing any regressions.

heaths: Not sure it's worth it. There are several "entry points" into the `git.go` file that need different information at different times. Most access is simply checking file paths, which should be pretty quick. I need to get file content, which some of the similar code paths don't need to do. On top of that, some fields e.g., `g.Ref` may get set later to something more...accurate...once `git` command lines are run. So maybe it's safer like this?

heaths:
Tested against main and linked worktrees as well as submodules. Switching branches immediately refreshes info e.g., repo modifications. cd'ing into submodules and back retains the cache for the specified duration e.g., 30s.

src/cache/cache.go
JanDeDobbeleer: @heaths with this you mean the current code changes? I know the git segment is rather heavy. One of those parts in the codebase that grew disproportionately.

heaths: Sorry. I meant that after following several code paths from `segment.go` into `git.go` (and `scm.go`), I wasn't sure it was worth refactoring - not without a major overhaul anyway. While there were several code paths that do at least basic checks for whether we're in a main or linked worktree, not all of them need to read the `HEAD` file content. Checking file paths repeatedly isn't so expensive, I'd think, but adding file content reads unnecessarily is likely worse.

Perhaps worth a separate change to consolidate incremental updates to various members e.g., `g.rootDir`, `g.Ref`, and a few others could be set early e.g., from `shouldDisplay()`, which gets called early in most code paths into `git.go`?

JanDeDobbeleer: @heaths reading the contents isn't that crazy expensive TBH. You can see it here:

```
[TRACE] 20:46:02.503 scm.go:FileContents(D:\oh-my-posh\.git/HEAD) - 0s
```

JanDeDobbeleer:
@all-contributors please add @heaths for code

This is pure magic 🪄! @holopin-bot @heaths wizard

PR #6129, opened 2025-01-29
Add Git Branch Pattern property

Zelnes:
### Prerequisites

### Description

This proposal aims to give control over the branch name that is extracted, with regular expressions.

I am not familiar with `Go`, but tried my best.

In my workflow, I usually have long branch name (such as `feature/PROJECT-12345-scope-of-ticket`), and while interacting in the terminal, I don't really care about the scope. This feature allows me to extract only the issue part, and can still be combined with the `mapping_branches`.

Can't wait to have feedback, and hopefully see something like this in the project, which is great btw, good job !

> Also, I have to say that the setup was pretty easy, thanks to the `devcontainer`. :+1:

### Question

Is the documentation/schema aimed to be edited by hand ? Or is there anything I missed to handle this automatically ?
I felt it hard to write this part, mainly due to alignment.

### DIFF
--- a/src/segments/git.go
+++ b/src/segments/git.go
@@ -121,6 +121,8 @@ const (
 	IgnoreSubmodules properties.Property = "ignore_submodules"
 	// MappedBranches allows overriding certain branches with an icon/text
 	MappedBranches properties.Property = "mapped_branches"
+	// BranchPatterns allows for a custom regexes to be used to extract parts of the branch name
+	BranchPatterns properties.Property = "branch_patterns"
 
 	DETACHED     = "(detached)"
 	BRANCHPREFIX = "ref: refs/heads/"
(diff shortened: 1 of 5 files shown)


JanDeDobbeleer:
@Zelnes your PR made me think about what we do with branches and I just made something more interesting there to allow leveraging templates with branch name formatting. So to enable this, all we need to do is expose the regex function in the template and done.

Zelnes:
Thanks for the reply, and glad to hear that something similar can show up.
What should we do here ? Close the PR ?

JanDeDobbeleer:
@Zelnes I would close it, but I'll still attribute you in the change I'm making. Would also be cool if you could validate that.

Zelnes:
I'm okay with this proposition :smile:

PR #6216, opened 2025-03-03
fix: use `.PhysicalAvailableMemory` instead of `PhysicalFreeMemory` to show memory usage

Edu4rdSHL:
### Prerequisites

### Description

When showing memory usage, what you want to know is the really **used** memory, which doesn't include cached memory; cache memory can be claimed by the kernel under memory pressure.

The Linux kernel will try to use as much cache as possible by default, but that doesn't mean that the RAM is not available. This PR fixes that and shows only the real used memory.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2982,6 +2982,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "BoscoDomingo",
+      "name": "Bosco Domingo",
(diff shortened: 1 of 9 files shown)


JanDeDobbeleer:
@all-contributors please add @Edu4rdSHL for design,doc

This is pure magic 🪄! @holopin-bot @Edu4rdSHL wizard

PR #6238, opened 2025-03-06
Fix #6237; omp.lua script errors if launching oh-my-posh fails

### LINKED ISSUE #6237: omp.lua script errors if launching oh-my-posh fails
### Code of Conduct

### What happened?

In the [clink](https://github.com/chrisant996/clink) repo, a user reported issue [clink#727 - clink crashes when cd'ing to a long path](https://github.com/chrisant996/clink/issues/727).

When [long paths are enabled](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry) if you CD to a path exceeding the 255 characters clink hangs for a while and then crashes.  (The crash was due to a bug in the C runtime's implementation of `_popen()`, where upon failure the C runtime tries to close the same file handle twice, which is an invalid operation, causing it to report a malfunction and terminate the process.  Clink v1.7.12 includes a workaround for the C runtime bug.)

The crash has been fixed now, but it led to also discovering some related issues in the omp.lua init script from oh-my-posh.

The problems in omp.lua have to do with missing error handling, which aborts script execution and prints Lua errors to the terminal.

**REPRO:**
1. Make sure long paths are enabled ([see here](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)).
2. Run `clink config prompt use oh-my-posh` to configure using oh-my-posh for the prompt.
3. Make a directory (in cmd.exe without clink running) `C:\source\temp\qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq\asdfasdfasdf\sdfasdfasdfasdfasdf\afsfasfasdfasdfasdfasdf`
4. In cmd.exe with clink running cd to the directory.

**EXPECTED:**
- Something reasonable, and especially no crash and no Lua error messages.

**ACTUAL:**
- In Clink v1.7.11 and lower, cmd.exe crashes.  That's not oh-my-posh's fault.
- In Clink v1.7.12 and higher, the omp.lua init script from oh-my-posh encounters various errors which it doesn't handle, and the Lua engine terminates the script and prints various Lua errors to the terminal.

### Theme

N/A -- Not related to themes, and occurs with all themes, including the default theme.

Windows

cmd

### Log output

```Shell
N/A
```

chrisant996:
fix: The omp.lua script was missing some error handling, which caused Lua to abort the script and print errors to the terminal.

- Error handling is needed when calling `io.popen()`.
- Error handling is needed when calling `io.popenyield()`.
- Error handling is needed when calling `os_clock_millis()`.
- When the omp.lua prompt filter failed to generate a prompt string, then there was no prompt at all, which was very confusing.  Now upon failure it writes a log entry into the clink.log file and generates a simple basic default prompt string plus a note that it failed.

resolves #6237


JanDeDobbeleer:
@all-contributors please add @chrisant996 for code

This is pure magic 🪄! @holopin-bot @chrisant996 wizard

PR #6338, opened 2025-04-02
feat(az): reverse match

tleepa:
### Prerequisites

### Description

This PR allows AZ segment to show information from PowerShell module first and then from Azure CLI, if the former does not exist.


JanDeDobbeleer:
@tleepa there's a smarter way to go about this by only leveraging the source keyword. If you allow it to also accept a pipe split string, it could look like this by default: `"source": "cli|pwsh"` and use `"source": "pwsh|cli"` to do what you want without an additional property. That would mean that `firstmatch` is deprecated and needs to default to `pwsh|cli`

tleepa:
Frankly speaking, I am not a developer, and definitely not familiar with the go code, and I will not be able to make suggested changes.
Shall I drop this PR and raise a feature request instead?

JanDeDobbeleer:
I'll address that here so you can have an opportunity to learn. Pretty cool you did this then though 👏🏻

tleepa:
Let's be honest this change was not the rocket science :wink: 
Nevertheless, I appreciate your interest in this "feature" :smile:

JanDeDobbeleer:
@all-contributors please add @tleepa for code, docs

This is pure magic 🪄! @holopin-bot @tleepa wizard

PR #6368, opened 2025-04-17
feat: fvm flutter segment

### LINKED ISSUE #6365: Support Flutter Version Manager
### Code of Conduct

Hi,

To manage Flutter projects we usually have [FVM](https://fvm.app) to manage different version per project.

The Dart & Flutter segment, use the global version, no the version defined via FVM.

Image

NB: from my point of view it's a new feature rather than a bug, since FVM is not mandatory to work with Dart / Flutter.

atiqsamtia:
### Prerequisites

### Description
This PR adds support for FVM (Flutter Version Manager) to show current directory fvm flutter version instead of the global flutter version

This fixes #6365

### DIFF
--- a/src/segments/dart.go
+++ b/src/segments/dart.go
@@ -18,6 +18,11 @@ func (d *Dart) Enabled() bool {
 	d.folders = dartFolders
 	d.commands = []*cmd{
 		{
+			executable: "fvm",
+			args:       []string{"dart", "--version"},
+			regex:      `Dart SDK version: (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
+		},
+		{
 			executable: "dart",
 			args:       []string{"--version"},
 			regex:      `Dart SDK version: (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
--- a/src/segments/dart_test.go
+++ b/src/segments/dart_test.go
@@ -22,9 +22,13 @@ func TestDart(t *testing.T) {
 			versionOutput: tc.Version,
 			extension:     "*.dart",
 		}
+
 		env, props := getMockedLanguageEnv(params)
+		env.On("HasCommand", "fvm").Return(false)
+
 		d := &Dart{}
 		d.Init(props, env)
+
 		assert.True(t, d.Enabled(), fmt.Sprintf("Failed in case: %s", tc.Case))
 		assert.Equal(t, tc.ExpectedString, renderTemplate(env, d.Template(), d), fmt.Sprintf("Failed in case: %s", tc.Case))
 	}
--- a/src/segments/flutter.go
+++ b/src/segments/flutter.go
@@ -13,6 +13,11 @@ func (f *Flutter) Enabled() bool {
 	f.folders = dartFolders
 	f.commands = []*cmd{
 		{
+			executable: "fvm",
+			args:       []string{"flutter", "--version"},
+			regex:      `Flutter (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
+		},
+		{
 			executable: "flutter",
 			args:       []string{"--version"},
 			regex:      `Flutter (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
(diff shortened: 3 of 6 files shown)


bchatard:
Hi @atiqsamtia 

Thanks for taking the topic!

If I understand your change, you add a new segment for FVM
So in configuration we will have flutter, dart and fvm segments, and saw global flutter version and fvm flutter version.

It's seems more convenient to have only flutter and dart segment, but with a more "smart" algorithm for finding the appropriate version, no?

Something like that: 
```
if file .fvmrc and fvm installed
    fvm flutter --version / fvm dart --version
else 
    flutter --version / dart --version
fi
```

There is something that seems similar for node segment: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segments/node.go

JanDeDobbeleer:
@bchatard agreed, this can be added as the first command to check for flutter and dart as well:

```golang
{
    executable: "fvm",
    args:       []string{"flutter", "--version"},
    regex:      `Flutter (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
},
```

atiqsamtia:
fvm can have different version depending on the current directory. That's why I created a new segment. 

I'm going to see if I can do this fvm exists check first. Go is not my native language either..

I'll test a new approach and send a PR if it works well.

Regarding node segment, as per my understanding, when you switch node version, it switches globally not per directory which is a bit easy to work with. But for fvm we want full compatibility of showing the global version where fvm is not specified, and show the local version where it is specified.

atiqsamtia:
Able to detect FVM and flutter both with this

```go
f.commands = []*cmd{
		{
			executable: "fvm",
			args:       []string{"flutter", "--version"},
			regex:      `Flutter (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
		},
		{
			executable: "flutter",
			args:       []string{"--version"},
			regex:      `Flutter (?P<version>((?P<major>[0-9]+).(?P<minor>[0-9]+).(?P<patch>[0-9]+)))`,
		},
	}
```

But unable to modify the tests to handle multiple commands without failing.

```go
func TestFlutter(t *testing.T) {
	cases := []struct {
		Case           string
		ExpectedString string
		Version        string
		isFVM          bool
	}{
		{Case: "Flutter 2.10.4", ExpectedString: "2.10.4", Version: "Flutter 2.10.4 • channel stable • https://github.com/flutter/flutter.git", isFVM: false},
		{Case: "Flutter 2.10.4", ExpectedString: "2.10.4", Version: "Flutter 2.10.4 • channel stable • https://github.com/flutter/flutter.git", isFVM: true},
	}
	for _, tc := range cases {
		params := &mockedLanguageParams{
			cmd:           "flutter",
			versionParam:  "--version",
			versionOutput: tc.Version,
			extension:     "*.dart",
		}
		if tc.isFVM {
			params.cmd = "fvm"
			params.versionParamList = []string{"flutter", "--version"}
		}
		env, props := getMockedLanguageEnv(params)
		if !tc.isFVM {
			env.Unset("HasCommand")
			env.On("HasCommand", "fvm").Return(false)
		}
		d := &Flutter{}
		d.Init(props, env)
		assert.True(t, d.Enabled(), fmt.Sprintf("Failed in case: %s", tc.Case))
		assert.Equal(t, tc.ExpectedString, renderTemplate(env, d.Template(), d), fmt.Sprintf("Failed in case: %s", tc.Case))
	}
}
```

JanDeDobbeleer:
@atiqsamtia I can solve that. Leave the PR open for me to have a look at that.

atiqsamtia:
I'll discard this PR and create a new one with new approach as it requires less number of files changes and cleaner PR.

Thanks

JanDeDobbeleer:
@atiqsamtia @bchatard if someone can confirm the output of `fvm dart --version` and `fvm flutter --version`, that would be ace.

bchatard:
hi @JanDeDobbeleer 

sorry for the late reply, I was on vacation 


here the outputs:

```
$ fvm dart --version
Dart SDK version: 3.2.3 (stable) (Tue Dec 5 17:58:33 2023 +0000) on "macos_arm64"

$ fvm flutter --version
Flutter 3.16.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision b0366e0a3f (1 year, 5 months ago) • 2023-12-05 19:46:39 -0800
Engine • revision 54a7145303
Tools • Dart 3.2.3 • DevTools 2.28.4
```

JanDeDobbeleer:
@atiqsamtia all fine and already live 😃

bchatard:
Am I missing something?

image

PR #6427, opened 2025-05-08
feat: instant prompt

### LINKED ISSUE #5310: p10k-like instant prompt
### Code of Conduct

Some segments can take a long time to load, such as the Last.fm music segment, as it has to complete an http request, or segments that run custom shell commands/scripts. This can slow down the load time considerably. `powerlevel10k` has a feature called "[instant shell](https://github.com/romkatv/powerlevel10k?tab=readme-ov-file#instant-prompt)" that allows you to start typing your command while different zsh plugins load.

A similar feature that allows you to type immediately, then loads each segment independently would go a long way to make the shell feel a lot faster. With the Last.fm example, every segment would practically load instantly, allowing you to start typing, only for the Last.fm segment to appear whenever it is ready.

#2861 suggests a similar feature for PowerShell, but the feasibility of the implementation was never discussed.

JanDeDobbeleer:
### Prerequisites

### Description

resolves #5310

This changes the way Oh my Posh initializes. Instead of evaluating a script,
it now saves that script in a file in the cache (or the shell's autoload folder)
and only adjusts this when changes happen in Oh My Posh' version or the configuration.
This improves the staartup time quite a bit and allows async loading for an instant prompt
feature.

BREAKING CHANGE: this requires Nushell users to
remove `source ~/.oh-my-posh.nu` from their `$nu.config-path` file.

BREAKING CHANGE: removes support for tcsh due do limited documentation (and my patience ran out).

[builds](https://github.com/JanDeDobbeleer/oh-my-posh/actions/runs/14997493670/artifacts/3114544721)


src/shell/io/io.go
```diff
@@ -0,0 +1,144 @@
...
+	if len(scriptPathCache) != 0 {
+		return scriptPathCache, nil
+	}
+
+	if env.Flags().Shell != "nu" {
+		scriptPathCache = filepath.Join(cache.Path(), scriptName(env))
+		return scriptPathCache, nil
+	}
+
+	path, err := env.RunCommand("nu", "-c", "$nu.user-autoload-dirs | get 0")
```
fdncred: You could also do `nu -c '$nu.user-autoload-dirs.0'`. I'm guessing it's a tiny bit faster, but I haven't seen any performance metrics around it. Either way should work fine.

NotTheDr01ds: I think we'd prefer that this be `$"($nu.data-dir)/vendor/autoload"`.  User autoloads should be for *user* configuration, but third-party code should use vendor autoloads. This vendor directory is guaranteed to be auto-loaded.

Issue #4378, opened 2023-10-25
WSL setup with spotify api cannot show utf-8


phong-nt-990:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What happened?

Using ohmyposh with spotify api to display song titles, in WSL, cannot display utf-8 characters.

### Theme

...        
{
          "foreground": "#ff20b5",
          "properties": {
            "paused_icon": "\uf04c ",
            "playing_icon": "\ue602 ",
            "stopped_icon": "\uf04d ",
            "track_separator": " - "
          },
          "style": "plain",
          "template": " \uf1bc {{ .Icon}}{{ if ne .Status \"stopped\" }}{{ .Track }}{{ end }} ",
          "type": "spotify"
        },
...

### What OS are you seeing the problem on?

Windows

### Which shell are you using?

bash

### Log output

```Shell
Version: 18.15.0

Shell: bash (4.4.20(1)-release)

Prompt:

┌  (WSL)⠀  bash⠀  M4PL3-DESKTOPm4pl3⠀                        Vi?t Nam Nh?ng Chuy?n Di |  21.1.0 |  base |  100  | 00:22:58 
└[ mnt »  »  » maple ]
➜

Segments:

ConsoleTitle(true)                        -   2 ms
Text(true)                                -   0 ms
Os(true)                                  -   0 ms
Shell(true)                               -   1 ms
Session(true)                             -   0 ms
Git(false)                                - 102 ms
Spotify(true)                             - 248 ms
Dart(false)                               -  14 ms
Node(true)                                -  13 ms
Python(true)                              -   8 ms
Battery(true)                             -   1 ms
Time(true)                                -   2 ms
Path(true)                                -   0 ms
Exit(true)                                -   0 ms

Run duration: 373.031768ms

Cache path: /home/m4pl3/.cache/oh-my-posh

Config path: /mnt/c/Users/maple/.config/powershell/hul10.omp.json

Logs:

[DEBUG] 00:22:58.768 shell.go:Getenv:392 → NO DATA
[TRACE] 00:22:58.768 shell.go:Getenv(TERM_PROGRAM) - 2.273097ms
[DEBUG] 00:22:58.768 shell.go:Getenv:392 → NO DATA
[TRACE] 00:22:58.768 shell.go:Getenv(OMP_CACHE_DIR) - 3.773µs
[DEBUG] 00:22:58.768 shell.go:Getenv:392 → NO DATA
[...]

PR #6431, opened 2025-05-11
fix [segment/spotify]: Use PowerShell with UTF-8 output for WSL

kaien07:
### Prerequisites

### Description

Previously, the WSL Spotify segment used tasklist.exe, which resulted in incorrect encoding, causing non-ASCII characters such as Chinese and Japanese words to appear as question marks.

This modifies the WSL implementation of the Spotify segment to execute a Powershell command to retrieve the Spotify.exe window title, forcing Powershell's console output to UTF-8 to ensure correct character transfer, before parsing the UTF-8 string to extract the artist and the track.

This resolves issues with displaying CJK and other Unicode characters in the Spotify segment on WSL when the Spotify desktop app is used on the Windows host.

Related to #4378.


src/segments/spotify_linux.go
```diff
@@ -47,40 +48,47 @@ func (s *Spotify) runLinuxScriptCommand(command string) string {
 
 // Implementation for WSL - moved from spotify_wsl.go
 func (s *Spotify) enabledWsl() bool {
-	tlist, err := s.env.RunCommand("tasklist.exe", "/V", "/FI", "Imagename eq Spotify.exe", "/FO", "CSV", "/NH")
-	if err != nil || strings.HasPrefix(tlist, "INFO") {
+	psCommand := `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8; (Get-Process Spotify -ErrorAction SilentlyContinue | Where-Object {$_.MainWindowTitle -ne ""} | Select-Object -First 1).MainWindowTitle`
+
+	cmd := exec.Command("powershell.exe", "-NoProfile", "-NonInteractive", "-Command", psCommand)
```
JanDeDobbeleer: you need to use `s.env.RunCommand` here and not the native exec. That way everything is logged for troubleshooting and you don't need to manage the buffer.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@all-contributors please add @kaien07 for code

This is pure magic 🪄! @holopin-bot @kaien07 wizard

PR #6506, opened 2025-06-03
fix(init): set session ID correctly

lewis-yeung:
### Prerequisites

### Description

Fix a bug that caches get messed up across shell sessions.


JanDeDobbeleer:
@lewis-yeung ah, because the script contains the same session ID rather than the session. I really didn't think that though... BTW, I'd rather not have newlines in the outputted source command. It used to give issues with powershell.

JanDeDobbeleer:
@lewis-yeung for `nushell` we still have an issue. I'll try to think about how to resolve that.

JanDeDobbeleer:
@lewis-yeung TBH, I don't think we need the session cache on `init`. This means we can generate the UUID from the script rather than inject it. That would solve it everywhere.

PR #6621, opened 2025-07-21
feat(dsc): add support

### LINKED ISSUE #4340: DSC 3 Resource for oh-my-posh
### Code of Conduct

### Suggested new feature or improvement

I'd like to be able to use the [WinGet Configuration](https://learn.microsoft.com/windows/package-manager/configuration/) feature to configure oh-my-posh.

In the future when DSC v3 supports the "export" behavior, this could also be used as a part of the export a configuration scenario.
* https://github.com/microsoft/winget-cli/issues/3274
* https://github.com/PowerShell/DSC/issues/73

### Scenario

I want to set up a new device and have the ability to configure oh-my-posh with WinGet Configure or by selecting the configuration in [Dev Home](https://learn.microsoft.com/windows/dev-home/).

### Additional details

I would expect the Microsoft.WinGet.DSC resource would be used to install oh-my-posh on my machine. Then as a dependency I'd like to be able to use a DSC resource to configure oh-my-posh the way I like it.

oh-my-posh

JanDeDobbeleer:
### Prerequisites

### Description

resolves #4340 

### TODO

- [x] rework the schema so each shell can have a different configuration if needed
- [x] copy the config and dsc schema to the website prior to deploying (action)
- [x] store font install metadata in dsc state when run
- [x] store configuration file data in dsc state on init (use `~` as a pointer to `$HOME`)
- [x] import fonts
- [x] import configurations
- [x] import shell setting
- [x] adjust shell init lines to be shell specific (`eval`, `Invoke-Expression`, ...)
- [x] add import switches to install fonts and initialize the shells (disabled by default)


packages/msi/oh-my-posh.dsc.resource.json
```diff
@@ -0,0 +1,59 @@
+{
+  "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/bundled/resource/manifest.json",
```
michaeltlombardi: With the official release of DSC, this is no longer a valid/recognized schema URI. Strongly recommend using this URI for forward compatibility:

```suggestion
  "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
```

If you're concerned about editing your schema in VS Code, I recommend you use the enhanced authoring schema (VS Code doesn't recognize canonically bundled schemas yet, as support for 2020-12 is only partially implemented):

```suggestion
  "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.vscode.json",
```

For more information about schema URIs and DSC, see [DSC JSON Schema URIs][01].

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/schema-uris

packages/msi/oh-my-posh.dsc.resource.json
```diff
@@ -0,0 +1,59 @@
...
+  "description": "Allows management of settings state via the DSC v3 command line interface protocol.",
```
michaeltlombardi: This description is surfaced from DSC itself, as with the [`dsc resource list`][01] command. I think the description should speak more to managing oh-my-posh's settings than to this enabling the use of DSC.

Maybe enumerate categories of settings or particularly common ones here?

packages/msi/oh-my-posh.dsc.resource.json
```diff
@@ -0,0 +1,59 @@
...
+      "test",
+      {
+        "jsonInputArg": "--schema",
+        "mandatory": true
+      }
+    ]
+  },
+  "tags": [
+    "OhMyPosh"
+  ],
```
michaeltlombardi: I would consider adding the platform tags here, like `Linux`, `macOS`, and `Windows` to indicate support. You may also want to add a tag like `prompt` or `theming` to help with searches.

packages/msi/oh-my-posh.dsc.resource.json
```diff
@@ -0,0 +1,59 @@
...
+  "type": "OhMyPosh/Settings",
+  "version": "1.10.0"
```
michaeltlombardi: Just to be clear, the version field in the manifest should represent the version of the _resource_, not (necessarily) the version of the software the resource manages.

You may want to do your initial release as `0.1.0` and when you're ready to consider it fully supported align to the software version.

src/cli/dsc/schema.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
michaeltlombardi: Strongly recommend using 2020-12 for your schema, which supports much cleaner expression of data types, composition, and validation.

src/cli/dsc/schema.json
```diff
@@ -0,0 +1,77 @@
...
+      "type": "array",
+      "title": "Configuration List",
+      "description": "List of Oh My Posh configurations",
+      "items": {
+        "type": "object",
+        "properties": {
+          "path": {
+            "type": "string",
+            "title": "Configuration Path",
+            "description": "File path to the Oh My Posh configuration file"
```
michaeltlombardi: May want to clarify whether this is an absolute file path. Presumably, this is a required property for the entry and should be in the array for the `required` keyword.

src/cli/dsc/schema.json
```diff
@@ -0,0 +1,77 @@
...
+          },
+          "format": {
+            "type": "string",
+            "title": "Configuration Format",
+            "description": "Format of the configuration file",
+            "enum": [
+              "json",
+              "yaml",
+              "toml"
+            ]
```
michaeltlombardi: Can this format mismatch with the `path`? How does it behave if it does?

src/cli/dsc/schema.json
```diff
@@ -0,0 +1,77 @@
...
+          },
+          "data": {
+            "title": "Configuration Data",
+            "description": "The actual Oh My Posh configuration object",
+            "$ref": "https://ohmyposh.dev/schema.json"
```
michaeltlombardi: This isn't necessarily an issue, but it's best practice as part of your build step to ensure you have a canonically bundled schema for your resource, so you'll want the data schema added to the `$defs` keyword.

src/cli/dsc/schema.json
```diff
@@ -0,0 +1,77 @@
...
+          }
+        },
+        "additionalProperties": false
+      }
+    },
+    "shells": {
+      "type": "array",
+      "title": "Shell List",
+      "description": "List of installed shells using Oh My Posh",
```
michaeltlombardi: Is this intended to be configurable from the resource, or just informational? If the latter, you'll want to define the `readOnly` keyword as `true` to indicate the non-configurable nature of this property.

michaeltlombardi (commented):
Just a quick review from a design perspective for the resource - haven't done a deep-dive into the design goals or implementation yet.

src/cli/dsc/schema.json
JanDeDobbeleer: It could, and then we end up in a broken state. Which can only happen when someone manually crafts a state though. I was thinking about just using the path's extension and not make this explicit.

src/cli/dsc/schema.json
JanDeDobbeleer: This can be configured. That logic is implemented.

packages/msi/oh-my-posh.dsc.resource.json
SteveL-MSFT: Would suggest adding `powershell` as a tag :)

packages/msi/oh-my-posh.dsc.resource.json
Frulfump: Is there a VS Code issue to track for that support?

Gijsreyn: Here you go: https://github.com/microsoft/vscode/issues/165219. It will also be documented in a known issue [document](https://github.com/PowerShell/DSC/pull/926).

src/cli/dsc/schema.json
michaeltlombardi: You could definitely just use the path extension and add a validation pattern to ensure it ends in a supported format extension.

PR #6652, opened 2025-08-03
Feat: add golang support to project segment

jsnfwlr:
### Prerequisites

- [ x] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [ x] The commit message follows the [conventional commits][cc] guidelines.
- [ x] Tests for the changes have been added (for bug fixes / features).
- [ x] Docs have been added/updated (for bug fixes / features).

### Description

Adding in support to display the Go version of a golang project as the `target` if the CWD is below a folder that contains either a go.work or go.mod file.

### DIFF
--- a/src/runtime/terminal.go
+++ b/src/runtime/terminal.go
@@ -444,6 +444,14 @@ func (term *Terminal) HTTPRequest(targetURL string, body io.Reader, timeout int,
 	return responseBody, nil
 }
 
+type ErrorTeminal string
+
+const ErrNoMatchAtRootLevel = ErrorTeminal("no match at root level")
+
+func (e ErrorTeminal) Error() string {
+	return string(e)
+}
+
 func (term *Terminal) HasParentFilePath(parent string, followSymlinks bool) (*FileInfo, error) {
 	defer log.Trace(time.Now(), parent)
 
@@ -475,7 +483,7 @@ func (term *Terminal) HasParentFilePath(parent string, followSymlinks bool) (*Fi
 		}
 
 		log.Error(err)
-		return nil, errors.New("no match at root level")
+		return nil, ErrNoMatchAtRootLevel
 	}
 }
 
--- a/src/segments/project.go
+++ b/src/segments/project.go
@@ -3,6 +3,7 @@ package segments
 import (
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"slices"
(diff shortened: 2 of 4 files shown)


src/runtime/environment.go
```diff
@@ -38,6 +38,7 @@ type Environment interface {
 	HasFolder(folder string) bool
 	HasParentFilePath(input string, followSymlinks bool) (fileInfo *FileInfo, err error)
 	HasFileInParentDirs(pattern string, depth uint) bool
+	FileInParentDirs(file string, depth uint) string
```
JanDeDobbeleer: What's the difference between this and `HasParentFilePath`?

jsnfwlr: My function returns the path to the file

JanDeDobbeleer: @jsnfwlr so does the other, no? In the file object.

jsnfwlr: Apologies, I just noticed you asked about `HasParentFilePath` not `HasFileInParentDirs` - I didn't even see `HasParentFilePath` when I was making this change.

I will correct the mistake and use that instead.

jsnfwlr: This has been remedied now.

src/runtime/terminal.go
```diff
@@ -444,6 +444,14 @@ func (term *Terminal) HTTPRequest(targetURL string, body io.Reader, timeout int,
 	return responseBody, nil
 }
 
+type ErrorTeminal string
```
JanDeDobbeleer, line 447: There's a typo here

src/segments/project.go
```diff
@@ -313,6 +323,65 @@ func (n *Project) getPowerShellModuleData(_ ProjectItem) *ProjectData {
...
+	var err error
+
+	if item.Name != "golang" {
+		return nil
+	}
+
+	for _, file := range item.Files {
+		match, err = n.env.HasParentFilePath(file, true)
+		if err != nil {
+			if errors.Is(err, runtime.ErrNoMatchAtRootLevel) {
```
JanDeDobbeleer, line 338: I don't think this error is really relevant as any error from this function call implies there's no match.

JanDeDobbeleer (changes requested):

src/segments/project.go
jsnfwlr: I wanted a way to determine if the error was just "couldn't find it so we don't need to raise an error to the user" vs "we encountered a permission error that should be raised to the user". If that isn't necessary I can remove it altogether

JanDeDobbeleer: Yup, because there's no such distinction ATM. I'll do it so we can merge ;-)

JanDeDobbeleer:
@jsnfwlr when refactoring a bit of this, I noticed we already have this functionality in the `golang` segment. Therefore I'm not going to merge this. Additionally, this segment is getting rather heavy, I'm going to mark this one as deprecated and move the project functionality over to the language base segment as that's perfectly possible and less heavy.

jsnfwlr:
all good - no worries.

Is there a way to combine the `golang` segment with the `project` and `docker` segments so they appear as one segment? Like a group segment that takes other segments as children?

Issue #6668, opened 2025-08-06
Support for UV Python package manager


ValerioCeccarelli:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Summary

[UV](https://docs.astral.sh/uv/) is a relatively new Python package and project manager that aims to simplify Python project development and management. Currently, the Python segment of `oh-my-posh` does not account for environments managed by UV, which can lead to discrepancies between the Python version shown by the segment and the one actually used by UV.

### Background: How UV Works

When running `uv run script.py`, UV executes the script using the appropriate Python interpreter based on the environment configuration. Its selection logic is as follows:

* **No Python-related files or folders in the current directory**: Uses UV's default Python version.
* **Only Python files in the directory**: Also uses the default version.
* **Presence of a virtual environment**: Uses the Python version from the virtual environment, **even if it is not activated**.

(this is a very small summary, but should be enough to understand the problem)

### Current Limitation in `oh-my-posh`

1. **Incorrect Python Version Displayed**:
   The Python segment currently does not consider UV. It selects the Python version by searching the PATH, which can lead to incorrect results. For example, I have Python 3.8, 3.9, and 3.10 installed. The segment shows 3.9 (as it appears first in the PATH), but `uv run script.py` uses 3.10 (my UV default). This misrepresents the version that will actually be used when running scripts via UV.

2. **Virtual Environment Detection**:
   The current logic relies on environment variables and activated virtual environments to detect which Python version to show. However, UV can detect and use virtual environments without activation. For instance, if there is a virtual environment using Python 3.8 in a directory, and UV's default is 3.10, the segment still displays 3.9, leading to further inaccuracy since `uv run script.py` would actually use Python 3.8 (even if the virtual environment is not activated).

### Proposal

Given UV’s growing popularity, I propose extending the Python segment with optional support for UV. Specifically:

* Introduce a new segment property, e.g., `use_uv`, which defaults to `false`.
* When `use_uv` is set to `true`, the Python segment should invoke UV to determine the effective Python version, replacing the current version-detection logic.

This change would allow interested users to benefit from accurate environment detection while maintaining backward compatibility for the majority who do not use UV.

### Next Steps

I will submit a pull request shortly, including both the code changes and the relevant documentation updates. I welcome feedback on implementation choices and am happy to iterate on the solution based on your input.

ValerioCeccarelli:
Hey @JanDeDobbeleer what is the status of this?
[...]

PR #6669, opened 2025-08-06
feat(python): add support for UV package manager

ValerioCeccarelli:
Related issues: #6668

### Prerequisites

Following issue #6668, this pull request adds support for [UV](https://docs.astral.sh/uv/) in the Python segment.

### Summary of Changes

1. **Added `use_uv` property**
   A new configuration property named `use_uv` has been introduced in the Python segment. When set to `true` (default is `false`), this enables the use of UV to determine the Python version.

2. **Updated version-detection logic** (conditional on `use_uv`)
   When `use_uv` is enabled, the segment now uses UV’s `run` command to resolve the Python version. This ensures alignment with UV's interpreter selection logic, which accounts for virtual environments and default versions even when environments are not activated.

3. **Documentation updates**
   The documentation has been updated to describe the `use_uv` property and its behavior.

4. **Clarification of `.Venv` property**
   I also updated the description of the `.Venv` property in the template to clarify that it reflects the active environment, not merely the presence of a virtual environment in the folder. This aligns more accurately with both the current behavior and the UV-based logic, where a present but inactive virtual environment will still be used by UV, though `.Venv` remains unset.

### Testing

I have not yet added automated tests, as I would first like to confirm whether this feature aligns with the project's direction and whether the current implementation is acceptable to you.

In the meantime, I have manually tested the changes in the following scenarios:

* Directory with no Python files or virtual environment
* Directory with Python files only
* Directory with a virtual environment (active and inactive)

In all cases, the Python segment correctly reflects the version and virtual environment name based on UV’s logic.

If you think it makes sense to add some automated tests for this feature, I’d be happy to include them; just let me know what kind of goals you have in mind for them.


pem120:
As a UV user, this feature would be awesome

JanDeDobbeleer:
@ValerioCeccarelli I want to use this as a way to introduce `features` into the config. The current properties mix overrides and features so I want to use my time to refactor that towards config V4 where properties is deprecated for `overrides` and `features`. I'll merge this with these changes once I add the docs, so it can go live prior to the huge refactor. But it will only happen after I come back from holidays in 2 weeks.

PR #6671, opened 2025-08-07
feat(jujutsu): add ChangeIDFull template property

raylu:
### Prerequisites

### Description

add support for the full (32-character canonical bit-reversed representation) jujutsu change ID

### DIFF
--- a/src/runtime/mock/environment.go
+++ b/src/runtime/mock/environment.go
@@ -3,7 +3,6 @@ package mock
 import (
 	"io"
 	"io/fs"
-	"time"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/runtime"
 	"github.com/jandedobbeleer/oh-my-posh/src/runtime/battery"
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer (changes requested):
I would replace the current `ChangeID` with the full ID and if people want a shorter ID, they can easily leverage template functions. You can adjust the current template to already do that not to break existing expectations.

raylu:
the current `ChangeID` is from `shortest()` https://jj-vcs.github.io/jj/latest/templates/#changeid-type which is the shortest _unique_ prefix. if you're on `aaab` but your repo has `aaac`, `ChangeID` will be `aaab`. if you abandon `aaac`, it could change to just `a`. it's common for the shortest form to be only 1-2 characters long

in the default `jj log` view, it's also the bit highlighted in a different color
image
so most people are used to typing just that bit

though I do admit `ChangeIDFull` is probably not useful for most people. maybe I could instead add a configuration property that lets people pass the min length argument to `shortest()` so they can get 8 chars and it'll show more on the rare off chance that's not enough to uniquely identify a change

JanDeDobbeleer:
@raylu but then wouldn't it make sense to use the full for `ChangID` and create `ChnageIDShort` for the short one? Because it's a bit in my head otherwise.

raylu:
I didn't want to break existing setups

but upon further reflection, adding a config property instead of a new template property both keeps existing setups intact while making it easier to do what most people will probably want (set `change_id_min_len` to 8 instead of having to do substr)

JanDeDobbeleer (approved):

PR #6679, opened 2025-08-12
feat(jujutsu): Add more information about the current change

nemith:
### Prerequisites

### Description

The templating system in Jujutsu is very advanced.  Almost all the command output go though the template system and can provide vast amounts of details.  Today the Jujutsu segment hardcodes the change output to two items.  `change_id.shortest()` and `diff.summary()`.

This adds a a bunch of new properties along with some logic on how to pass and parse the information to the `jj log` command.

See the updated documentation the properties added. 

This is an alternative to #6671. This change originated from allowing users to specify their own changes but changed to a set of a common change properties that are clearer to use.

### DIFF
--- a/src/segments/jujutsu.go
+++ b/src/segments/jujutsu.go
@@ -1,6 +1,8 @@
 package segments
 
 import (
+	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/log"
@@ -12,10 +14,19 @@ import (
 const (
 	JUJUTSUCOMMAND = "jj"
 
-	jjLogTemplate = `change_id.shortest() ++ "\n" ++ diff.summary()`
-
 	IgnoreWorkingCopy properties.Property = "ignore_working_copy"
+	MinChangeIDLen    properties.Property = "min_change_id_len"
+	MinCommitIDLen    properties.Property = "min_commit_id_len"
 )
+
+// JujutsuID is used for both change_id and commit_id
+type JujutsuID struct {
+	Shortest string
+	Full     string
+	Rest     string
+}
+
+func (id *JujutsuID) String() string { return id.Shortest + id.Rest }
 
 type JujutsuStatus struct {
 	ScmStatus
@@ -35,13 +46,30 @@ func (s *JujutsuStatus) add(code byte) {
 }
 
 type Jujutsu struct {
-	Working  *JujutsuStatus
-	ChangeID string
+	Working         *JujutsuStatus
+	ChangeID        string
+	Working         JujutsuStatus
+	ChangeID        JujutsuID
+	CommitID        JujutsuID
+	LocalBookmarks  []string
+	RemoteBookmarks []string
+	Description     string
+
+	Conflict  bool
+	Immutable bool
+	Empty     bool
+	Divergent bool
+	Hidden    bool
+	Mine      bool
+
+	AuthorID    User // defined in git.go
+	CommitterID User // defined in git.go
+
 	Scm
 }
 
 func (jj *Jujutsu) Template() string {
-	return " \uf1fa{{.ChangeID}}{{if .Working.Changed}} \uf044 {{ .Working.String }}{{ end }} "
+	return " \uf1fa{{ .ChangeID }}{{if .Working.Changed}} \uf044 {{ .Working.String }}{{ end }} "
 }
 
 func (jj *Jujutsu) Enabled() bool {
@@ -51,8 +79,8 @@ func (jj *Jujutsu) Enabled() bool {
 		return false
 	}
 
-	statusFormats := jj.props.GetKeyValueMap(StatusFormats, map[string]string{})
-	jj.Working = &JujutsuStatus{ScmStatus: ScmStatus{Formats: statusFormats}}
+	statusFormats := jj.props.GetKeyValueMap(StatusFormats, nil)
+	jj.Working = JujutsuStatus{ScmStatus: ScmStatus{Formats: statusFormats}}
 
 	if displayStatus {
 		jj.setJujutsuStatus()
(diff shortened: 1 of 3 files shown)


nemith:
Alternatively here, if we don't want users doing thier own JJ templates, we could combine #6671 and ChangeID would return a new object that would have fields of `Prefix`,`Rest`,`Full`,`Short`,`Empty`,`Divergent`,etc

Then also add a lot of missing data like User, CommitID (i.e git hash), local, remote branch name, etc.  

I could try to come up with another list.   But the null seperated list is working pretty well to extend the functionality.

nemith:
Here is a short list of what it could look like if templates were managed by OMP instead. 

 It can make things more discoverable  and less error prone and could mean some better default templates that utilize this.  There is also the advantage of being able to parse lists like `local_branch()` and present them to the template as a Go list instead of a string.
 
 I am sorta in the middle but maybe preffering the hardcoded route below a bit more. 
 
 @raylu, @JanDeDobbeleer  thoughts?   This is probably more inline with what you had initially. 


```go
// Additional props
jj.props.GetInt("commit_id_min_len")
jj.props.GetInt("change_id_min_len")

type Jujutsu struct {
	CommitID       ID // commit_id()
	ChangeID       ID // change_id()
	LocalBranches  []string // local_branches.join(",")
	RemoteBranches []string // local_branches.join(",")
	Description    string   // decription()

	Conflict  bool // conflict()
	Immutable bool // immutable()
	Empty     bool // empty()
	Divergent bool // divergent()
	Hidden    bool // hidden()
	Mine      bool // mine()

	AuthorID   User // author() (User struct same as Git)
	CommiterID User // comitter() (User struct same as Git)
}

// N comes from the two props above.  Defaults to 8.
type ID struct {
	Short string // <id>.shortest(N).prefix() 
	Full  string // <id>. Full change ID
	Rest  string // <id>.shorted(N).rest() The remaining id
}

func (i *ID) String() string { return i.Short } // Backwards compatable with existing configs
```

raylu:
makes sense to me! definitely way more flexible than what I had 

why do you need multiple log templates, though?

nemith:
> why do you need multiple log templates, though?

Otherwise it come as just one big string.  so if you wanted to do any formatting/colors/etc it would a pain.  There are also use cases outside of changeid that you may want.  

Also why i think doing https://github.com/JanDeDobbeleer/oh-my-posh/pull/6679#issuecomment-3180113087 makes more sense the more i think about it.

JanDeDobbeleer:
@nemith I just wouldn't make those properties ints as you're passing them straight to the CLI as a string anyways. Strings are OK as you don't want needless boxing and unboxing.

nemith:
Updated with the changes i proposed in the comment above.  I think this is way more clear for OMP users and should allow change/commit id coloring, and other things like tracking  empty or conflict commits.

syrm:
nice one !

syrm:
Is it possible to add bookmarks and distance to current commit like [starship-jj](https://crates.io/crates/starship-jj) ?

nemith:
> Is it possible to add bookmarks and distance to current commit like [starship-jj](https://crates.io/crates/starship-jj) ?

Do you have an example of this?  starship-jj isn't using the jj command directly so it depends.  Also sounds like maybe this would be good as different PR.

raylu:
sorry, I just came back to this and realized your question was "should we support custom jj templates or should we instead hardcode the templates and provide all these template variables?"

custom jj templates are more flexible and might avoid some performance issues. I don't think they're particularly hard to write. so I guess I'd slightly rather have that version (e9fd09dd5a81794bc578387f9fce98f971d8117b)

but what you have now is fine too 🤷‍♂️

syrm:
> > Is it possible to add bookmarks and distance to current commit like [starship-jj](https://crates.io/crates/starship-jj) ?
> 
> Do you have an example of this? starship-jj isn't using the jj command directly so it depends. Also sounds like maybe this would be good as different PR.

For example : 
➜  jj
@  qsxlzrlq srm@sample.com 2025-09-07 20:37:28 14f4dbc3
│  (empty) (no description set)
○  zmnvvqnv srm@sample.com 2025-09-07 20:37:26 git_head() d7dec052
│  (empty) (no description set)
○  psnqyolu srm@sample.com 2025-09-07 20:37:18 2bdb03ea
│  (no description set)
◆  kuznnowx srm@sample.com 2025-09-07 17:20:58 **main** c0e7ef80
│  chore: refactor process, view now query data (#10)

The prompt is : 
myproject : main🠕3  [0m +0-0]

Other example : 
➜  jj
@  qsxlzrlq srm@sample.com 2025-09-07 20:37:28 14f4dbc3
│  (empty) (no description set)
○  zmnvvqnv srm@sample.com 2025-09-07 20:37:26 **other-bookmark** git_head() d7dec052
│  (empty) (no description set)
○  psnqyolu srm@sample.com 2025-09-07 20:37:18 2bdb03ea
│  (no description set)
◆  kuznnowx srm@sample.com 2025-09-07 17:20:58 **main** c0e7ef80
│  chore: refactor process, view now query data (#10)

The prompt is : 
myproject: other-bookmark🠕1 [0m +0-0]

Other example : 
➜  jj
@  qsxlzrlq srm@sample.com 2025-09-07 20:37:28 14f4dbc3
│  (empty) (no description set)
○  zmnvvqnv srm@sample.com 2025-09-07 20:37:26 **other-bookmark second-bookmark** git_head() d7dec052
│  (empty) (no description set)
○  psnqyolu srm@sample.com 2025-09-07 20:37:18 2bdb03ea
│  (no description set)
◆  kuznnowx srm@sample.com 2025-09-07 17:20:58 **main** c0e7ef80
│  chore: refactor process, view now query data (#10)

The prompt is : 
myproject: other-bookmark🠕1 on second-bookmark🠕1 [0m +0-0]

nemith:
Now with #6671 landed this is now a breaking change and overlaps in functionality.  

This has been open for a month now and I am wondering what I did wrong or what the issue is?  I guess I will just close this out?

JanDeDobbeleer:
@nemith reaching out here to share what I just replied on the Discord as I can’t mention you so I’m assuming you quit the server. Let me know if you’re open to this, I understand if you’re not willing to reply, but if at any moment you do, don’t hesitate.

image

nemith:
I am no longer interested in working on this.   I am no longer using omp.  

Best wishes.

JanDeDobbeleer:
@nemith I understand. If at any moment you feel differently, don't hesitate to get in touch.

Gravifer:
I'm interested in reviving this; can anyone catch me up?

JanDeDobbeleer:
@Gravifer I'm not sure @nemith will reply. What we need is a simple, straightforward solution. I really don't want this struct to become overly complicated as I'm not at home in `jj` and if everyone else quits, I'll still be here maintaining. We can add useful functionality, as long as we're smart about it. That's what I meant originally.

nemith:
To be clear when I had this PR open before I asked multiple times about the approach , I trued to get feedback and look on IRC.  I saw PR after PR get accepted and mine sat on the review queue for months,  mostly alone.  It was when an overlapping PR which was inefficient and didn't actually solve the problem that I lost it.  

It was on discord where my PR was mentioned in passing only as a mess and never got any real feedback.  I don't remember the exact wording.  

I was willing to do anything @JanDeDobbeleer wanted but and was willing to change my approach.  Before coding this up I asked and tried to get an opinion on a approach.  After not getting much I decided to try to code something up that allowed for a lot more information  but I also didn't want so many exec syscalls JJ to slow down for every potencial piece of information that I (ab)used the template feature of JJ.  

I agree there could have been other approaches and understand the burden of maintaining code.  I would have been extremely happy to rework this with any sort of feedback on the tradeoffs of maintaining and not but I got NOTHING abut a passing comment on Discord that I wasn't supposed to see.

As I stated before I have stopped using ohmyposh and anyone is free to pick this up and do whatever they want with it and I need zero credit.   Please don't tag me again.

JanDeDobbeleer:
I understand, part of the reason this wasn't addressed earlier is that I couldn't get my head around it as I was occupied building my career at that time. Your work remains valuable as it's still being looked at and requested, and I'm genuinely sorry my comment threw us in this situation, that was never the intent. But anyways, @Gravifer feel free to have a go at this.

Gravifer:
I ran into [senekor's jj tutorial](https://jj-for-everyone.github.io/), which I find to be a highly enjoyable read even for a git veteran    @JanDeDobbeleer 
(it also points to some other posh (pun intended) tools like [`mise`](https://mise.jdx.dev/) and [`mergiraf`](https://mergiraf.org/), which I've also found useful for my git-only projects)

PR #6728, opened 2025-09-07
fixed a space for bubblesline.omp.json theme for python virtual envir…

whxtedvrknxss:
There's an issue while using bubblesline theme. When creating python virtual environment the spacing is wrong, so i added extra space and it looks nice.
image


JanDeDobbeleer:
@whxtedvrknxss this is actually terminal bug, not an oh-my-posh spacing issue unfortunately. One terminal renders it OK, the other doesn't.

PR #6760, opened 2025-09-11
Document the need to upgrade bash on macOS

iandunn:
### Prerequisites

- [ ] ~~Tests for the changes have been added (for bug fixes / features).~~

### Description

This adds a tip to the macOS instructions about the need to upgrade bash to a recent version, and make it load by default. Knowing this up front would have saved me 2 hours of trial-and-error, searching, chatgpt-ing, etc.

It seems like every macOS user will need to do this, and I'd guess that's a large percentage of the userbase.

### DIFF
--- a/website/docs/installation/macos.mdx
+++ b/website/docs/installation/macos.mdx
@@ -42,6 +42,18 @@ sudo port install oh-my-posh
 
 Oh My Posh will be installed at `/opt/local/bin/oh-my-posh`.
 
+## Troubleshooting
+
+:::tip
+If you see an error that says, "[conditional binary operator expected][outdated-bash]" you probably need to update to a newer version of bash.
+
+```bash
+brew install bash
+grep -qxF "$(brew --prefix)/bin/bash" /etc/shells || sudo bash -c 'echo "$(brew --prefix)/bin/bash" >> /etc/shells'
+chsh -s "$(brew --prefix)/bin/bash" $USER
+```
+:::
+
 ## Updating
 
 ```bash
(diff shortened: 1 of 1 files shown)


website/docs/installation/macos.mdx
```diff
@@ -42,6 +42,18 @@ sudo port install oh-my-posh
...
 Oh My Posh will be installed at `/opt/local/bin/oh-my-posh`.
 
+## Troubleshooting
+
+:::tip
+If you see an error that says, "[conditional binary operator expected][outdated-bash]" you probably need to update to a newer version of bash.
+
+```bash
+brew install bash
+sudo bash -c 'echo "$(brew --prefix)/bin/bash" >> /etc/shells'
```
oalders: Thank you so much for taking the time to document this. 🙏🏻 I wonder if we could make this slightly more robust via something like: 

```
grep -qxF "$(brew --prefix)/bin/bash" /etc/shells || sudo bash -c 'echo "$(brew --prefix)/bin/bash" >> /etc/shells'
```

It's possible you might have the correct entry in `/etc/shells` but you still need to run `chsh`. Could golf that down further so that `brew --prefix` only runs once, but that may be overkill.

iandunn: That's a great idea, thanks! I added it 👍🏻

JanDeDobbeleer:
@all-contributors please add @iandunn for doc

This is pure magic 🪄! @holopin-bot @iandunn wizard

iandunn:
@JanDeDobbeleer Does this work for you at https://ohmyposh.dev/docs/installation/macos ?

For me, the `Troubleshooting` link appears in the sidebar, but the content is hidden. I tested in Safari too in case my Chrome extensions were causing it.

### Live

Screenshot 2025-09-12 at 12 21 50 PM

### Fixed in devtools

Manually adding `display: block` to `div.tabItem_Ymn6` makes it appear:

Screenshot 2025-09-12 at 12 20 42 PM

JanDeDobbeleer:
@iandunn it's broken indeed. I didn't check 🫣

iandunn:
I opened #6770 to track the bug.

PR #6788, opened 2025-09-17
feat(theme): material: show active Python venv

Ashfaqbs:
Adds a minimal python segment that renders only when a virtual environment is active (display_mode=environment) and disables folder-name fallback to avoid confusing folder-as-env rendering. The segment sits after the path and before git, matching material's minimal style.


JanDeDobbeleer:
@Ashfaqbs thanks for the suggestion. Unfortunately we don't accept changes to themes outside of fixes.

PR #6875, opened 2025-10-08
Jujutsu closest bokmakrs

ADIX7:
### Prerequisites

### Description

This PR extends the `jujutsu` component with the ability to display the closest bookmark, that is usually the git branch the user currently working on.

Screesnhot:

image

### DIFF
--- a/src/segments/jujutsu.go
+++ b/src/segments/jujutsu.go
@@ -35,10 +35,23 @@ func (s *JujutsuStatus) add(code byte) {
 }
 
 type Jujutsu struct {
-	Working  *JujutsuStatus
-	ChangeID string
-	ClosestBookmarks string
+	Working          *JujutsuStatus
+	ChangeID         string
+	closestBookmarks string
 	Scm
+}
+
+func (jj *Jujutsu) ClosestBookmarks() string {
+	if jj.closestBookmarks == "" {
+		statusString, err := jj.getJujutsuCommandOutput("log", "-r", "heads(::@ & bookmarks())", "--no-graph", "-T", "bookmarks")
+		if err != nil {
+			return ""
+		}
+		lines := strings.Split(statusString, "\n")
+		jj.closestBookmarks = lines[0]
+	}
+
+	return jj.closestBookmarks
 }
 
 func (jj *Jujutsu) Template() string {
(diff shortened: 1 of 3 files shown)


ADIX7:
I extended the default template but I can revert it if needed.

Also, for this functionality I had to add an other jj log call, so I'm not sure if I should add an other fetch flag.

src/segments/jujutsu.go
```diff
@@ -37,11 +37,12 @@ func (s *JujutsuStatus) add(code byte) {
 type Jujutsu struct {
 	Working  *JujutsuStatus
 	ChangeID string
+	ClosestBookmarks string
```
JanDeDobbeleer: Thinking if this should be a function rather than a field. That way you don't need a switch to enable this and it stays identical from a template POV. The downside is that the result is never cached and will always be called again during rendering of a transient prompt for example.

ADIX7: I pushed a change, it is now a function and also created a private field for caching. Is it valid, or is it unconventional for Go? I don't have much experience with Go...

src/segments/jujutsu.go
```diff
@@ -35,12 +35,25 @@ func (s *JujutsuStatus) add(code byte) {
...
-	ChangeID string
-	ClosestBookmarks string
+	Working          *JujutsuStatus
+	ChangeID         string
+	closestBookmarks string
 	Scm
 }
 
+func (jj *Jujutsu) ClosestBookmarks() string {
+	if jj.closestBookmarks == "" {
```
JanDeDobbeleer, line 45: it's cleaner to do an early return instead for readability:

```suggestion
	if jj.closestBookmarks != "" {
	    return jj.closestBookmarks
	}
	
	// actual logic
```

src/segments/jujutsu.go
```diff
@@ -35,12 +35,25 @@ func (s *JujutsuStatus) add(code byte) {
...
+		}
+		lines := strings.Split(statusString, "\n")
+		jj.closestBookmarks = lines[0]
+	}
+
+	return jj.closestBookmarks
+}
+
 func (jj *Jujutsu) Template() string {
 	return " \uf1fa{{.ChangeID}} \ue0a0{{.ClosestBookmarks}} {{if .Working.Changed}} \uf044 {{ .Working.String }}{{ end }} "
```
JanDeDobbeleer, line 58: I would still remove this from the default template, there's no need to call it when you don't want it.

JanDeDobbeleer:
resolved by #6871

JanDeDobbeleer:
@all-contributors please add @ADIX7  for doc, code

This is pure magic 🪄! @holopin-bot @ADIX7 wizard

PR #7007, opened 2025-12-01
fix(bash): append hook to PROMPT_COMMAND instead of prepending

scop:
### Prerequisites

### Description

I don't know if there's a particular reason to prepend `_omp_hook` to `PROMPT_COMMAND` as opposed to appending, but in case it doesn't matter: 

Provides for better coexistence with things that work on `PROMPT_COMMAND` as a scalar. They'll effectively access the array's first element only, and if we prepend an element, they won't be able to find their additions if they for example want to remove them.

https://gist.github.com/scop/210266a4351ecdb34ce7f2b6efb31750

### DIFF
--- a/src/shell/scripts/omp.bash
+++ b/src/shell/scripts/omp.bash
@@ -158,7 +158,7 @@ function _omp_install_hook() {
         prompt_command+=("$cmd")
     done
 
-    PROMPT_COMMAND=(_omp_hook "${prompt_command[@]}")
+    PROMPT_COMMAND=("${prompt_command[@]}" _omp_hook)
 }
 
 _omp_install_hook


src/shell/scripts/omp.bash
```diff
@@ -158,7 +158,7 @@ function _omp_install_hook() {
         prompt_command+=("$cmd")
     done
 
-    PROMPT_COMMAND=(_omp_hook "${prompt_command[@]}")
+    PROMPT_COMMAND=("${prompt_command[@]}" _omp_hook)
```
scop, line 161: By the way, `PROMPT_COMMAND` as an array is a [bash >= 5.1 feature](https://cgit.git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=bash-5.1#n97). I don't know what versions of bash Oh My Posh is intended to work with, but thought I'd mention it.

JanDeDobbeleer: Always latest, I don't want to support outdated software

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @scop for code

This is pure magic 🪄! @holopin-bot @scop wizard

PR #7213, opened 2026-01-15
feat: faster fvm flutter version detection

EscapeTo5D:
## What
  Read Flutter version from `.fvmrc` file instead of running `fvm flutter --version` command for faster detection.

  ## Why
  Terminal command execution is slow, reading file directly is much faster.

  ## How
  - Parse `.fvmrc` file (primary)
  - Parse `.dart_tool/version` (fallback)
  - Validate fvm environment with `.fvm` folder check

### DIFF
--- a/src/segments/flutter.go
+++ b/src/segments/flutter.go
@@ -1,4 +1,5 @@
 package segments
+import "errors"
 
 type Flutter struct {
 	Language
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
@EscapeTo5D will this file always be there? I feel a fallback to the old behaviour might still be a good idea.

JanDeDobbeleer (changes requested):
See my previous comment

PR #7215, opened 2026-01-16
fix(segments): fix winget segment header indexing

stmach:
### Prerequisites

### Description

The Winget segment does not work reliably when `winget upgrade` takes some time to complete. In that case `winget upgrade` emits some visual sugar (spinning `- \ | / -` chars and loading bars) to `stdout` before producing the desired output. All these [control] characters can end up in the header line which is assumed to just be the header fields.

In my case, printing `headerLine` with `%q` produces:
```
"- \r   \\ \r   | \r                                                                                                                        \r\r   - \r   \\ \r   | \r   / \r                                                                    
                                                    \rName               Id                 Version          Available        Source\r"
```

Within `parseWinGetOutput`, "Name" is assumed to be at index 0 while other header indices are taken from the actual string. This leads to all data lines being considered "too short" as they lack the gibberish prepended to the header line.

This PR fixes this by calculating the header field indices relative to the "Name" header, which matches the current behavior in the absence of any extra characters before the header line.

### DIFF
--- a/src/segments/winget.go
+++ b/src/segments/winget.go
@@ -74,10 +74,11 @@ func (w *WinGet) parseWinGetOutput(output string) []WinGetPackage {
 	}
 
 	// Determine column positions from header (calculated once)
-	idIndex := strings.Index(headerLine, "Id")
-	versionIndex := strings.Index(headerLine, "Version")
-	availableIndex := strings.Index(headerLine, "Available")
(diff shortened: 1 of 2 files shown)


JanDeDobbeleer:
@all-contributors please add @stmach for code

This is pure magic 🪄! @holopin-bot @stmach wizard

Issue #6094, opened 2025-01-14
Lazy loading segments or async execution


maratori:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see added?

I work with a large monorepo where running `git status` takes around 4 seconds to complete. Currently, Oh-My-Posh is not designed for handling such scenarios. To work around this, I’ve used the `untracked_modes` property (alternatively, `ignore_status` property can be used):
```yaml
version: 3
blocks:
  - type: prompt
    segments:
      - type: git
        style: plain
        properties:
          untracked_modes:
            /path/to/monorepo: no
```

While this approach improves performance, it comes at the cost of losing some key pieces of information in the prompt.

Meanwhile, Oh My Zsh handles such cases seamlessly by leveraging an [async prompt](https://github.com/ohmyzsh/ohmyzsh/blob/7a3695aadf51aaa24e79175b1684d4836e31956d/lib/async_prompt.zsh).

I propose adding similar asynchronous support to Oh-My-Posh. The key idea is as follows:
- If the execution time of a segment exceeds a configurable threshold (e.g., X milliseconds):
  - Display the prompt immediately with cached output for the slow segment.
  - Continue executing the segment in the background and update the cache when the result is ready.
  - Use the updated result from the background process in the next prompt rendering.

This enhancement, especially when combined with periodic refreshing (#5036), would significantly improve the user experience.
Additionally, it could serve as an alternative to p10k-like instant prompts (#5310).

JanDeDobbeleer:
@maratori this is a very challenging feature as oh-my-posh isn't shell native. It can be achieved, but nowhere near easily. And unfortunately that's one of these features where I simply don't have the time until this project's funding allows me to partially quit my day job.

maratori:
> this is a very challenging feature ...

Yes, I agree.

> ... as oh-my-posh isn't shell native.

Can you please clarify what you mean and how it relates to this feature?

> And unfortunately that's one of these features where I simply don't have the time until this project's funding allows me to partially quit my day job.

I completely understand your time constraints. Perhaps I could work on implementing this feature. However, we would need to design it first, as there are several key design decisions to address before creating a PR.

JanDeDobbeleer:
> Can you please clarify what you mean and how it relates to this feature?

[...]

PR #7244, opened 2026-01-24
feat(daemon): add background daemon for faster prompt rendering

po1o:
### Prerequisites

### TLDR

This adds a daemon capability in zsh, pwsh, fish and Bash(with ble.sh) for asynchronous rendering.
I have been careful to minimize the changes to the core so that the existing mode (non-daemon) remains unchanged. 
By default the daemon is not activated. To activate it use --daemon on init. 
On very large git repo, this is a *huge* performance benefit. The prompts always render in 100ms and finishes async.
I have been using it on macOS + Linux in zsh for a few weeks and it works fine. I fixed all the bugs I encountered. 
I don't have a windows machine, so I haven't been able to test with windows other than go test. 
For full disclosure, I used ClaudeCode and Gemini to iterate quickly but I reviewed all the code. 
Please review the change to .github/ and .devcontainer/ to ensure this is fine. 
This change opens the door to many possibilities :-) like adding notifications capabilities, etc... 

### Description

This commit introduces a daemon mode for Oh My Posh that dramatically improves prompt rendering performance by running a persistent background process. Instead of spawning a new process for each prompt, the shell communicates with the daemon via gRPC over Unix domain sockets (or named pipes on Windows).

The daemon (`src/daemon/`) consists of several key components:

- **daemon.go**: Core daemon implementation with gRPC server, session management, and render orchestration. Uses a single master config (prototype) that is cloned for each request to ensure isolation.

- **ipc/**: Inter-process communication layer using Protocol Buffers and gRPC.
  - `daemon.proto`: Defines the service interface with streaming RenderPrompt, ToggleSegment, CacheClear, CacheSetTTL, and CacheGetTTL RPCs.
  - `protocol.go`: Handles conversion between proto messages and runtime types.
  - `socket.go`/`socket_unix.go`/`socket_windows.go`: Platform-specific socket path resolution and cleanup.

- **client.go**: gRPC client used by the CLI to communicate with the daemon. Handles connection management and provides a clean API for render requests.

- **cache.go**: In-memory cache with TTL support and multiple caching strategies:
  - `AsyncRendering`: Always recomputes but uses cached value for pending display
  - `Session`: Cached for the session lifetime, cleaned when session ends
  - `Folder`: Invalidated when the current directory changes

- **session.go**: Tracks active shell sessions by PID with platform-specific process monitoring (session_linux.go, session_macos.go, session_windows.go). Automatically cleans up when shell processes exit.

- **config_cache.go** & **config_watcher.go**: Caches parsed configurations and watches config files for changes using fsnotify. Automatically reloads config when modified.

- **lock.go**: Ensures single daemon instance using lock files with PID tracking.

The daemon supports streaming prompt updates:
1. When a render request arrives, the daemon starts rendering all segments
2. After `daemon_timeout` milliseconds (default 100ms), partial results are returned
3. Segments still computing show a pending indicator (configurable icon/background)
4. As each segment completes, updates stream back to the shell
5. The shell script updates the prompt in-place as updates arrive

Updated shell scripts (`src/shell/scripts/`) for bash, zsh, fish, and PowerShell:
- Check for daemon availability and fall back to direct rendering
- Parse streaming output format (prefixed lines: `primary:`, `right:`, `status:update|complete`)
- Handle prompt updates with cursor repositioning for smooth in-place updates
- Start daemon automatically on shell init with `--daemon` flag

- **Toggle command** (`src/cli/toggle.go`): When daemon is running, toggle requests are sent to the daemon via gRPC `ToggleSegment` RPC. The daemon maintains toggle state per session in its memory cache, ensuring toggles persist across prompts without file I/O.

- **Cache command** (`src/cli/cache.go`): Extended to support daemon mode:
  - `cache clear`: Clears daemon's in-memory cache via `CacheClear` RPC
  - `cache ttl [days]`: Gets/sets daemon cache TTL via `CacheGetTTL`/`CacheSetTTL` RPCs
  - Falls back to file-based cache when daemon is not running

New config options in `src/config/config.go`:
- `daemon_timeout`: Milliseconds before returning partial results (default: 100)
- `daemon_idle_timeout`: Minutes to keep daemon alive with no sessions (default: "5", "none" to disable)
- `render_pending_icon`: Icon for segments still loading (global and per-segment)
- `render_pending_background`: Background color for pending segments (global and per-segment)

New `daemon` subcommand (`src/cli/daemon.go`):
- `oh-my-posh daemon start`: Start daemon in background (detached)
- `oh-my-posh daemon status`: Check if daemon is running
- `oh-my-posh daemon serve`: Run daemon in foreground (for debugging)
- `oh-my-posh init <shell> --daemon`: Initialize shell with daemon support

- Created `website/docs/advanced/daemon.mdx`: Comprehensive daemon documentation
- Updated `website/docs/configuration/general.mdx`: Added daemon config fields
- Updated `website/docs/configuration/segment.mdx`: Added per-segment pending options
- Updated `website/docs/installation/prompt.mdx`: Added daemon mode tip
- Updated `website/sidebars.js`: Added daemon to Advanced section

- **`.github/workflows/build_code.yml`**: Added `Install Protoc` step using bootstrap-protoc action
- **`.github/workflows/code.yml`**: Added `Install Protoc` step for test/lint workflow
- **`.github/workflows/composite/bootstrap-protoc/action.yml`**: New composite action that:
  - Downloads protoc for Linux/macOS/Windows
  - Installs Go protobuf plugins (protoc-gen-go, protoc-gen-go-grpc)
  - Runs `go generate ./...` to generate protobuf code

- **`.devcontainer/Dockerfile`**: Added protobuf tooling for development:
  - Installs `protobuf-compiler` via apt
  - Installs Go protobuf plugins (protoc-gen-go, protoc-gen-go-grpc)

Comprehensive test coverage in `src/daemon/`:
- Unit tests for cache, config watcher, lock file, sessions
- Integration tests for full daemon lifecycle
- Benchmark tests comparing daemon vs direct rendering
- Shell integration test scripts for bash, zsh, fish, and PowerShell


po1o:
Took me a few iterations for the Windows tests and the FreeBSD 32/64 builds , but now it seems all green! 
I don't have a Windows machine so I could not manually test on Windows + PowerShell. I think s/o would need to test run a bit.

po1o:
This PR should address https://github.com/JanDeDobbeleer/oh-my-posh/issues/6094

JanDeDobbeleer:
@po1o this is one of these thing where I would love to have been able to sync up front about. It's appreciated, but this is a significant addition that requires a design conversation up front. I was experimenting with this before and have one question. In the design I see you still use `oh-my-posh render` to access the daemon, how is this faster as the executable startup is where the bottleneck is at, not the execution?

po1o:
@JanDeDobbeleer thanks for taking a look. 

Yeah it is a big PR unfortunately. And as GitHub does not (yet) natively support stacked PRs, it is not easy to break it down in multiple smaller ones :-( 

The executable startup is not where the bottleneck is. It really is on the execution when segments may take a long time. 

This PR is not meant to be faster than "omp print" in all cases. It is meant to improve the latency of the prompt rendering by bounding the time to first byte within a predictible ~100ms (configurable). The goal is to make the terminal snappier. It is not about which one *is* faster but which one *feels* faster.

"omp render" is simply a gRPC client. Creating this client process is not going to be faster than creating an "omp print" process. The fork() cost remains the same.

"omp render" does not do any processing. It sends the necessary data to the daemon and waits for a response. The daemon will always return an answer within 100ms (configurable).

So technically, if the prompt can be rendered below 100ms, "omp print" will be slightly faster than the "omp render" + daemon. The difference is going to be the gRPC overhead over the unix socket. However, as per my research the cost of a gRPC call over a Unix socket is 0.1ms to 0.3ms. So the overhead of the daemon for cases under 100ms is going to be ~1-2ms all things considered. For a prompt rendering, it is not noticeable. And this may be offset by the in-memory caching when the cache is used.

Where things start to get interesting is when the prompt takes longer to render than the 100ms barrier (and for folks that work with a large git repo, it is not uncommon to have 2+ seconds)

In the case of "omp print", as it is synchronous, the latency is cascaded to the whole prompt rendering. When you get beyond 200-300ms you start "feeling" it. 

With the "omp render" + daemon, the daemon times out at 100ms but does not stop the computation of the not-finished segment. It sends whatever is finished after 100ms of running and uses the previous value from the cache with an hourglass prepended for the ones still being rendered. It also tells the client: "I am not finished". The client passes the prompt to the shell for display. 

=> At this point the user can use the shell. The hourglass (and configurable background color) is a visual cue for the user that this segment in the prompt is still pending.

In parallel, the client waits for updates from the daemon. Everytime the client receives an update, it triggers a prompt redraw until the daemon is finished. This has not impact on the user even if they started writing something.

The main reason why this is a daemon rather than a simple binary doing the computation is because we can cache everything in memory. 
So we can cache all the previous values of all the segments at little cost. This is used to display the not-yet-finished segments with their previous value until they are updated; the same way powerlevel10k is doing it for zsh. 
And for segments with a cache strategy,  the daemon is going to be much faster than the disk-based cache because everything is in memory. (Additional benefit, no disk wear).

po1o:
_Note:_ While working on a large repo, I noticed one of my last commit to fix lint issues regressed the async functionality in ways that the tests did not catch. I need to fix and improve the tests.

po1o:
Here is an example on the chromium git repo.

Standard oh-my-posh:
sync

Daemon-based oh-my-posh:
async

Daemon-based oh-my-posh showing it is not needed to wait to enter new commands: 
async2

po1o:
> _Note:_ While working on a large repo, I noticed one of my last commit to fix lint issues regressed the async functionality in ways that the tests did not catch. I need to fix and improve the tests.

I'll send a fix later this weekend.

po1o:
Fixed the regression. Added tests to catch it. Tested on macos + zsh (see above)

JanDeDobbeleer:
@po1o I started an alternative implementation. The reason is that I feel this doesn't really need a separate grpc deamon with all this logic, it can be handled more pragmatically. I could be wrong, but I'm an atheist so I want to see it first.

po1o:
@JanDeDobbeleer ok, sounds good! Let me know if I can help.

I call this a "daemon" but it's not a daemon in the Unix/Linux sense of the term. It gets started only when it's needed the first time a shell is started if there is no instance already running and then moves to detached mode. There is only a single instance: it can serve all the requests from the different shells concurrently; it is thread-safe. If there is no shell process anymore, it will stop after a period of idle time. 

gRPC is crazy fast and already provides the streaming capability but anything that handles streaming would work. protobuf generated ser/de code usually outperforms anything else but for such small payload, JSON may be fast enough (maybe faster?).

One of the complexity I stumbled upon in the "daemon" implementation was because of the global states in the existing code. It is totally normal as it is but when moving to serving requests concurrently it needed some adjustment to make it thread safe and avoid race conditions. If you can find a better approach that does not need to protect shared resources, the code will be much simpler. I tried to avoid synchronization points by moving to dedicated data structured per request. 
An example: The writer in the segment holds buffered output state tied to the config. Multiple prompts simultaneous rendering could interleave or corrupt the shared writer buffer without mutex protection.

Thanks a lot for looking at this! Much appreciated.

po1o:
@JanDeDobbeleer FYI only, I also have a vim-mode implementation on top of the daemon mode that should also work in non-daemon mode. It is quite tricky because it repaints the **current** prompt. And with the async mode, it needs to do it while not interfering with the async computation.

I haven't pushed it yet because I built it on top of the daemon mode for now and I still have a small rendering bug that I need to look into. But its quite functional. See also how it changed the cursor shape like vim does. 

vim-mode

JanDeDobbeleer:
@po1o I have a working prototype without a "deamon" and GRPC. Will push it out later this week so you can have a look and play around. It leverages the existing parallel goroutines and settings instead. Now, credit goes where it's due, I wouldn't have been able to figure this out without this PR.

po1o:
Awesome @JanDeDobbeleer. No problem :) More than happy to code review when ready.

JanDeDobbeleer:
@po1o It works but I still have a race condition with the rendering I'm trying to solve. I'll push as soon as I have a semi stable solution.

luisdavim:
I love the idea and it seems like with something like this we'll finally have some prompt that can get close to `powerlevel10k` in terms of (perceived) speed but I wonder, instead of a socket and gRPC, why not just a routine for the "daemon" and a channel for the communication, wound't that be a simpler architecture and maybe even faster?

I also like the idea that this can be used for other scenarios, like for example keeping the active prompt updated across multiple tmux panes or terminals by periodically re-rendeding and sneding an updated version.

JanDeDobbeleer:
@luisdavim yes, my implementation reuses existing goroutines that respect the timeout on segments and allows for subsequent prompts. I just have a serious bug in PowerShell where refreshing the prompt in fast, short sequences breaks the rendering.

luisdavim:
I see, I do like the idea of a central controller that manages the timeouts and refresh.

po1o:
@JanDeDobbeleer what do you mean by respect segment timeouts? Do you mean that instead of having a single time barrier, this can be further customized per segment with the timeout property at segment level? 

@luisdavim the daemon is likely to be much faster. All the caching is done in memory, no disk access. The config is read once and kept in memory unless it is changed (json parsing is costly). The daemon evaluates everything at once (transient, right and left prompt) instead of calling 3x omp print and therefore parsing 3x the config per prompt as in the current implementation (this might change in the new async version; I haven't seen the code yet)

That being said, it is also much more complex and for sthg as used as oh-my-posh, I agree that simpler is better imho. We could keep the daemon as an experimental thing for later potentially.

luisdavim:
I do like the idea of the daemon, what I was talking about was the communication, replacing the gRPC part with a channel.

po1o:
You mean Go channels? If yes, they can't be used for IPC

luisdavim:
Right but why separate processes?

po1o:
I may not understand your question; sorry if that is the case and if my answer is totally off-base.

The shell spawns an oh-my-posh process to render the prompt. At that point you have two processes. These two processes need to communicate with each other: the shell gives parameters; oh-my-posh returns a prompt.

Then you have two options: 
- oh-my-posh exits when it is finished. The shell spawns a new instance for the next prompt with new parameters. (That's how it works right now).
- oh-my-posh remains active and the shell needs to communicate with it somehow to pass new parameters for each prompt request and receive the rendered prompt each time. That is where you need some form of IPC. If you have one oh-my-posh instance per shell, maybe something like a pair of pipes could be enough with a very simple text based protocol. If you have a single oh-my-posh instance for multiple shells (what this daemon implementation does), that's where you need something more elaborate and a Unix socket is usually the way to go. And then a proper protocol is needed and I chose gRPC because I did not want to reinvent the wheel.

I hope this answers the question

luisdavim:
Yes, I see. I was thinking there was already some mechanism for the shell to interact with the running process 🤦 

In that case I do think that this approach is the best. It opens the potential for further expansion and solves the problem that so far only `powerlevl10k` managed to solve, I actually don't know the approach used there...

I think a central render timeout, letting the segments do their thing on their own time and then update when they are ready is what we need.

luisdavim:
I guess another option is to spawn the main process, start rendering and on timeout. Spawn a daemon that will continue the work, and return waht we got. Then the daemon triggers the update and exits when ready. But that has a bigger overhead.

And with the central persistent daemon we can multiplex and serve multiple clients from different shell sessions (separate terminals or tmux panes)...

po1o:
> It opens the potential for further expansion and solves the problem that so far only powerlevl10k managed to solve, I actually don't know the approach used there...

P10k never leaves the shell because it is purely written in zsh (with the exception of a custom made  gitstatus process to replace git status)

JanDeDobbeleer:
I disagree with the GRPC design though. Way too much overhead if you still need an executable anyways to start from. In my implementation, which I'm almost ready to share as POC, it starts just one executable run which can stream the updates allowing to reload the prompt on subsequent updates. It leverages the existing segment timeout setting as you can then decide for yourself which data you want to see at all times, and wait for it, and which not.

po1o:
For the daemon additional complexity: +1 

Additional overhead: I would not be so sure about that. 

Among other things the daemon caches the config. Compared to a gRPC exchange of ~1k parsing a 3k json (which is the size of my config) is 10x more costly in terms of CPU cycle, not to mention the disk I/O cost. And omp does it at every prompt if I am not mistaken.

But I am not arguing for the daemon, the complexity is pbbly not worth it. Simpler is better. 

Thanks again for looking into this @JanDeDobbeleer

JanDeDobbeleer:
@po1o see #7306

JanDeDobbeleer:
@po1o closing this one, will continue improving streaming.

JanDeDobbeleer:
@all-contributors please add @po1o for code

This is pure magic 🪄! @holopin-bot @po1o wizard

PR #7251, opened 2026-01-26
perf: use CutPrefix/Suffix instead of Has+Trim

scop:
### Prerequisites

### Description

Avoid repeated Has+Trim string operations by using relevant Cut*. Should be good for performance (no numbers), but also arguably simplifies code.

### DIFF
--- a/src/segments/git.go
+++ b/src/segments/git.go
@@ -283,18 +283,18 @@ func (g *Git) Commit() *Commit {
 			refs := strings.SplitSeq(line, ", ")
 			for ref := range refs {
 				ref = strings.TrimSpace(ref)
-				switch {
-				case strings.HasSuffix(ref, "HEAD"):
+				if strings.HasSuffix(ref, "HEAD") {
 					continue
-				case strings.HasPrefix(ref, "tag: refs/tags/"):
-					g.commit.Refs.Tags = append(g.commit.Refs.Tags, strings.TrimPrefix(ref, "tag: refs/tags/"))
-				case strings.HasPrefix(ref, "refs/remotes/"):
-					g.commit.Refs.Remotes = append(g.commit.Refs.Remotes, strings.TrimPrefix(ref, "refs/remotes/"))
-				case strings.HasPrefix(ref, "HEAD -> refs/heads/"):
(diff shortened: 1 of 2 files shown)


src/segments/git.go
```diff
@@ -283,18 +283,18 @@ func (g *Git) Commit() *Commit {
...
-					g.commit.Refs.Remotes = append(g.commit.Refs.Remotes, strings.TrimPrefix(ref, "refs/remotes/"))
-				case strings.HasPrefix(ref, "HEAD -> refs/heads/"):
-					g.commit.Refs.Heads = append(g.commit.Refs.Heads, strings.TrimPrefix(ref, "HEAD -> refs/heads/"))
-				case strings.HasPrefix(ref, "refs/heads/"):
-					g.commit.Refs.Heads = append(g.commit.Refs.Heads, strings.TrimPrefix(ref, "refs/heads/"))
-				default:
+				}
+				if tag, found := strings.CutPrefix(ref, "tag: refs/tags/"); found {
+					g.commit.Refs.Tags = append(g.commit.Refs.Tags, tag)
+				} else if remote, found := strings.CutPrefix(ref, "refs/remotes/"); found {
```
JanDeDobbeleer, line 291: This goes against the conventions. No if/else allowed unless there's no other way.

JanDeDobbeleer (changes requested):

src/segments/git.go
scop: Known, but I think if/else is the best that can be done here. Switch doesn't work, no other good suggestions spring to mind.

JanDeDobbeleer: I have an idea, will refactor later this week 👍🏻

scop: Hm, this is closed now, but I don't see corresponding changes applied in main or next yet. Was there a decision that the (possible/likely/small) performance improvements are less important than if/else avoidance?

While not that interesting, at least the fonts change part of this PR would not have had the if/else issue.

Did I miss something?

JanDeDobbeleer: @scop the closing was likely a side effect of something else and not a deliberate close on my end (at least not manual).

PR #7255, opened 2026-01-28
Add the ability to install local theme's and make loading the configuration extension agnostic.

IsaacFG2:
### Prerequisites

- [ ] The commit message follows the [conventional commits][cc] guidelines.

### Description

I've made it so user's can add local configuration under `/usr/share/oh-my-posh/themes` and `~/.config/oh-my-posh/themes`.
I've also made it so that, when loading the configuration without the full path(locally on specified location's), or when a theme is not locally present, and is downloaded online, the extension will automatically get trimmed.
I've also made it install theme's under `~/.config/oh-my-posh/themes` instead of in the cache(`~/.cache/oh-my-posh/themes`).
I've also made the small change to make the `This directory is NOT in your $PATH!` warning to include quote's and i also added a `:` to the end of the first line of the warning.
I'd like to note i tested EVERYTHING, and i mean EVERYTHING.
Also, sorry for not following Conventional Commit's, in later pull request's i will do it.

### DIFF
--- a/src/build/version.go
+++ b/src/build/version.go
@@ -2,5 +2,5 @@ package build
 
 var (
 	Date    string
-	Version = "0.0.0-dev"
+	Version = "29.1.0"
 )
--- a/src/build/version.go
+++ b/src/build/version.go
@@ -2,5 +2,5 @@ package build
 
 var (
 	Date    string
-	Version = "29.1.0"
+	Version = "0.0.0-dev"
 )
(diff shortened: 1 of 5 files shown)


JanDeDobbeleer:
@IsaacFG2 thank you for the contribution, however, this is a change I won't be accepting. The reason is that I do not want this hardcoded convention as themes should not get such a prominent place (except in the Windows installer where we keep it for historical reasons). Themes act as inspiration, and are ideally not used out of the box.

IsaacFG2:
Okay, i understand this decision.

PR #7258, opened 2026-01-29
fix(theme): robbyrussell git dirty indicator and status counters

maxvictor:
## Description

This PR fixes and enhances the `robbyrussell` theme to match the original oh-my-zsh behavior for git dirty state indication, and adds detailed git status counters.

### Problem

The current oh-my-posh `robbyrussell` theme shows the ✗ indicator only when the **last command fails** (exit code ≠ 0). However, the [original robbyrussell theme in oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/robbyrussell.zsh-theme) shows a **yellow ✗ when the git repository has uncommitted changes**.

### Solution

- Added `fetch_status: true` to enable git status detection
- Added git status counters showing added (+), modified (~), and untracked (?) files
- Added conditional ✗ indicator when `Working.Changed` or `Staging.Changed` is true
- Removed the `status` segment (exit code) that wasn't part of the original theme
- Changed `options` to `properties` (correct property name)

### Before
```
➜ my-project git:(main) # No indicator even with uncommitted changes
```
### After
```
➜ my-project git:(main +2 ~5 ?1) ✗ # Shows file counters and ✗ when dirty
```

Where:
- `+2` = 2 files added
- `~5` = 5 files modified
- `?1` = 1 untracked file
- `✗` = indicates uncommitted changes

## Related

- Original oh-my-zsh theme: https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/robbyrussell.zsh-theme

### DIFF
--- a/themes/robbyrussell.omp.json
+++ b/themes/robbyrussell.omp.json
@@ -22,17 +22,12 @@
         {
           "foreground": "#D0666F",
           "options": {
-            "branch_icon": ""
+            "branch_icon": "",
+            "fetch_status": true
           },
           "style": "plain",
-          "template": " <#5FAAE8>git:(</>{{ .HEAD }}<#5FAAE8>)</>",
+          "template": " <#5FAAE8>git:(</>{{ .HEAD }}{{ if .Working.Changed }} <#FFEB3B>{{ .Working.String }}</>{{ end }}<#5FAAE8>)</>{{ if or .Working.Changed .Staging.Changed }} <#FFEB3B>\u2717</>{{ end }}",
           "type": "git"
-        },
-        {
-          "foreground": "#BF616A",
-          "style": "plain",
-          "template": " \u2717",
-          "type": "status"
         }
       ],
       "type": "prompt"


themes/robbyrussell.omp.json
```diff
@@ -12,7 +12,7 @@
         },
         {
           "foreground": "#56B6C2",
-          "options": {
+          "properties": {
```
Marukome0743: `properties` is old name.
It was changed to `options` at #7046.
Could you revert it?

themes/robbyrussell.omp.json
```diff
@@ -21,18 +21,13 @@
         },
         {
           "foreground": "#D0666F",
-          "options": {
-            "branch_icon": ""
+          "properties": {
+            "branch_icon": "",
+            "fetch_status": true
```
JanDeDobbeleer, line 26: this should be not be added. The user should be in the lead to decide to fetch status as it will make the rendering slower.

JanDeDobbeleer (changes requested):

JanDeDobbeleer:
@all-contributors please add @maxvictor for design

This is pure magic 🪄! @holopin-bot @maxvictor wizard

PR #7288, opened 2026-02-08
fix timezone issues on android

abdalmoniem:
### Prerequisites

### Description

on android `go` always reports time in **`UTC`** zone, this PR fixes the issue by getting the correct timezone from the `persist.sys.timezone` system property

### DIFF
--- a/src/runtime/terminal.go
+++ b/src/runtime/terminal.go
@@ -38,7 +38,31 @@ type Terminal struct {
 	networks []*Connection
 }
 
+func initLoacl() {
+	var timeZone *time.Location
+
+	systemTimeZone, err := exec.CommandContext(context.Background(), "/system/bin/getprop", "persist.sys.timezone").Output()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Failed to get systemTimeZone from OS")
+
+		systemTimeZone = []byte("UTC")
+	}
+
+	timeZone, err = time.LoadLocation(strings.TrimSpace(string(systemTimeZone)))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to compute timeZone from systemTimeZone: %s\n", string(systemTimeZone))
+
+		timeZone = time.Now().UTC().Location()
+	}
+
+	time.Local = timeZone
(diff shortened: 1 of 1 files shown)


src/runtime/terminal.go
```diff
@@ -38,7 +38,31 @@ type Terminal struct {
...
+		fmt.Fprintf(os.Stderr, "Failed to compute timeZone from systemTimeZone: %s\n", string(systemTimeZone))
+
+		timeZone = time.Now().UTC().Location()
+	}
+
+	time.Local = timeZone
+}
+
 func (term *Terminal) Init(flags *Flags) {
+	if term.GOOS() == ANDROID {
```
JanDeDobbeleer, line 62: the timezone likely doesn't change often, which implies it's not relevant to get the timezone every time, we can cache the value on `oh-my-posh init` and set it when we have a cached value.

src/runtime/terminal.go
```diff
@@ -38,7 +38,31 @@ type Terminal struct {
 	networks []*Connection
 }
 
+func initLoacl() {
+	var timeZone *time.Location
+
+	systemTimeZone, err := exec.CommandContext(context.Background(), "/system/bin/getprop", "persist.sys.timezone").Output()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Failed to get systemTimeZone from OS")
```
JanDeDobbeleer, line 46: Do not print, log. This will ruin the prompt on every run if we can' t fetch this data.

src/runtime/terminal.go
```diff
@@ -38,7 +38,31 @@ type Terminal struct {
 	networks []*Connection
 }
 
+func initLoacl() {
+	var timeZone *time.Location
+
+	systemTimeZone, err := exec.CommandContext(context.Background(), "/system/bin/getprop", "persist.sys.timezone").Output()
```
JanDeDobbeleer, line 44: Has this been validated? Also, it needs to use the `cmd` package we have internally.

JanDeDobbeleer (changes requested):

PR #7308, opened 2026-02-13
refactor(config): use generic SegmentWriter for cache storage

Mertozturkk:
### Prerequisites
- [x] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [x] The commit message follows the [conventional commits][cc] guidelines.
- [x] Tests for the changes have been added (for bug fixes / features).
- [ ] Docs have been added/updated (for bug fixes / features).
### Description
- removed unnecessary JSON marshalling/unmarshalling for cache entries
- implemented generic cache.Get[SegmentWriter] and cache.Set
- resolved TODO regarding generic type indicator for SegmentWriter
- improved type safety and reduced serialization overhead

PR #7316, opened 2026-02-17
refactor(config): use generic cache for SegmentWriter

Mertozturkk:
### Prerequisites

### Description

> This is a reworked version of #7308, which was unintentionally closed during a base branch rebase. The changes have been rebased onto the latest `main` and address all review feedback from the previous PR.

Refactored the segment caching mechanism to use generic types instead of JSON serialization, resolving the TODO in `setCache`.

**Changes:**
- Replaced `json.Marshal`/`json.Unmarshal` with `cache.Get[SegmentWriter]` and `cache.Set` for direct interface storage
- Added graceful handling of legacy string cache entries (deletes old keys without error-level logging)
- Added `TestSegmentCache` unit test covering both generic cache and legacy data migration
- Added `defer cache.DeleteAll` to prevent test state leakage

### DIFF
--- a/src/config/segment.go
+++ b/src/config/segment.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"bytes"
+	"encoding/gob"
 	"encoding/json"
 	"fmt"
 	"slices"
@@ -412,15 +414,33 @@ func (segment *Segment) restoreCache() bool {
 	}
 
 	key, store := segment.cacheKeyAndStore()
-	data, OK := cache.Get[string](store, key)
+
+	data, OK := cache.Get[any](store, key)
 	if !OK {
 		log.Debugf("no cache found for segment: %s, key: %s", segment.Name(), key)
 		return false
 	}
 
-	err := json.Unmarshal([]byte(data), &segment.writer)
-	if err != nil {
-		log.Error(err)
+	switch v := data.(type) {
+	case []byte:
+		// Decode into the writer initialized by MapSegmentWithWriter instead of
+		// replacing it: the snapshot only carries exported fields, while the
+		// writer's unexported runtime state (env, options) must stay intact or
+		// any method relying on it panics after a restore.
+		if err := gob.NewDecoder(bytes.NewReader(v)).Decode(segment.writer); err != nil {
+			log.Error(err)
+			cache.Delete(store, key)
+			return false
+		}
+	case string:
+		// legacy JSON cache entry, remove it so it gets re-cached in the new format
+		log.Debugf("removing legacy cache key: %s", key)
+		cache.Delete(store, key)
+		return false
+	default:
+		log.Debugf("unexpected cache type for segment: %s, key: %s", segment.Name(), key)
+		cache.Delete(store, key)
+		return false
 	}
 
 	segment.Enabled = true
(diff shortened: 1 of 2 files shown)


src/config/segment.go
```diff
@@ -311,48 +311,40 @@ func (segment *Segment) restoreCache() bool {
...
-	return true
 }
 
 func (segment *Segment) setCache() {
 	if segment.restored || !segment.hasCache() {
 		return
 	}
 
-	// Never cache pending state to avoid polluting cache with incomplete data
-	if segment.Pending {
```
luisdavim, line 467: don't we want to keep this?

Mertozturkk: That check was accidentally removed during the refactor. I've restored it to ensure we don't pollute the cache with incomplete data when a segment is still in a pending state. Re-pushed the fix

JanDeDobbeleer:
@Mertozturkk this breaks when in streaming mode, didn't look at the details just yet but I can make the prompt fail now. I pushed the udpated branch, maybe I missed something silly.

JanDeDobbeleer:
@Mertozturkk there was a race condition introduced here which caused the behavior I noticed. I'll share my findings this week after more validation.

JanDeDobbeleer:
## Review report: why the original approach broke the prompt, and what changed

While validating this change, prompts sometimes rendered a template error, and in streaming mode nothing at all. This comment documents the root cause and the fixes, for future reference. The failure modes are subtle and worth remembering when touching the segment cache again.

### Issue 1: restoring a cached writer dropped its runtime state

The original revision stored `segment.writer` itself in the cache and, on a hit, assigned the cached value back with `segment.writer = v`.

The cache persists to disk through gob, which skips unexported fields. `segments.Base` keeps `env` and `options` in unexported fields, so a writer decoded from the device cache comes back with both set to `nil`. `restoreCache()` then replaced the freshly initialized writer from `MapSegmentWithWriter()` with that hollow copy, and every later code path touching `env` dereferenced `nil`:

- **Streaming mode** (`oh-my-posh stream`, used by the pwsh integration): `Execute()` continues after a cache hit and calls `writer.Enabled()` directly. The `nil` dereference panics inside the segment goroutine, which has no `recover`, so the process dies before printing anything. The user sees a blank prompt. Because the broken entry stays in the device cache until its TTL expires, every following prompt crashes the same way, with no way to recover short of clearing the cache. Reproduced with a cached git segment: the second `stream` run panics in `Git.Enabled()`.
- **Non-streaming mode**: `text/template` wraps method-call panics into an error, so a template such as `{{ .StashCount }}` renders `unable to create text based on template` instead of the segment content.

The pre-refactor JSON code avoided this by design: `json.Unmarshal(data, &segment.writer)` decodes into the existing writer, overlays the exported fields, and leaves `env` and `options` intact.

### Issue 2: the cached value was a live, shared object

`cache.Get` returns the same pointer that was stored. Render passes keep mutating the writer after `setCache()` runs (`SetText`, `SetIndex`), so the cached "snapshot" changed after the fact. In a long-lived process such as the serve daemon, two segments resolving to the same key would share one mutable writer across concurrent renders.

A smaller variant of the same problem: storing raw writers means the gob encode of the whole store file at close depends on every writer being encodable. One unencodable value fails the entire file. The JSON version serialized per segment, so a failure skipped that single segment and nothing else.

### The fix

`setCache()` now stores a gob-encoded `[]byte` snapshot of the writer, and `restoreCache()` decodes that snapshot into the writer `MapSegmentWithWriter()` initialized. Gob leaves fields that are absent from the stream untouched, so `env` and `options` survive the restore. This keeps the overlay semantics of the old JSON code while staying on the typed cache API.

- The snapshot is immutable; later mutations of the live writer can't leak into the cache.
- An encode failure logs and skips caching that one segment; the store file stays healthy.
- Legacy JSON entries (`string`) and raw-writer entries written by earlier revisions of this branch are deleted on first read and re-cached in the new format on the next render, so existing caches self-heal without user action.

### Verification

- Unit tests in `segment_cache_test.go` cover the round trip (asserting with `assert.Same` that restore reuses the initialized writer instance), snapshot immutability, and cleanup of legacy and unexpected entry types.
- A/B test against a binary built from the original revision, with a cached git segment: the old binary rendered a template error on the second run and panicked in streaming mode; the fixed binary restores from the snapshot, renders `{{ .StashCount }}` (which needs `env`) correctly, and recovers a cache poisoned by the old format.
- `go test ./...` and `golangci-lint run` pass.

JanDeDobbeleer:
@all-contributors please add @Mertozturkk for code

This is pure magic 🪄! @holopin-bot @Mertozturkk wizard

Issue #475, opened 2021-03-02
[wish] TMUX status line integration


tigerinus:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

The only reason I am still staying with Powerline is its integration with TMUX. I've been looking for an alternative to Powerline + TMUX + bash for a long time, but no luck. OMP has a cross-shell prompt customizer seems to be promising. It'd be nice to support TMUX.

JanDeDobbeleer:
@tigerinus what exactly do you mean by supporting TMUX? The way I understand it, TMUX is a multiplexer running a shell (which we already support).

tigerinus:
> @tigerinus what exactly do you mean by supporting TMUX? The way I understand it, TMUX is a multiplexer running a shell (which we already support).

TMUX has a customizable status line. The legacy powerline has a TMUX plugin that shares its segments within TMUX status line.
See https://powerline.readthedocs.io/en/latest/usage/other.html#tmux-statusline

bewing:
I'm interested in this as I'm currently only running `powerline-daemon` to render a tmux status line, and spent the last day or two of my vacation doing research.

Powerline supports a large amount of tmux versions -- 1.7 through 2.1+

tmux used to (pre-2.9) only support rendering a left bar, the window list (center), and a right bar.  Powerline's default config was to statically set the window list coloring during init, and then tell tmux to call powerline every 2 seconds for the left and right bar lines, as separate commands.  The default config only had the session name in the left bar, a powerline-like window list in the current colorscheme, and then uptime, load, time, date, and hostname in the right statusbar.

Newer tmux (2.9+) supports rendering the entire status line via the `status-format` array.  By default, `status-format[0]` keeps with the tmux standard of the left bar, the window list, and the right bar, but this could be overwritten.  

Given the current approach of posh regarding static shell inits without coloring information present, and only using the engine and renderer when rendering a status line, the best approach might be to only support tmux 2.9+ via the `status-format` line, and have tmux be a different "shell", with different config files for your standard shell and tmux.  Then we could tell tmux to completely delegate `status-format[0]` (and perhaps other lines, in a multi-line format) to omp.

I'm happy to slowly work on this if there is interest in my copious spare time, but given my lack of familiarity with the current codebase, would welcome any advice or guidance regarding the above.

ghost:
[...]

PR #7323, opened 2026-02-19
feat(tmux): Adds ability to use oh-my-posh for the tmux statusline

colings86:
### Prerequisites

### Description

This change adds functionality to allow the statusline in tmux to use oh-my-posh for rendering. This is similar to the claude statusline feature but in this case for tmux. There is a previous issue related to this where this kind of functionality was suggested: https://github.com/JanDeDobbeleer/oh-my-posh/issues/475

The feature consists of adding the following pieces:
1. `tmux.status-left` and `tmux.status-right` configuration sections to enable configuration of blocks and segments to be rendered on the left and right tmux statusline positions (which can support most segment types)
2. `tmux_session` segment type to render the current tmux session name (could also be configured on terminal prompt and claude statusline if running in tmux)
3. `tmux_window_list` segment type to render the window list for the current tmux session (could also be configured on terminal prompt and claude statusline if running in tmux)

### Current State

This PR contains the functionality, tests and documentation for the new functionality and has been manually tested in tmux sessions. However, before spending more time on this PR I wanted to raise it as a draft to get some feedback on whether this project is interested in adding this functionality and on the approach taken. Happy to work on the PR to get it to a state where it can be merged if this seems in line with the desired project directions. Please let me know what you think.


src/config/config.go
```diff
@@ -244,6 +245,22 @@ func (cfg *Config) migrateSegmentProperties() {
...
 		}
 	}
+
+	if cfg.Tmux == nil {
+		return
+	}
+
+	for _, block := range cfg.Tmux.StatusLeft.Blocks {
+		for _, segment := range block.Segments {
+			segment.MigratePropertiesToOptions()
```
JanDeDobbeleer: this is irrelevant as it's a new feature which starts after the migration.

src/config/config.go
```diff
@@ -244,6 +245,22 @@ func (cfg *Config) migrateSegmentProperties() {
...
+		}
+	}
+
+	for _, block := range cfg.Tmux.StatusRight.Blocks {
+		for _, segment := range block.Segments {
+			segment.MigratePropertiesToOptions()
```
JanDeDobbeleer: Sam comment as above

src/config/tmux.go
```diff
@@ -0,0 +1,12 @@
+package config
+
+// TmuxConfig holds the configuration for rendering tmux status bar sections.
+type TmuxConfig struct {
+	StatusLeft  TmuxStatusSection `json:"status_left"  yaml:"status_left"  toml:"status_left"`
+	StatusRight TmuxStatusSection `json:"status_right" yaml:"status_right" toml:"status_right"`
+}
+
+// TmuxStatusSection holds a list of blocks to render for one tmux status section.
+type TmuxStatusSection struct {
```
JanDeDobbeleer, line 10: This is never used for custom functions or am I missing something?

src/prompt/tmux.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return e.string()
+}
+
+var ansiSGRRe = regexp.MustCompile(`\x1b\[([0-9;]*)m`)
+
+// ansiToTmux converts ANSI SGR escape sequences to tmux format strings.
+// Tmux strips ESC (0x1b) bytes from #(command) output in status bars, so raw
+// ANSI sequences must be translated to tmux's native #[fg=...,bg=...] syntax.
+func ansiToTmux(s string) string {
```
JanDeDobbeleer: This is an interesting hack, can't we "just" extend how we colorize instead?

src/segments/tmux_window_list.go
```diff
@@ -0,0 +1,166 @@
...
+		})
+	}
+
+	return windows
+}
+
+// renderWindows builds the full powerline-connected ANSI string for all windows.
+// This method accesses terminal.Colors (read-only) and terminal.Plain but does NOT
+// call terminal.Write(), keeping it safe to call from concurrent goroutines.
+func (t *TmuxWindowList) renderWindows(windows []tmuxWindow, activeFG, activeBG, inactiveFG, inactiveBG color.Ansi, symbol string) string {
```
JanDeDobbeleer: We never color directly from a segment but use color overrides instead so the actual ANSI translation (or TMUX specific coloring) happens from the engine.

JanDeDobbeleer (changes requested):
Additionally, I'd rather have one TMUX segement that can do all of the things you want to TMUX.

luisdavim:
This is the type of situation/setup where a daemon mode that @po1o proposed, would help, having a common, shared backend that handles tthe processing of the config andcrendering and some simple clients that just request a prompt could greatly improve eficiency and latency, multiple tmux panes and the status bar, all using the same shared backend. I think most of the latency comes from the process start and reading files, with the daemon, the cache can be in memory.

JanDeDobbeleer:
@luisdavim you always need an executable to talk to the daemon. So that advantage is gone. With the streaming mode this can be achieved as well, provided the shell supports it. And here we'll probably conclude it doesn't (which also ruins deamon mode if none of this runs on an interval).

colings86:
Thanks for the review @JanDeDobbeleer. Will work on the changes over the next couple of days

luisdavim:
> @luisdavim you always need an executable to talk to the daemon. So that advantage is gone. With the streaming mode this can be achieved as well, provided the shell supports it. And here we'll probably conclude it doesn't (which also ruins deamon mode if none of this runs on an interval).

The client can be much simpler, doesn't need to load any configs or read any files, just make a request to get a prompt.

website/package-lock.json
```diff
@@ -232,7 +232,6 @@
       "resolved": "https://registry.npmjs.org/@algolia/client-search/-/client-search-5.46.2.tgz",
       "integrity": "sha512-ZsOJqu4HOG5BlvIFnMU0YKjQ9ZI6r3C31dg2jk5kMWPSdhJpYL9xa5hEe7aieE+707dXeMI4ej3diy6mXdZpgA==",
       "license": "MIT",
```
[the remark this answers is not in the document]
JanDeDobbeleer: This is correct and should be adjusted

src/config/tmux.go
colings86: This holds the config for the left and right parts of the tmux status bar. This is the equivalent of the `claude` block for the claude code status line

colings86 (commented):
@JanDeDobbeleer Sorry, its taken me a few weeks to get back to this. 

There are still some tests to finish adding I think but I wanted to get your thoughts on a couple of the changes here...

1. I implemented the tmux colorise but I want to make sure I understood your intention and the implementation is in line with how you were thinking?
2. I am doubting the tmux segment now as I have found that in the tmux statusline calls to `tmux display-message ...` etc. do not work so I end up using raw tmux formatted templates (e.g. `#S` for session name). This means you can get the same behaviour with the text segment so I am considering removing the dedicated tmux segment and have this PR add tmux statusline support without a dedicated segment. WDYT?

PR #7333, opened 2026-02-21
fix: add missing zorin os icon

adackny:
### Prerequisites

### Description

The OS segment, for Linux systems, retrieves the platform information using this package https://github.com/shirou/gopsutil. In that package, there are some distros that are not recognized,
like Zorin OS or Manjaro, which had a workaround in the **Platform()** function in the **terminal_unix.go** file.
This fix extends that workaround to Zorin OS, and puts that basic check inside a new function **getSpecialLinuxDistros()**.
There is a small update to the unicode escape sequence for the freebsd and mint icons in os.go file.

### DIFF
--- a/src/runtime/terminal_unix.go
+++ b/src/runtime/terminal_unix.go
@@ -122,7 +122,9 @@ func (term *Terminal) getSpecialLinuxDistros(platform string) string {
 	if platform == "arch" && strings.Contains(strings.ToLower(lsbInfo), "manjaro") {
 		// validate for Manjaro
 		return "manjaro"
-	} else if platform == "debian" && strings.Contains(strings.ToLower(lsbInfo), "zorin") {
+	}
+
+	if platform == "debian" && strings.Contains(strings.ToLower(lsbInfo), "zorin") {
 		// validate for Zorin OS
 		return "zorin"
 	}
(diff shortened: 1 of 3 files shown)


src/runtime/terminal_unix.go
```diff
@@ -110,15 +110,23 @@ func (term *Terminal) Platform() string {
...
+	lsbInfo := term.FileContent("/etc/lsb-release")
+
+	if platform == "arch" && strings.Contains(strings.ToLower(lsbInfo), "manjaro") {
 		// validate for Manjaro
-		lsbInfo := term.FileContent("/etc/lsb-release")
-		if strings.Contains(strings.ToLower(lsbInfo), "manjaro") {
-			platform = "manjaro"
-		}
+		return "manjaro"
+	} else if platform == "debian" && strings.Contains(strings.ToLower(lsbInfo), "zorin") {
```
JanDeDobbeleer: don't use an if/else here but an early return instead.

JanDeDobbeleer (changes requested):

src/runtime/terminal_unix.go
adackny: Done

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @adackny for code,doc

This is pure magic 🪄! @holopin-bot @adackny wizard

PR #7360, opened 2026-03-03
Feat/zsh transient multiline

kovjanos:
### Prerequisites

### Description

Previously while multiline command was typed the prompt changed to transient_prompt. If a multiline prompt was used with e.g. a single line transient_prompt, that caused the prompt replaced with transient prompt right after the first enter was hit after "\\" in the multiline command.
With the changes the multiline prompt is kept until the multiline command is submitted.

   1. Code Change: Implemented multi-line command support for Zsh transient prompts in src/shell/scripts/omp.zsh.
   2. Conventional Commits: Verified and committed with the following messages:
       * feat(zsh): support transient prompt for multi-line commands
       * docs: update transient prompt documentation for Zsh multi-line support
   3. Tests: Ran go test ./... in the src directory, and all 23 tests passed. Manual verification confirmed the fix works for both
      multi-line entry and Ctrl-C interrupts.
   4. Documentation: Added a tip to website/docs/configuration/transient.mdx regarding the improved Zsh multi-line support. 

### DIFF
--- a/src/shell/scripts/omp.zsh
+++ b/src/shell/scripts/omp.zsh
@@ -37,8 +37,8 @@ _omp_serve_pid=${_omp_serve_pid:-0}
 _omp_serve_cycle=0
 _omp_serve_failures=0
 
-# set secondary prompt
-_omp_secondary_prompt=$($_omp_executable print secondary --shell=zsh)
+# set secondary prompt (also exports POSH_MULTILINE_KEEPPROMPT)
+eval "$($_omp_executable print secondary --shell=zsh --eval)"
 
 function _omp_set_cursor_position() {
   # not supported in Midnight Commander
@@ -544,13 +544,65 @@ function _omp_restore_rprompt() {
   zle .reset-prompt
 }
 
+# Returns 0 when the current buffer is a complete command, 1 when more input
+# is expected (used to keep the primary prompt while a multi-line command is
+# still being typed).
+function _omp_is_buffer_complete() {
+  local buf=$PREBUFFER$BUFFER
+
+  # 1. Syntax check. The buffer travels over a pipe rather than a here-document,
+  # so no delimiter word can collide with whatever the user happens to type.
+  if ! print -r -- "$buf" | zsh -n 2>/dev/null; then
+    return 1
+  fi
+
+  # 2. An unterminated here-document parses cleanly, because the parser simply
+  # ends the body at EOF. Step 1 therefore cannot see one. Re-parse with `;;`
+  # appended: outside a case branch that is a syntax error, but an open
+  # here-document swallows it as body text. A clean parse means the body was
+  # never closed. The `<<` test only skips the extra parse when no here-document
+  # can possibly be open.
+  if [[ $buf == *'<<'* ]] && print -r -- "$buf"$'\n;;' | zsh -n 2>/dev/null; then
+    return 1
+  fi
+
+  # 3. An odd number of trailing backslashes escapes the newline that follows the
+  # buffer, which parses as a line continuation rather than an incomplete command.
+  local trailing=${buf##*[^\\]}
+  (( ${#trailing} % 2 )) && return 1
+
+  return 0
+}
+
 function _omp_zle-line-init() {
   [[ $CONTEXT == start ]] || return 0
 
   # Start regular line editor.
   (( $+zle_bracketed_paste )) && print -r -n - $zle_bracketed_paste[1]
-  zle .recursive-edit
-  local -i ret=$?
+
+  local -i ret=0
+  if [[ $POSH_MULTILINE_KEEPPROMPT == "true" ]]; then
+    # Keep the full primary prompt while a multi-line command is still being
+    # typed: re-enter the editor until the buffer is a complete command.
+    while true; do
+      zle .recursive-edit
+      ret=$?
+
+      if ((ret)) || _omp_is_buffer_complete; then
+        break
+      fi
+
+      # Nothing may be prefixed to the continuation line. BUFFER is the command
+      # itself, so anything added here is executed. Even pure whitespace breaks a
+      # here-document, whose terminator only matches at the very start of a line.
+      BUFFER+=$'\n'
+      CURSOR=$#BUFFER
+    done
+  else
+    zle .recursive-edit
+    ret=$?
+  fi
+
   (( $+zle_bracketed_paste )) && print -r -n - $zle_bracketed_paste[2]
 
   # We need this workaround because when the `filler` is set,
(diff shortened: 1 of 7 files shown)


src/shell/scripts/omp.zsh
```diff
@@ -24,6 +24,7 @@ _omp_streaming_supported=""
 
 # set secondary prompt
 _omp_secondary_prompt=$($_omp_executable print secondary --shell=zsh)
+_omp_secondary_prompt_plain=$($_omp_executable print secondary --shell=zsh --plain --escape=false)
```
JanDeDobbeleer: why this change?

kovjanos:
@JanDeDobbeleer The _omp_secondary_prompt_plain is used to re-inject prompt content w/o the zsh markers. It unfortunately sinks some control sequences (like coloring) but I didn't find an easy way to re-encode them.
This would of course affect the secondary_prompt behavior so I extended it (actually rearranged/changed the logic) and put it behind a config flag for those who prefer the look of the secondary_prompt over keeping the prompt against transient_prompt for multi-line commands.

src/shell/scripts/omp.zsh
kovjanos: this was needed to avoid ZHS markers that couldn't be re-encoded to mess up the secondary_prompt by directly injecting into the buffer.  Made a change to be able to keep original behavior by default - see new changes.

kovjanos:
fyi: rebased the changes and also providing the details about the behavioral change:

Imagine you have:

- transient_prompt with a single line with time prefix

 - prompt is multiline: one separator line, one info line with e.g. time, user, host, path and the 3rd line is the actual imput prompt
 
 - seondary prompt is only spaces to ident input


 Current behaviour is that if you have some old propts with transient_prompt and the actual input:
```
 [23:48:43]❯ return 0

  23:48:43 - jkovacs@macbook:~/ 
❯ curl \
```

and you hit an enter after "\\", your prompt is immediately replaced with the transient prompt, e.g.:

```
[23:48:43]❯ return 0
[23:51:46]❯ curl \
  -k
```

With the new property set to true, e.g.:

```
[secondary_prompt]
  multiline_keepprompt = true
  template = "  "
```

the behavior will change. e.g. from the same initial state:

```
[23:48:43]❯ return 0

  23:48:43 - jkovacs@macbook:~/ 
❯ curl \
```

if you hit enter, the prompt will not change to transient_prompt:

```
[23:54:57]❯ return 0

  23:54:57 - jkovacs@macbook:~/ 
❯ curl \
  -k \
  http://localhost/nice
```

only if you submit the command:

```
[23:54:57]❯ return 0
[23:57:31]❯ curl \
  -k \
  http://localhost/nice
Hello, World!

  23:57:31 - jkovacs@macbook:~/ 
❯
```

kovjanos:
@JanDeDobbeleer could you please have another look on the change? 
There is also another one - #7369 - on top of this to further enhance prompt behavior.

JanDeDobbeleer:
@kovjanos I've been crazy busy but I'll get to it latest next week!

kovjanos:
FYI: I tried to rebase it for a clean merge, hope it worked

JanDeDobbeleer:
`_omp_is_buffer_complete()` checks the buffer with:

```zsh
zsh -n <<EOF
$PREBUFFER$BUFFER
EOF
```

The heredoc terminator is the literal word `EOF`. If the buffer itself contains a line reading `EOF` (say, the user is typing their own `cat <<EOF ... EOF` heredoc), the outer heredoc ends there, and `zsh -n` only sees the truncated prefix. A genuinely incomplete command, one ending in a trailing `\` for instance, can then be misreported as complete, because the part that would reveal the incompleteness got cut off.

I confirmed the truncation with a bash reproduction of the same POSIX heredoc rule: a buffer ending in `\` (incomplete) was reported "complete" once an embedded `EOF` line truncated it.

Effect: the keep-prompt loop can exit early and hand off to the transient prompt while the user is still mid-command. That's the exact bug this PR fixes, just triggered a different way.

Suggested fix: use a delimiter that can't collide with user input, such as a randomly generated marker, or skip heredocs and pipe the buffer instead: `print -r -- "$PREBUFFER$BUFFER" | zsh -n -`.

JanDeDobbeleer (changes requested):

kovjanos:
Thanks for looking at this!
Back into `_omp_is_buffer_complete` with fresh eyes I came out with two bugs 

**1. The loop was corrupting the command it was building.**

On each continuation line it prefixed the secondary prompt to `BUFFER` — but `BUFFER` *is* the command, so anything added there gets executed. The old code guarded against printable text, assuming whitespace was harmless. It isn't:

```zsh
if true; then
cat <<EOF
hello
EOF
fi
```

The injected indent turns the terminator into `  EOF`, which no longer matches, so the here-document swallows `EOF` and `fi`, the `if` never closes, and the buffer stays permanently "incomplete". The loop never exits and the command can't be submitted at all — Ctrl-C is the only way out.

It also quietly changed data. `echo "hello \` + `hello"` printed `hello   hello` instead of `hello hello`, because the injected spaces landed inside the quoted string.

Fixed by simply not prefixing anything. That also made `_omp_secondary_prompt_plain` dead, so the ANSI/percent-escape stripping on the Go side went away with it.

**2. Here-documents never engaged the feature at all.**

`zsh -n` reports an *unterminated* here-document as complete — it just ends the body at EOF rather than complaining. So `cat <<EOF` looked like a finished command, the loop broke immediately, and the transient prompt took over on the first line. In other words, exactly the bug this PR is meant to fix, still fully present for her
e-documents.

To spot an open here-document I re-parse with `;;` appended. Outside a `case` branch that is a syntax error, but an open here-document swallows it as body text — so a *clean* parse means the body was never closed:

```zsh
if [[ $buf == *'<<'* ]] && print -r -- "$buf"$'\n;;' | zsh -n 2>/dev/null; then
  return 1
fi
```

The `<<` test keeps the extra parse off the common path, so ordinary commands still fork `zsh -n` exactly once. Tested with `<<-EOF` tab-stripping, here-strings (`<<<`), arithmetic shifts (`1 << 2`), a `case` statement's `;;`, `;;` inside a here-document body, and `;;` used as the delimiter itself. All seems OK.

## Summary of the push

- no more injection into `BUFFER` (fixes the hang and the quoted-string corruption)
- open here-documents are now detected, so `multiline_keepprompt` covers them too
- syntax check goes over a pipe instead of a here-document
- docs: `multiline_keepprompt` now states that continuation lines are part of one buffer and get no secondary prompt, so `template` isn't rendered for them

Net effect is a smaller diff than before. Thanks again!

kovjanos:
...and now with the tests

JanDeDobbeleer (approved):

kovjanos:
@JanDeDobbeleer could you please check if something went wrong with this? I can see the approval in the history, and the auto-merge update, but seems like the :next branch was deleted before the auto-merge happened and this got closed, instead of merged.

JanDeDobbeleer:
You're absolutely right, will fix!

JanDeDobbeleer:
Added to #7678

PR #7366, opened 2026-03-05
fix(spotify): use correct D-Bus interface name on Linux

### LINKED ISSUE #7365: Spotify segment broken on Linux: D-Bus calls org.freedesktop.DBus.options instead of org.freedesktop.DBus.Properties
### Code of Conduct

### What happened?

The Spotify segment does not work on Linux. The D-Bus call uses `org.freedesktop.DBus.options.Get` instead of the correct `org.freedesktop.DBus.Properties.Get`, causing the error:

```
Error org.freedesktop.DBus.Error.UnknownMethod: No such interface "org.freedesktop.DBus.options" on object at path /org/mpris/MediaPlayer2
```

It appears the config key name `options` (which replaced the old `properties` key) is leaking into the D-Bus interface name in the shell command that queries Spotify's MPRIS interface.

From the debug trace:
```
terminal.go:RunCommand(bash -c dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.options.Get string:org.mpris.MediaPlayer2.Player string:PlaybackStatus | awk -F '"' '/string/ {print tolower($2)}')
```

The D-Bus interface should always be `org.freedesktop.DBus.Properties.Get` regardless of the config key name.

Manually calling D-Bus with the correct interface works fine:
```
$ dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:org.mpris.MediaPlayer2.Player string:PlaybackStatus
# Returns: variant string "Playing"
```

This appears to be the same issue as #7207 which was closed without resolution.

Linux (Ubuntu, x86_64, KDE)

zsh

### Oh My Posh version

29.7.1

### Theme

```toml
[[blocks]]
type = "rprompt"
alignment = "right"

  [[blocks.segments]]
  type = "spotify"
  style = "plain"
  template = '{{ .Icon }} {{ if ne .Status "stopped" }}{{ .Artist }} - {{ .Track }}{{ end }} '
  foreground = "green"
```

### Log output

```
Spotify(false)     -   3 ms

[DEBUG] 15:04:15.946 segment.go:Execute:173 → segment: Spotify
[DEBUG] 15:04:15.949 terminal.go:RunCommand:280 → Error org.freedesktop.DBus.Error.UnknownMethod: No such interface "org.freedesktop.DBus.options" on object at path /org/mpris/MediaPlayer2
[TRACE] 15:04:15.949 terminal.go:RunCommand(bash -c dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.options.Get string:org.mpris.MediaPlayer2.Player string:PlaybackStatus | awk -F '"' '/string/ {print tolower($2)}') - 3.001363ms
```

aeriondyseti:
## Summary

- The Spotify segment's D-Bus command uses `org.freedesktop.DBus.options.Get` instead of the correct `org.freedesktop.DBus.Properties.Get`, which causes the segment to silently fail on Linux
- This was likely introduced when the config key was renamed from `properties` to `options` — the D-Bus interface name was inadvertently changed along with it
- One-line fix in `spotify_linux.go` and the corresponding test

## Reproduction

With Spotify running on Linux:
```
# This fails (what oh-my-posh currently does):
dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.options.Get string:org.mpris.MediaPlayer2.Player string:PlaybackStatus
# Error: No such interface "org.freedesktop.DBus.options"

# This works (the fix):
dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:org.mpris.MediaPlayer2.Player string:PlaybackStatus
# Returns: variant string "Playing"
```

Fixes #7365

### DIFF
--- a/src/segments/spotify_linux.go
+++ b/src/segments/spotify_linux.go
@@ -40,7 +40,7 @@ func (s *Spotify) Enabled() bool {
 }
 
 func (s *Spotify) runLinuxScriptCommand(command string) string {
-	dbusCMD := "dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.options.Get string:org.mpris.MediaPlayer2.Player"
+	dbusCMD := "dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:org.mpris.MediaPlayer2.Player"
 	val := s.env.RunShellCommand(shell.BASH, dbusCMD+command)
 	return val
 }
--- a/src/segments/spotify_linux_test.go
+++ b/src/segments/spotify_linux_test.go
@@ -31,7 +31,7 @@ func TestSpotifyLinux(t *testing.T) {
 		env := new(mock.Environment)
 		env.On("IsWsl").Return(false)
 
-		dbusCMD := "dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.options.Get string:org.mpris.MediaPlayer2.Player"
+		dbusCMD := "dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:org.mpris.MediaPlayer2.Player"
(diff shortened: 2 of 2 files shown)


JanDeDobbeleer:
@all-contributors please add @aeriondyseti for code

This is pure magic 🪄! @holopin-bot @aeriondyseti wizard

PR #7369, opened 2026-03-06
Feat/transient show interrupt

kovjanos:
### Prerequisites

### Description

This is the 2nd feature on top of the Feat/zsh transient multiline #7360 PR 

This feature adds a new status flag .Interrupted that can be used e.g. in transient_template:

I have a transient command that shows the command and the return code if it is non-zero:

```
[00:06:47]❯ return 1
✘ 1
[00:06:48]❯ return 1
✘ 1
```

The problem is that in quick typing I sometimes interrupt the command with Ctrl-C, which then sometimes can not be differentiated from the submitted but failed commands. To be able to differentiate them, I'm adding a new state which then can be used in the transient_template for this reason, e.g.:

```
[transient_prompt]
  template = "{{ if .Interrupted }}<red>^C </>{{ end }}"
```

With this, you can easily see which commands were interrupted:

```
[00:15:08]❯ return 1
✘ 1
[00:15:10]❯ ^C return 1
✘ 1
```

Note: the reason of the status flag and template usage showing ^C in front of the command instead of having ^C in the position where it was pressed is because sometimes in lazy mode - especially with long multi-line commands - I prefer to be able to copy part of the previously interrupted command, which would be hard if the ^C would be in the middle of the command messing it up

### DIFF
--- a/src/cache/template.go
+++ b/src/cache/template.go
@@ -28,6 +28,7 @@ type SimpleTemplate struct {
 	Code          int
 	WSL           bool
 	Root          bool
+	Interrupted   bool
 }
 
 func (t *Template) AddSegmentData(key string, value any) {
--- a/src/cli/data.go
+++ b/src/cli/data.go
@@ -54,5 +54,9 @@ func applyDataFile(flags *runtime.Flags, changed func(name string) bool) error {
 		flags.PipeStatus = *envFlags.PipeStatus
 	}
 
+	if envFlags.Interrupted != nil && !changed("interrupted") {
+		flags.Interrupted = *envFlags.Interrupted
+	}
+
 	return nil
 }
--- a/src/cli/data_test.go
+++ b/src/cli/data_test.go
@@ -52,7 +52,7 @@ func changedSet(names ...string) func(string) bool {
 func TestApplyDataFile_EmptyPathIsNoop(t *testing.T) {
 	withDataPath(t, "")
 
-	flags := &runtime.Flags{PWD: "/live/pwd", ErrorCode: 1, ExecutionTime: 2.5, PipeStatus: "0"}
+	flags := &runtime.Flags{PWD: "/live/pwd", ErrorCode: 1, ExecutionTime: 2.5, PipeStatus: "0", Interrupted: true}
 
 	err := applyDataFile(flags, noneChanged)
 	require.NoError(t, err)
@@ -61,6 +61,7 @@ func TestApplyDataFile_EmptyPathIsNoop(t *testing.T) {
 	assert.Equal(t, 1, flags.ErrorCode)
 	assert.InDelta(t, 2.5, flags.ExecutionTime, 0)
 	assert.Equal(t, "0", flags.PipeStatus)
+	assert.True(t, flags.Interrupted)
 	assert.Nil(t, flags.EnvData)
 	assert.Nil(t, flags.SegmentData)
 }
@@ -85,9 +86,9 @@ func TestApplyDataFile_UnsupportedExtensionErrors(t *testing.T) {
 	assert.Error(t, err)
 }
 
-func TestApplyDataFile_RoutesAllFourKeysWhenNoFlagsChanged(t *testing.T) {
+func TestApplyDataFile_RoutesEveryKeyWhenNoFlagsChanged(t *testing.T) {
 	path := writeDataFile(t, `{
-		"env": {"PWD": "/data/pwd", "Code": 3, "ExecutionTime": 12.5, "PipeStatus": "0 1"},
+		"env": {"PWD": "/data/pwd", "Code": 3, "ExecutionTime": 12.5, "PipeStatus": "0 1", "Interrupted": true},
 		"segments": {"az": {"Name": "my-sub"}}
 	}`)
 	withDataPath(t, path)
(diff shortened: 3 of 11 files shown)


kovjanos:
FYI: I tried to rebase it for a clean merge, hope it worked

src/config/segment.go
```diff
@@ -77,6 +77,7 @@ type Segment struct {
 	Toggled                bool           `json:"toggled,omitempty" toml:"toggled,omitempty" yaml:"toggled,omitempty"`
 	Pending                bool           `json:"-" toml:"-" yaml:"-"`
 	Interactive            bool           `json:"interactive,omitempty" toml:"interactive,omitempty" yaml:"interactive,omitempty"`
+	MultilineKeepPrompt    bool           `json:"multiline_keepprompt,omitempty" toml:"multiline_keepprompt,omitempty" yaml:"multiline_keepprompt,omitempty"`
```
JanDeDobbeleer: this is related to the other PR, not this one.

kovjanos: @JanDeDobbeleer  I thought I messed up the rebase so tried to rebase again but now I see where it comes from. This pull request is based on top of https://github.com/JanDeDobbeleer/oh-my-posh/pull/7360 , that's why the aggregated changes are showing up. Not sure if github can generate pull request's diff from another pull request or only capable to do this only based on local refs. 
Anyway, the changes related only to this one can be seen at https://github.com/kovjanos/oh-my-posh/commit/f664562fd1fb5d6d874c5c65bca2ef21175c60d4 .

kovjanos: Oh and here it is locally: https://github.com/JanDeDobbeleer/oh-my-posh/pull/7369/changes/f664562fd1fb5d6d874c5c65bca2ef21175c60d4 the 2nd commit's changes from this pull reqest's listed 2 commits

kovjanos:
@JanDeDobbeleer could you please have a look on this - https://github.com/JanDeDobbeleer/oh-my-posh/pull/7369/changes/736ded4d3406b37134d7609b96402a14ecd02ded - and the dependent - https://github.com/JanDeDobbeleer/oh-my-posh/pull/7369/changes/f664562fd1fb5d6d874c5c65bca2ef21175c60d4 - change it is on?  Would these be ok - both say can be cleanly merged - or should I rebase both to current again?

JanDeDobbeleer:
I'll have a look in the course of next week. Sorry, been extremely busy.

JanDeDobbeleer:
Nice addition. The `.Interrupted` flag solves a real annoyance, and the fast-path trade-off on `Ctrl-C` is well explained in the comments.

Two things worth a look before merge:

**1. `Interrupted` is missing from `routedEnvDataKeys` in `src/template/cache.go`.** `loadCache` sets `tmpl.Interrupted` from `env.Flags().Interrupted`, the same way it sets `tmpl.Code` from `env.StatusCodes()`:

```go
routedEnvDataKeys = []string{"PWD", "Code", "ExecutionTime", "PipeStatus"}
```

`Code`, `PWD`, `ExecutionTime`, and `PipeStatus` are protected here precisely because they're CLI-flag-resolved values that a `--data` file must not overwrite. `Interrupted` fits that same category but isn't in the list, so a data file's `"Interrupted"` key would silently clobber the flag-derived value in `overlayEnvData`. For example, replaying this data file:

```json
{
  "Interrupted": true
}
```

with `oh-my-posh print transient --data=that-file.json` (no `--interrupted` flag passed) would set `.Interrupted` to `true` from the file, even though the actual invocation was never interrupted. Adding it to the list closes the gap:

```go
routedEnvDataKeys = []string{"PWD", "Code", "ExecutionTime", "PipeStatus", "Interrupted"}
```

**2. The docs may overpromise cross-shell support.** `transient.mdx` lists `nu`, `fish`, `zsh`, `powershell`, `bash`, and `cmd` as supported shells, and the new `.Interrupted` section carries no shell caveat. Only `omp.zsh` passes `--interrupted`, so users on the other shells will always see `.Interrupted` as `false`, even after a real `Ctrl-C`. Worth a note that this is zsh-only for now (if that's the intent), or a follow-up to wire the rest.

JanDeDobbeleer (changes requested):

kovjanos:
Thanks for the review!

## First the docs part

`--interrupted` is only passed from `omp.zsh`, so on every other shell `.Interrupted` is permanently `false` with nothing to explain why. I've added a caution to the
 `.Interrupted` section:

> `.Interrupted` is only set in `zsh`. Every other shell reports `false`, even after a real `Ctrl-C`.

## Second is the  `routedEnvDataKeys`

You're right that an explicit `--interrupted` shouldn't lose to a data file. On the other hand, adding `Interrupted` to `routedEnvDataKeys` on its own would have traded that for a different problem: `overlayEnvData` strips those keys, and `applyDataFile` doesn't route `Interrupted`... so routing it properly instead, mirroring the other four.
The `*bool` is deliberate: it keeps an explicit `"Interrupted": false` distinguishable from an absent key, so a data file can turn it off as well as on.

Since it wasn't obvious this might be better with:

> A key belongs here if and only if `applyDataFile` routes it - listing one it does not route would instead make that key unsettable from a data file.

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @kovjanos for code,doc

This is pure magic 🪄! @holopin-bot @kovjanos wizard

PR #7397, opened 2026-03-16
feat(segments): add gradle segment

### LINKED ISSUE #7187: Gradle
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

Can you add gradle support.  Something like `mvn`

MO2k4:
Closes #7187

## Summary

- Adds a new `gradle` segment that displays the active Gradle version
- Exposes embedded component versions: Kotlin, Groovy, Ant, and Launcher JVM, parsed from `gradle --version` / `gradlew --version` output
- Automatically uses a `gradlew` wrapper when found in any parent directory

## Test Plan

- [x] `go test ./segments/ -run TestGradle -v` — 4 sub-tests pass (global binary, gradlew wrapper, minimal output, no gradle files)
- [x] `go build ./...` — full binary builds cleanly
- [x] JSON schema validated

### DIFF
--- a/src/segments/gradle.go
+++ b/src/segments/gradle.go
@@ -5,11 +5,11 @@ import (
 )
 
 type Gradle struct {
-	Language
 	KotlinVersion string
 	GroovyVersion string
 	AntVersion    string
 	JVMVersion    string
+	Language
 }
 
 func (g *Gradle) Template() string {
--- a/src/segments/gradle.go
+++ b/src/segments/gradle.go
@@ -1,0 +1,66 @@
+package segments
+
+import (
+	"github.com/jandedobbeleer/oh-my-posh/src/regex"
+)
+
+type Gradle struct {
+	Language
+	KotlinVersion string
+	GroovyVersion string
+	AntVersion    string
+	JVMVersion    string
+}
+
+func (g *Gradle) Template() string {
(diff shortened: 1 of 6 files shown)


src/segments/gradle.go
```diff
@@ -0,0 +1,66 @@
+package segments
+
+import (
+	"github.com/jandedobbeleer/oh-my-posh/src/regex"
+)
+
+type Gradle struct {
+	Language
+	KotlinVersion string
```
JanDeDobbeleer, line 8: Not sure if I'm a fan of the `Version` suffix in all of these. I know why it's there, but we might be better off just naming it `Kotlin` etc for example?

JanDeDobbeleer (changes requested):

src/segments/gradle.go
MO2k4: I used this since other places like Unity/Dotnet also uses the Version suffix, but the others would also work for me.

MO2k4: @JanDeDobbeleer does the thumbsup mean i should adopt into not using the version suffix?

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @MO2k4 for design

Issue #317, opened 2021-01-04
Spotify - Pause / Spotify doesn't work?


bahree:
I don't know what the expected behavior is - when I have music playing, I can see the song as expected, but when I pause or have the music stopped, that segment just disappears. Is that how it is supposed to be? Resuming playing it comes back.

I wonder if this is a setting for Spotify as it is sending out some metadata.

Thanks.

![image](https://user-images.githubusercontent.com/3529468/103584976-2a33d400-4e97-11eb-965b-b211986a73c7.png)
![image](https://user-images.githubusercontent.com/3529468/103584989-2dc75b00-4e97-11eb-8fff-58a907e0f4d9.png)

lnu:
On Windows that's expected. We retrieve the information from the window title and paused/stopped is the same(no title).  
On macOS it's handle differently(AppleScript) and you get the track title when paused.  
On linux it's not yet done, we'll probably do it using dbus(one day maybe).

Issue #153, opened 2020-11-13
Spotify segment doesn't work on Windows


fflaten:
### Prerequisites

- [x] I have read and understand the `CONTRIBUTING` guide
- [x] I looked for duplicate issues before submitting this one

### Description

The spotify-segment doesn't work on Windows, at least with Microsoft Store version. Not only that, but the segment also breaks oh-my-posh3 when Spotify-process is not running.

### Environment

- Oh my Posh version: 3.34.0 beta
- Theme: jandedobbeleer (and custom theme)
- Operating System: Windows 10 20H2
- Shell: PowerShell 7.1
- Terminal: Windows Temrinal 1.4.3141.0

### Steps to Reproduce

1. Install Spotify using Microsoft Store (haven't tested standalone version)
2. Open PowerShell and set theme to jandedobbeleer
3. Verify spotify icon shows stopped-icon
4. Play song in spotify - should still show stopped-icon
5. Close/exit spotify and write ex. cls
6. Oh-my-posh prompt is gone. Only showing `PS>`

**Expected behavior:**
Song playing: Spotify segment should show song
Song stoppped: Spotify segment should show stopped icon
Spotify not running: Spotify segment should not be visible

**Actual behavior:**
Song playing: Spotify segment shows stopped icon
Song stoppped: Spotify segment shows stopped icon
Spotify not running: Oh-my-posh prompt missing

fflaten:
The broken prompt when spotify isn't running might be same as issue mentioned PR #152, but it doesn't mention song-detection bugfix. MS Store-version of Spotify also shows song in title, so the concept should work

JanDeDobbeleer:
@fflaten merged the bugfix, can you check when that release drops? Should be pretty fast.

fflaten:
Great. It fixed the crash (missing prompt) and song is detected. 👏

Would still like the Spotify-segment to be invisible when the process is not running. Currently shows stopped icon. New issue?

JanDeDobbeleer:
That's [by design](https://github.com/JanDeDobbeleer/oh-my-posh3/blob/main/segment_spotify_windows.go#L17) on Windows if I'm not mistaken.

What do you expect to see? I didn't write this myself, credit goes to @lnu

fflaten:
I'd expect it to be hidden when not applicable (no process running), just like git, python virtenv (iirc) etc.

JanDeDobbeleer:
[...]

PR #7496, opened 2026-04-29
feat(spotify): switch Windows and WSL to SMTC-based detection

JamBalaya56562:
### Prerequisites

### Description

The Windows / WSL Spotify segment used to scrape the `spotify.exe` window title, which can only ever surface the `playing` state — `paused` and `stopped` look the same as "no song", and the Microsoft Store version suffers from a few additional rough edges. This PR replaces that path with a query against Windows [System Media Transport Controls][smtc] (SMTC) so Windows / WSL reach feature parity with macOS / Linux, and adds optional advertisement detection for free-tier users.

#### What changes

- **Paused / stopped detection on Windows and WSL.** Same icons, same template fields as macOS / Linux — no segment config change required.
- **Optional `ad` state on Windows and WSL** based on the observed Spotify ad pattern (`AlbumTitle == "" && TrackNumber == 0`). Users can pick the icon via the new `ad_icon` option (default ` `, matching the YouTube Music segment).
- **Microsoft Store Spotify is detected** alongside the native Win32 client (the SMTC `SourceAppUserModelId` covers both). This implicitly resolves the long-standing complaints in #153.
- **Edge PWA fallback is preserved** — when SMTC has no Spotify session, the existing `msedge.exe` window-title scraper still kicks in.

#### How it works

A new shared file `src/segments/spotify_smtc.go` (`//go:build windows || (linux && !darwin)`) holds:

- A small PowerShell script that talks to `Windows.Media.Control.GlobalSystemMediaTransportControlsSessionManager` and prints `<status>|<title>|<artist>|<album>|<trackNumber>` (UTF-8).
- `querySMTC()` — runs the script via `env.RunCommand("powershell.exe", …)`.
- `parseSMTCOutput()` — fills `s.Status / Track / Artist`, applies the ad heuristic, and calls `resolveIcon()`.

`spotify_windows.go` is now a thin wrapper that calls `querySMTC()` and falls through to the Edge PWA scraper. `spotify_linux.go`'s `enabledWsl()` delegates to the same `querySMTC()` — WSL invokes the host's `powershell.exe` through interop, so the Linux side reads the exact same SMTC sessions as Windows.

#### Trade-offs

- The PowerShell startup cost adds roughly 200–400 ms to the prompt, comparable to the macOS `osascript` path. A future improvement would be calling WinRT directly (e.g. via `saltosystems/winrt-go`); I went for the minimal change first since it mirrors the existing WSL approach. Happy to follow up if preferred.
- Linux / macOS `ad` detection is intentionally out of scope here. The `Album=="" && TrackNumber==0` heuristic was observed via SMTC specifically; the equivalent MPRIS / AppleScript signals still need on-platform observation. The shared `ad` constant and `AdIcon` resolution land in `spotify.go` so the cross-platform follow-up is a localized change.

#### Tests

- `TestSpotifyWindowsSMTC` — 9 cases (playing / paused / **ad** / stopped / closed / Unicode-quoted track / PowerShell error / two malformed-output shapes).
- `TestSpotifyWindowsPWA` — updated to mock the SMTC fallback path.
- `TestSpotifyWSL` — rewritten with 7 SMTC cases mirroring the Windows ones.
- All `Spotify` tests pass on Windows GOOS and on a Linux test binary (`GOOS=linux go test -c`) executed inside WSL Ubuntu.

#### Manual verification

End-to-end SMTC output observed on the contributor's machine, including an actual Spotify ad caught in the wild:

```text
playing|<song title>|<artist>|<album>|<track number>   ← regular tracks
playing|君のこころが観たいもの　プライムビデオ|プライムビデオ||0   ← ad
```

Refs #317 (Spotify pause not detected on Windows), #153 (MS Store Spotify not detected on Windows).

### DIFF
--- a/src/runtime/environment.go
+++ b/src/runtime/environment.go
@@ -51,6 +51,7 @@ type Environment interface {
 	Flags() *Flags
 	BatteryState() (*battery.Info, error)
 	QueryWindowTitles(processName, windowTitleRegex string) (string, error)
+	QuerySpotifySMTC() (string, error)
 	WindowsRegistryKeyValue(key string) (*WindowsRegistryValue, error)
 	HTTPRequest(url string, body io.Reader, timeout int, requestModifiers ...http.RequestModifier) ([]byte, error)
 	IsWsl() bool
--- a/src/runtime/environment.go
+++ b/src/runtime/environment.go
@@ -22,6 +22,18 @@ const (
 	PRIMARY = "primary"
 )
 
+// MediaInfo holds a single media session read from the OS media-transport
+// layer (e.g. Windows System Media Transport Controls). It is player-agnostic:
+// the SMTC mechanism can surface any app that publishes a session.
+type MediaInfo struct {
+	// Status is the lowercased playback status: playing/paused/stopped/closed/opened/changing.
+	Status      string
+	Title       string
+	Artist      string
+	Album       string
+	TrackNumber int
+}
+
 type Environment interface {
 	Getenv(key string) string
 	Pwd() string
@@ -51,7 +63,7 @@ type Environment interface {
 	Flags() *Flags
 	BatteryState() (*battery.Info, error)
 	QueryWindowTitles(processName, windowTitleRegex string) (string, error)
-	QuerySpotifySMTC() (string, error)
+	QueryMediaPlayer(player string) (*MediaInfo, error)
 	WindowsRegistryKeyValue(key string) (*WindowsRegistryValue, error)
 	HTTPRequest(url string, body io.Reader, timeout int, requestModifiers ...http.RequestModifier) ([]byte, error)
 	IsWsl() bool
--- a/src/runtime/smtc_windows.go
+++ b/src/runtime/smtc_windows.go
@@ -1,0 +1,380 @@
+//go:build windows
+
+package runtime
+
+import (
+	"errors"
+	"fmt"
+	goruntime "runtime"
+	"strings"
+	"syscall"
+	"time"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+// querySpotifySMTC reads the active Spotify session from the Windows System
+// Media Transport Controls (SMTC) using a minimal in-tree WinRT binding —
+// `combase.dll` exports + COM vtable dispatch. It returns the line
+//
+//	"<status>|<title>|<artist>|<album>|<trackNumber>"
+//
+// which the segment-side parseSMTCOutput already understands, including the
+// `playing && album=="" && trackNumber=="0"` ad-detection heuristic.
+//
+// Vtable indices and IIDs were verified by .NET reflection against
+// Windows.Media.Control.winmd on Windows 11 (see the table in the const
+// blocks below). All WinRT interfaces inherit from IInspectable, which sits
+// on top of IUnknown:
+//
+//	IUnknown      [0] QueryInterface  [1] AddRef  [2] Release
+//	IInspectable  [3] GetIids  [4] GetRuntimeClassName  [5] GetTrustLevel
+//
+// so interface-specific methods always start at slot 6.
+
+// COM vtable layout. The actual object in memory starts with a pointer to
+// its vtable; the vtable starts with N function pointers. Modelling both as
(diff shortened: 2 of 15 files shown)


JanDeDobbeleer:
@JamBalaya56562 I don't want to call pwsh from the Windows binary as that's too slow. I asked Copilot to see if they can figure out a native winRT binding from golang instead.

JamBalaya56562:
> I don't want to call pwsh from the Windows binary as that's too slow. I asked Copilot to see if they can figure out a native winRT binding from golang instead.

Thanks — agreed, pwsh-per-render was unsustainable. Pushed a native WinRT implementation.

What's there: an in-tree binding in src/runtime/smtc_windows.go that calls `combase.dll` directly (`RoInitialize` / `RoGetActivationFactory` / HSTRING) and dispatches via COM vtables. Zero third-party deps. New `Environment.QuerySpotifySMTC()` is the seam, so the segment-side parser (and the `playing && album=="" && trackNumber=="0"` ad heuristic) is unchanged — only the transport flipped. `goruntime.LockOSThread()` keeps the `RoInitialize`/`RoUninitialize` pair on one OS thread.

WSL is still pwsh. WinRT isn't reachable from a Linux ELF, so spotify_smtc.go is now linux-only and keeps the script. If a future PR ships a small Windows-side bridge (named pipe / out-of-proc helper), WSL can switch too — out of scope here.

Verification: vtable indices and IIDs were derived by .NET reflection over `Windows.Media.Control.winmd` on Win11 (e.g. `IGlobalSystemMediaTransportControlsSessionManager` IID `2050c4ee-11a0-57de-aed7-c97c70338245`, `get_PlaybackStatus` at vtable slot 7, `get_Artist` at slot 9, `get_AlbumTitle` at slot 10, `get_TrackNumber` at slot 11). End-to-end test against a live Spotify session round-trips the full 5-field tuple including UTF-16 → UTF-8 across non-ASCII titles.

Saw your Copilot exploration at #7508 — the architecture (Environment seam, `comObj`/`comVTable` go-vet trick) was helpful as a sanity check, though the WinMD-derived indices and the manager IID needed correcting and ad detection had been dropped, so I didn't pull it in directly. Happy to close #7508 once this lands.

src/runtime/smtc_windows.go
```diff
@@ -0,0 +1,380 @@
...
+	if hr := comCall(playbackInfo, playbackInfoGetPlaybackStatus, uintptr(unsafe.Pointer(&status))); hr != 0 {
+		return "", false
+	}
+
+	statusStr := smtcStatusString(status)
+
+	// Title/artist/album/trackNumber are best-effort; if the async op fails
+	// we still return the status so the segment can hide cleanly.
+	title, artist, album, trackNumber := readMediaProperties(session)
+	return fmt.Sprintf("%s|%s|%s|%s|%d", statusStr, title, artist, album, trackNumber), true
```
JanDeDobbeleer: this needs to return an pbject using a pointer instead of a string, this approach isn't the most optimal one.

src/runtime/environment.go
```diff
@@ -51,6 +51,7 @@ type Environment interface {
 	Flags() *Flags
 	BatteryState() (*battery.Info, error)
 	QueryWindowTitles(processName, windowTitleRegex string) (string, error)
+	QuerySpotifySMTC() (string, error)
```
JanDeDobbeleer: is the SPotify specific or can we query ANY media steam?

JanDeDobbeleer (changes requested):

src/runtime/smtc_windows.go
JamBalaya56562: Done — it now returns a `*runtime.MediaInfo` struct instead of the packed string, so no more pack/parse round-trip. Status mapping and the ad heuristic moved into a shared `applyMediaInfo`.

src/runtime/environment.go
JamBalaya56562: Good call — SMTC is player-agnostic, so I generalized the seam to `QueryMediaPlayer(player string)`, matching the session by a case-insensitive `SourceAppUserModelId` substring. The Spotify segment just passes `"spotify"`; any other player works too.

JanDeDobbeleer (approved):

PR #7499, opened 2026-04-30
feat(aws): expose all shared config settings via .Settings map

### LINKED ISSUE #6456: AWS segment additional Properties
### Code of Conduct

Would be nice to get properties with
- sso_account_id
- aws_access_key_id

Might be useful to have bool properties like `display_account_id` and `display_access_key_id`. I feel that the first bool property might have default true, the second false.

JamBalaya56562:
### Prerequisites

### Description

Closes #6456.

The AWS segment used to surface only `.Profile` and `.Region`. This PR
extends it so any setting that AWS SDKs and the AWS CLI recognize in
the shared config and credentials files can be used from a template,
without bolting on a new struct field for every key AWS adds in the
future.

#### What's new

- **`.AccountID`** — populated from `sso_account_id` (preferred) or
  `aws_account_id` of the active profile.
- **`.AccessKeyID`** — populated from `AWS_ACCESS_KEY_ID`,
  `~/.aws/credentials`, or `~/.aws/config` (in that order, matching SDK
  precedence).
- **`.Settings`** (`map[string]string`) — every key/value pair from the
  active profile in both files. Templates access individual keys via
  `{{ index .Settings "<key>" }}` or `{{ .Settings.<key> }}`.
- **`.SSOSession`** (`map[string]string`) — when the active profile has
  `sso_session = <name>`, the matching `[sso-session <name>]` section
  is resolved into this map. Useful for the new IAM Identity Center
  config layout (`sso_start_url`, `sso_region`,
  `sso_registration_scopes`, …).

#### New options

| Option                  | Default | Description                                                            |
| ----------------------- | :-----: | ---------------------------------------------------------------------- |
| `display_account_id`    | `true`  | populate `.AccountID` from the active profile                          |
| `display_access_key_id` | `false` | populate `.AccessKeyID` (off by default since access keys are secrets) |

#### Examples

```template
{{ .Profile }}@{{ .Region }} ({{ .AccountID }}/{{ index .Settings "sso_role_name" }})
```

```template
{{ .Profile }}{{ with index .Settings "role_arn" }} → {{ . }}{{ end }}
```

```template
{{ .Profile }} via {{ index .SSOSession "sso_start_url" }}
```

#### Implementation notes

- INI parsing was switched to `gopkg.in/ini.v1` (already a dependency
  via the `gcp` segment), which correctly handles section boundaries.
  The previous string-prefix scan worked only because it bailed out
  after the first matching key.
- File precedence mirrors the AWS SDK: config file is read first, then
  the credentials file overrides credential keys for the same profile.
  Section names follow the documented format —
  `[default]` / `[profile <name>]` for the config file,
  `[default]` / `[<name>]` for the credentials file.
- `AWS_CONFIG_FILE` and `AWS_SHARED_CREDENTIALS_FILE` are honored.
- Backward compatible: `.Profile`, `.Region`, and `.RegionAlias` keep
  their existing behavior; all pre-existing test cases pass unchanged.

#### Verification

- `go build ./segments/`
- `gofmt -l ./segments/aws.go ./segments/aws_test.go` — clean
- `go vet ./segments/`
- `golangci-lint run ./segments/...` — 0 issues
- `go test ./...` — all packages pass, including new test cases for
  `.AccountID`, `.AccessKeyID` (env / credentials / config fallback),
  `.Settings` (`role_arn`, `mfa_serial`, `output`), `.SSOSession`
  resolution, and credentials-file-overrides-config behavior.

#### Files changed

- `src/segments/aws.go`
- `src/segments/aws_test.go`
- `website/docs/segments/cloud/aws.mdx` — adds the new options /
  properties and a categorized reference list of every setting AWS
  recognizes (Credentials, IAM Identity Center, Assume role, Region &
  endpoints, S3, EC2/IMDS, Retry, Auth, General, CLI-only).
- `themes/schema.json` — adds the two new options under the AWS
  segment.

### DIFF
--- a/src/segments/aws.go
+++ b/src/segments/aws.go
@@ -4,19 +4,54 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/jandedobbeleer/oh-my-posh/src/log"
 	"github.com/jandedobbeleer/oh-my-posh/src/regex"
 	"github.com/jandedobbeleer/oh-my-posh/src/segments/options"
+
+	"gopkg.in/ini.v1"
 )
 
 type Aws struct {
 	Base
 
-	Profile string
-	Region  string
+	// Settings holds every key/value pair from the active profile in the AWS shared
+	// config and credentials files. Credential-file entries take precedence over
+	// config-file entries for the same key, mirroring the AWS SDK's resolution order.
+	// Templates can read any AWS-recognized setting via {{ .Settings.<key> }}, e.g.
+	// {{ .Settings.role_arn }} or {{ .Settings.sso_role_name }}.
+	Settings map[string]string
+
+	// SSOSession holds the resolved [sso-session <name>] section keys when the
+	// active profile references one via the sso_session key. Use as
+	// {{ .SSOSession.sso_start_url }}, etc.
+	SSOSession map[string]string
+
+	Profile     string
+	Region      string
+	AccountID   string
+	AccessKeyID string
 }
 
 const (
 	defaultUser = "default"
+
+	// DisplayAccountID toggles populating the AccountID convenience field
+	// from sso_account_id (preferred) or aws_account_id.
+	DisplayAccountID options.Option = "display_account_id"
+	// DisplayAccessKeyID toggles populating the AccessKeyID convenience field
+	// from AWS_ACCESS_KEY_ID, the credentials file, or the config file.
+	DisplayAccessKeyID options.Option = "display_access_key_id"
+
+	// AWS shared config keys we promote to convenience fields.
+	awsKeyRegion       = "region"
+	awsKeyAccessKeyID  = "aws_access_key_id"
+	awsKeyAccountID    = "aws_account_id"
+	awsKeySSOAccountID = "sso_account_id"
+	awsKeySSOSession   = "sso_session"
+
+	awsConfigSectionDefault    = "default"
+	awsConfigSectionPrefix     = "profile "
+	awsConfigSectionSSOSession = "sso-session "
 )
 
 func (a *Aws) Template() string {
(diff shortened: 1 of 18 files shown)


src/segments/aws.go
```diff
@@ -36,63 +73,123 @@ func (a *Aws) Enabled() bool {
...
+		a.copySection(cfg, awsConfigSectionSSOSession+sessionName, a.SSOSession)
 	}
+}
 
-	if a.Profile == "" && len(a.Region) != 0 {
-		a.Profile = defaultUser
+func (a *Aws) loadCredentialsFile() {
+	credentialsPath := a.env.Getenv("AWS_SHARED_CREDENTIALS_FILE")
+	if credentialsPath == "" {
+		credentialsPath = fmt.Sprintf("%s/.aws/credentials", a.env.Home())
```
JanDeDobbeleer: filepath.Join()

src/segments/aws.go
```diff
@@ -36,63 +73,123 @@ func (a *Aws) Enabled() bool {
 	}
 
 	displayDefaultUser := a.options.Bool(options.DisplayDefault, true)
+	displayAccountID := a.options.Bool(DisplayAccountID, true)
```
JanDeDobbeleer: I wouldn't switch this as it doesn't impact any performance, we still parse anyways. Unless if it reduces the lifting we do, no need to add a switch.

JanDeDobbeleer (changes requested):

src/segments/aws.go
JamBalaya56562: Good point — dropped the switch, `.AccountID` is now always populated.

JanDeDobbeleer (approved):

PR #7577, opened 2026-06-11
fix(transient): cancel transient newlines as with primary newlines

rightaditya:
Currently, if a prompt has `newline` set to `true`, the newline will be omitted in several sensible cases. However this behaviour isn't followed if `transient_prompt` has `newline` set to `true`, meaning the positioning of the transient prompt relative to the primary one differs depending on whether the cancellation criteria are met and the primary prompt has had its newline removed.

This PR makes sure that both prompt types use the same cancellation logic.

### Prerequisites

### Description

### DIFF
--- a/src/prompt/engine.go
+++ b/src/prompt/engine.go
@@ -541,6 +541,11 @@ func (e *Engine) rectifyTerminalWidth(diff int) {
 	e.Env.Flags().TerminalWidth += diff
 }
 
+func (e *Engine) cancelNewline() bool {
+	row, _ := e.Env.CursorPosition()
+	return e.Env.Flags().Cleared || e.Env.Flags().PromptCount == 1 || row == 1
+}
+
 // New returns a prompt engine initialized with the
 // given configuration options, and is ready to print any
 // of the prompt components.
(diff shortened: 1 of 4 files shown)


rightaditya:
@JanDeDobbeleer Assuming the closing wasn't by accident, would you be willing to accept a version of this where this is a toggle that the user can change? That way no current behaviour would be affected. The reason I want to add this is the inconsistency that we can see in the first prompt of a new terminal compared to subsequent ones.

New terminal, primary prompt:
Screenshot From 2026-06-20 00-22-54

Then run `pwd` and the above prompt is replaced with the secondary (transient) prompt:
Screenshot From 2026-06-20 00-23-03

Then run `ls` and the new primary prompt is similarly replaced:
Screenshot From 2026-06-20 00-23-08

IMO the latter behaviour is correct. But the transient prompt doesn't do the newline deletion in the first case but it's missing the set of extra checks that the primary prompt has to account for this case (and similar).

rightaditya:
I can also just open a new issue if you'd rather keep discussion there.

JanDeDobbeleer:
This was by accident. I'll open it after my house work is done. Or, if you could recreate this PR we'll get it sorted. I'm trying to get more stuff in over the coming weeks.

rightaditya:
NP, thanks for maintaining :) I'll rebase and submit a new one.

PR #7614, opened 2026-06-27
fix(config): recover theme from cache when session config is lost

acdtrx:
## What

When the per-session config cache is lost mid-session, the prompt silently falls back to the **default built-in theme** for the rest of that shell. This recovers the configured theme (and self-heals the cache) instead.

## Why / root cause

A prompt render calls `oh-my-posh print` **without** `--config`, so config resolution depends entirely on the session cache (`zsh.$POSH_SESSION_ID.omp.cache` → `CONFIG`/`CONFIG_SOURCE`).

- Only `init` ever writes those keys; `print --save-cache` only *preserves* them.
- If that session cache file is lost mid-session — purged under disk pressure (e.g. macOS `com.apple.cache_delete`), or **corrupted by a concurrent writer that inherited the same `$POSH_SESSION_ID`** (the gob file is read‑modify‑written without locking) — the next `print` rebuilds the file without the config.
- `config.Get("", false)` then hits the cache miss and calls `Load("")` → `Parse("")` → `ErrNoConfig` → `Default()`, i.e. the default theme. It stays default until a new shell re-runs `init`.

This presents as a random "my theme reverted mid-session" that a new terminal fixes. It's easy to hit when a second process shares the session id (multiplexers, subshells, agent tooling, etc.).

### Reproduction
```sh
eval "$(oh-my-posh init zsh --config ~/path/theme.omp.json)"   # theme renders
rm ~/.cache/oh-my-posh/zsh.$POSH_SESSION_ID.omp.cache          # purge / simulate a corrupt concurrent write
# press Enter -> prompt is now the DEFAULT theme for the rest of the session
```
`oh-my-posh debug` in that state shows `no --config set, using default built-in configuration`, `(session) key not found: CONFIG`, `no cached config found`.

## How

In `config.Get`, when no explicit `--config` was provided and the session cache is missing/corrupt, recover the source before falling back to the default:

1. session `CONFIG_SOURCE` — still present when only the config *blob* is corrupt;
2. device-level **DSC** (`Resource[*Configuration]`) — persists across sessions with an infinite TTL and is populated on every `init`. Since `init` *requires* `--config` (`MarkPersistentFlagRequired("config")`), the DSC always reflects a real, user-chosen theme, so this introduces no "resurrect an unwanted theme" path. The most recently added state is used.

On recovery the config is re-parsed and `Store`d again, so the session cache self-heals and subsequent renders stay on the fast path. Behaviour is unchanged when `--config` is supplied or when the cache is valid.

## Tests

Added `src/config/gob_test.go`:
- recovery from the device DSC when the session cache is gone (+ asserts the session cache is repopulated);
- recovery from the session source when only the config blob is missing;
- fallback to the default when there is genuinely nothing to recover.

Verified end-to-end against a release build: before, `debug` reports the default theme after the cache is removed; after, it reports `recovered configuration from source: …` and renders the configured theme.

### DIFF
--- a/src/config/gob.go
+++ b/src/config/gob.go
@@ -34,23 +34,66 @@ func Get(configFile string, reload bool) *Config {
 		}
 	}
 
-	base64String, found := cache.Get[string](cache.Session, configKey)
-	if !found {
-		log.Debug("no cached config found")
-		cfg := Load(configFile)
-		cfg.Store()
-		return cfg
-	}
+	if base64String, found := cache.Get[string](cache.Session, configKey); found {
+		var cfg Config
+		if err := cfg.Restore(base64String); err == nil {
+			return &cfg
+		}
 
-	var cfg Config
-	if err := cfg.Restore(base64String); err != nil {
 		log.Debug("failed to restore config from cache")
-		loaded := Load(configFile)
-		loaded.Store()
-		return loaded
+	} else {
+		log.Debug("no cached config found")
 	}
 
-	return &cfg
+	// A prompt render relies on the session cache for its configuration: it's
+	// invoked without an explicit --config, so configFile is empty here. When the
+	// session cache is missing or corrupt - the file got purged (e.g. under disk
+	// pressure) or clobbered by a concurrent writer sharing the session - falling
+	// back to Load("") silently renders the default theme for the rest of the
+	// session. Recover the previously initialized source first so the configured
+	// theme keeps rendering and the session cache self-heals.
+	if configFile == "" {
+		if source := recoverSource(); source != "" {
+			if cfg, err := Parse(source); err == nil {
+				log.Debug("recovered configuration from source: ", source)
+				cfg.Store()
+				return cfg
+			}
+
+			log.Debug("failed to recover configuration from source: ", source)
+		}
+	}
+
+	cfg := Load(configFile)
+	cfg.Store()
+	return cfg
+}
+
+// recoverSource locates the most recently initialized configuration source when
+// the session config cache is unavailable. It first checks the session source,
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer:
Thanks for the analysis, it's spot on: a prompt render depends entirely on the session cache, and losing that cache mid-session silently reverts the theme to the default. I kept the diagnosis, the restructured `Get`, and the test setup, but reworked the recovery itself to use `POSH_CONFIG` instead of the DSC.

The DSC is the wrong source to recover from. It records every configuration ever parsed, deduplicated by source, so its last entry is not the most recent init: `extends` parents land after their child, one-off `--config` invocations are recorded too, and a device with different themes per shell recovers whichever happens to be last. It also skips remote configurations entirely, which this PR's main scenario (a purged cache file) needs.

The shell session itself is the reliable source. The root command already resolves `POSH_CONFIG` into `--config` when no flag is given, so the init script now pins `POSH_CONFIG` to the session's resolved configuration source. Seven shells export it from the init command's output; Nushell gets it baked into its autoload script since it has no eval step. When the session cache is lost, the render recovers the exact theme the session was initialized with, remote configurations included, and stores it again so the cache heals itself.

Two details to be aware of:

- A healthy session cache always wins, so `POSH_CONFIG` cannot change the configuration mid-session.
- When the recovery source is temporarily unreadable (a failing remote config fetch, for example), we return the default without caching it. The next render retries the recovery instead of pinning the default for the rest of the session.

JanDeDobbeleer:
@all-contributors please add @acdtrx for code

This is pure magic 🪄! @holopin-bot @acdtrx wizard

PR #7220, opened 2026-01-19
fix: `Failed to write init script` when multiple instances of OMP are started simultaneously

BPplays:
### Prerequisites

- [x] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [x] The commit message follows the [conventional commits][cc] guidelines.
- [ ] Tests for the changes have been added (for bug fixes / features).
- [ ] Docs have been added/updated (for bug fixes / features).

### Description
i was getting an error when i was using oh-my-posh with a terminal that restores all the tabs on startup:
```
Failed to write init script: open C:\Users\irlname\AppData\Local\oh-my-posh\init.16695873114097618080.ps1: The process cannot access the file because it is being used by another process.
```

i implemented a function that only writes a file if it's different from the one on disk already and made the `writeScript` function retry until a deadline is reached

[CONTRIBUTING.md]: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/CONTRIBUTING.md
[cc]: https://www.conventionalcommits.org/en/v1.0.0/#summary

PR #7640, opened 2026-07-07
fix (resubmitted): `Failed to write init script` when multiple instances of OMP are started simultaneously

BPplays:
### Prerequisites

### Description
i'm resubmitting #7220 because searching the commit history, and looking at the code in `main` it doesn't seem that any fix was ever made

### Original description
i was getting an error when i was using oh-my-posh with a terminal that restores all the tabs on startup:
```
Failed to write init script: open C:\Users\irlname\AppData\Local\oh-my-posh\init.16695873114097618080.ps1: The process cannot access the file because it is being used by another process.
```

i implemented a function that only writes a file if it's different from the one on disk already and made the `writeScript` function retry until a deadline is reached

### DIFF
--- a/src/shell/filesystem.go
+++ b/src/shell/filesystem.go
@@ -1,11 +1,13 @@
 package shell
 
 import (
+	"bytes"
 	"fmt"
 	"hash/fnv"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/build"
 	"github.com/jandedobbeleer/oh-my-posh/src/cache"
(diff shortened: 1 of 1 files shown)


JanDeDobbeleer:
Thanks for resubmitting this. The approach is right: skip the write when the script is unchanged, and retry when the file is locked. I reworked the implementation before merging, for three reasons.

The permission check never matches on Windows. Go's FileInfo.Mode().Perm() reports only 0666 (writable) or 0444 (read-only) on Windows, never 0644. So filesEqual always returned false on the exact platform this fix targets, and every instance still wrote the file on every startup. I dropped the check; we always create the file with 0o644, so content equality is what matters.

The write is now atomic. os.WriteFile truncates the target before writing, so a shell sourcing the script mid-write could read a partial file. The new code writes to a temp file in the same directory and renames it over the target. Concurrent instances no longer collide on the write itself, only on the rename, and the retry covers that.

The retry loop is simpler and shorter. A fixed 8 attempts with backoff doubling from 50ms (capped at 1s, about 3.5s total) replaces the 10-second deadline and the firstErr bookkeeping. This code runs during shell startup, and blocking a new tab for 10 seconds is worse than failing. The loop returns the last error, the one that persisted through all retries, and logs it once instead of on every attempt.

PR #7665, opened 2026-07-12
feat(transient): support right-aligned template

### LINKED ISSUE #4822: Allow transient prompt alignment (segments or left/right_template or whatever...)
### Code of Conduct

Related: [3844](https://github.com/JanDeDobbeleer/oh-my-posh/issues/3844)

Mostly care about powershell, but would like to use for zsh.
Pretty much a dup of referenced issue, but I'll give my use cases:

1. I have a multi-line prompt and I want to reduce it to 1 line with the more/most important info via transient mechanics.
2. Filler, which sort of works, should automatically be applied after the command input instead of before.

As opposed to going all out and supporting multiple segments, I would be thrilled just to have:
`left_template` -> alias  for 'template'
`right_template` -> right aligned (after optional filler) template.

Image attached of my powerlevel10k zsh prompt usage that I'd like to replace.  It looks noisy especially because I used short-output commands with a somewhat narrow screen to demo.

image

JamBalaya56562:
## What

Adds a `right_template` option to the transient prompt, rendered right-aligned at the end of the line:

```json
"transient_prompt": {
  "template": "❯ ",
  "right_template": "{{ now | date \"15:04:05\" }}",
  "filler": "."
}
```

resolves #4822

## Demo

Three accepted commands collapsed to a single transient line each, followed by the multi-line primary prompt. Same theme in both shots — the only difference is `right_template`.

**Before** — the transient prompt has no right side; a `filler` just runs to the edge of the terminal:

transient prompt without a right-aligned template

**After** — `right_template` renders the exit status and time at the right edge, and the `filler` now bridges the gap:

transient prompt with a right-aligned template

<sub>Generated with the built-in image exporter (<code>image.Renderer</code>) from real <code>print transient</code> output, the same way the theme previews on ohmyposh.dev are produced.</sub>

<details><summary>demo theme</summary>

```json
"transient_prompt": {
  "foreground": "#A6E3A1",
  "template": "❯ ",
  "right_template": "{{ if gt .Code 0 }}<#F38BA8> {{ .Code }}</>{{ else }}<#A6E3A1></>{{ end }} <#7F849C>{{ now | date \"15:04:05\" }}</>",
  "filler": "<#45475A>·</>"
}
```

</details>

## How

Kept deliberately minimal — a single new property, no changes to any shell script:

- **zsh**: the eval output now sets `RPROMPT=<rendered right template>` instead of the hardcoded `RPROMPT=''`, letting zsh align it natively (same pattern the primary prompt already uses). Without a `right_template` the output is byte-identical to before.
- **pwsh**: the right side is embedded in the returned prompt string via cursor save/restore after `ClearAfter()`, mirroring `writePrimaryRightPrompt()`. The cursor is restored to right after the left part, so the accepted command is drawn over the gap — same graceful degradation as the primary right prompt.
- **filler**: when configured together with `right_template`, the filler now fills the gap between both parts.
- **streaming (zsh)**: a streamed transient record can only carry PS1 (the script resets `RPROMPT=''`, see `_omp_zle-line-init`), so the record is skipped when a `right_template` is configured — the script then falls back to the eval path and gets both values. pwsh needs no special handling since everything is embedded in the returned string. `serve`'s Clink wait protocol is untouched (cmd is out of scope).
- **`shouldFill`**: guarded against a negative padding length, which would panic in `strings.Repeat` (the right side subtraction makes this reachable; a too-wide left prompt could already trigger it).

### Scope

`zsh` and `powershell` only, as both work through the existing CLI surface. `fish`/`cmd`/`nu` would each need a new way to fetch the right side separately (and for cmd a serve protocol change), so they are intentionally left out — happy to follow up if there's demand. This covers the use case from #4822 (pwsh + zsh). The docs note the supported shells.

## Verification

- New unit tests in `src/prompt/extra_test.go` covering zsh (±eval, ±filler, quoting, unknown width), pwsh (±filler, exact fit, insufficient width, newline), the streaming skip/regression cases, and the `shouldFill` guard — plus byte-identical regression assertions for configs without `right_template`.
- `go test ./...`, `golangci-lint run`, `fieldalignment` and `modernize` all clean.
- Verified end-to-end on Windows: `print transient --shell pwsh` emits `left + ClearAfter + DECSC + gap + right + DECRC`, `--shell zsh --eval` emits `PS1=...` + `RPROMPT=...`, and configs without `right_template` produce unchanged output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/prompt/extra.go
+++ b/src/prompt/extra.go
@@ -2,6 +2,7 @@ package prompt
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/jandedobbeleer/oh-my-posh/src/color"
 	"github.com/jandedobbeleer/oh-my-posh/src/config"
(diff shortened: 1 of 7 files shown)


src/prompt/extra.go
```diff
@@ -78,34 +79,105 @@ func (e *Engine) ExtraPrompt(promptType ExtraPromptType) string {
...
-				str += padText
-			}
+			padText, _ = e.shouldFill(prompt.Filler, consoleWidth-length-rightLength)
 		}
 	}
 
+	// for pwsh, the padding moves inside the cursor save/restore sequence
+	// when a right-aligned template is rendered, see transientPWSH
+	if e.Env.Shell() != shell.PWSH || rightLength == 0 {
+		str += padText
```
JanDeDobbeleer, line 99: should this be done prior to the transient check?

JamBalaya56562: Good catch, it should - moved the transient check up so everything below it only runs for the transient prompt, which also drops the two per-block guards. Folded into f423971.

JanDeDobbeleer: I'll merge this into next and wait for the latest winget PR to be merged as otherwise this will have annoying side-effects. Thanks!

PR #7686, opened 2026-07-16
fix(init): isolate strict init scripts

garysassano:
## Problem

Version managers can remove an old Oh My Posh installation as part of an upgrade. For example, `mise upgrade` replaces `~/.local/share/mise/installs/oh-my-posh/<version>` and removes the previous version, while an existing shell still has `_omp_executable` set to that old absolute path. The next prompt then fails with `No such file or directory` until the shell is restarted.

The existing `init --strict` option is intended to avoid this by resolving `oh-my-posh` through `PATH`, but strict and default initialization currently share the same cached script and cache key. Switching modes can therefore reuse the wrong script, and concurrent shells can overwrite each other's executable choice. The current installation and upgrade docs also do not describe the PATH-based behavior. The only existing documentation I found is a one-line mention in a 2022 release post, while the CLI help says only "run in strict mode".

## Change

This change gives strict and default initialization separate cache keys and script files, ensuring each mode consistently uses its intended executable path. It also clarifies the purpose of `--strict` in the CLI help and installation documentation.

## Validation

Validation included `go test ./...`, `go test -race ./shell`, `go vet ./shell ./cli`, and `npm run build`. I also generated default and strict Bash init scripts from the same configuration and verified that they remain separate and contain the expected executable paths.

### DIFF
--- a/src/cli/init.go
+++ b/src/cli/init.go
@@ -62,7 +62,7 @@ See the documentation to initialize your shell: https://ohmyposh.dev/docs/instal
 	}
 
 	initCmd.Flags().BoolVarP(&printOutput, "print", "p", false, "print the init script")
-	initCmd.Flags().BoolVarP(&strict, "strict", "s", false, "run in strict mode")
+	initCmd.Flags().BoolVarP(&strict, "strict", "s", false, "resolve the executable through PATH")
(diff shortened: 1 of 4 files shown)


JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @garysassano for code,doc

PR #7694, opened 2026-07-18
feat(fish): add vimode segment support

rightaditya:
### Prerequisites

### Description

Extend vimode segment support to fish. The approach is straightforward. Because fish's default vimode key bindings provide more detail, I made sure to propagate that info through to `Keymap` while still mapping them to the appropriate `Mode` so that there is sensible behaviour for existing vimode segment configurations being used on fish for the first time.


JamBalaya56562:
Hi @rightaditya, thanks for picking this up!
A heads-up so we don't duplicate effort: I implemented fish `vimode` support a while back in #7651.
It was reviewed and approved by @JanDeDobbeleer and merged — but the base was switched from `main` to `feature/template-data` right before merge, and that branch has since been removed, so the change never actually landed in `main` (fish still treats `VIMode` as a no-op there).
That's probably why it wasn't on your radar.

For reference, the two implementations differ a bit:

- **#7651** maps fish's bind modes onto the keymap names the segment already uses on ZSH/PowerShell (`default`→`vicmd`, `insert`→`viins`, `replace`/`replace_one`→`replace`, `visual` passthrough), so existing `vimode` configs behave identically across shells.
- **#7694** propagates fish's more detailed native keymaps through to `Keymap` while still mapping them to `Mode`, exposing the extra fish-specific detail at the cost of `Keymap` differing from other shells.

Both are reasonable — it mostly comes down to prioritizing cross-shell consistency vs. surfacing fish's native detail.
Happy to help land this either way: I can rebase #7651 onto `main`, or we fold the best of both into your PR. @JanDeDobbeleer what would you prefer?

JanDeDobbeleer:
@JamBalaya56562 I added your incorrectly dropped change to #7695. Sorry about that, was doing too much last week.

rightaditya:
Ah yeah, I didn't see it on `main`. Given the documentation description of `Keymap` (that it's meant to be a more raw value) I think it makes a lot of sense to surface fish's more granular state there. I'd be happy to submit a separate PR for that now that the fish vimode support has landed on `main` if that's acceptable @JanDeDobbeleer.

JanDeDobbeleer:
Go for it

PR #7707, opened 2026-07-21
test(serve): wait on records instead of wall-clock silence

ihopenre-eng:
### Prerequisites

### Description

The serve harness collected records until 500 ms passed without a new one, so every
assertion raced render latency instead of checking behaviour. Under `go test ./...` with
a cold build cache, which is what CI runs, a cycle's first record regularly arrives later
than that: `collect` returns empty and the test fails at exactly the idle mark.

Reproduced on an untouched `main` (Windows, Go 1.26.4):

```console
$ go clean -cache && go test -count=1 ./...
--- FAIL: TestServeLoop_RenderProducesIDPrefixedRecords (0.51s)
--- FAIL: TestServeLoop_AbortStopsRecordFlowThenNewRenderWorks (0.51s)
--- FAIL: TestServeLoop_WaitRenderEmitsExactlyTwoRecords (1.01s)
FAIL	github.com/jandedobbeleer/oh-my-posh/src/cli	7.817s
```

`go test ./cli/...` on its own always passes, so this only shows up in a full-suite run,
which is why it reads as a phantom when it hits a PR.

The fix collects until the records prove what the test asserts:

- a completed cycle ends at its transient record, which is the record `collect` now stops
  on rather than on a timer;
- `TestServeLoop_AbortStopsRecordFlowThenNewRenderWorks` waits for a record carrying cycle
  2's id, because an aborted cycle 1 may never reach a transient;
- the idle window still ends a collection whose condition never becomes true, and a 30 s
  bound keeps a stalled daemon a test failure rather than a hang.

No production code changes, `src/cli/serve.go` is untouched.

Verified with `go clean -cache && go test -count=1 ./...` (the failing scenario above) plus
`golangci-lint run`, `go vet ./...`, markdownlint and `vale .agents/skills`.

The gotcha is recorded in `.agents/skills/project-knowledge/references/testing.md`, per the
project-knowledge contract in `AGENTS.md`.

### DIFF
--- a/.agents/skills/project-knowledge/references/testing.md
+++ b/.agents/skills/project-knowledge/references/testing.md
@@ -72,3 +72,13 @@ for usage. Gotchas baked into it, relevant to any future pty work:
 
 - One `bufio.Scanner` per pipe, ever - a second scanner on the same pipe loses buffered data
   (use a shared reader helper in Go tests).
+- Never decide "the records are done" from wall-clock silence (verified 2026-07-21). The serve
+  harness collected until a 500 ms gap with no record, which made every assertion a race against
+  render latency: under `go test ./...` with a cold build cache (i.e. CI), the first record of a
+  cycle regularly lands later than that, so `collect` returned empty and
+  `TestServeLoop_RenderProducesIDPrefixedRecords`, `_WaitRenderEmitsExactlyTwoRecords` and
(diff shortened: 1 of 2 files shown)


JanDeDobbeleer (approved):

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @ihopenre-eng for code

This is pure magic 🪄! @holopin-bot @ihopenre-eng wizard

PR #7711, opened 2026-07-23
next

JanDeDobbeleer:
template.Render's func map exposed cmd/readFile/stat/glob/env/expandenv
to any string passed as the template argument, with no way to tell a
template the user authored in their config from one built at runtime
out of external data (filesystem names, command output, ...). That
distinction is exactly what let a malicious folder name reach cmd
(fixed for that one call site in a prior commit); nothing stopped the
same class of bug from reappearing at a future call site.

Split Render into two explicitly named functions instead: RenderTrusted
keeps the full func map and is for template text read verbatim from a
config field (segment/block/palette templates, mapped_locations keys,
folder_separator_template, ...) — every existing call site converts to
it. RenderUntrusted drops cmd/readFile/stat/glob/env/expandenv and is
for text that may contain or be composed from runtime data; pt.Path in
path.go's setStyle(), the one sink that re-renders a string composed
from raw filesystem folder names, uses it. Neither name is shorter or
more "default" than the other, so there's no ambient plain Render a
future call site could reach for without first deciding which one it
means. The parsed-template cache key includes the trust level so a
trusted and an untrusted render of identical text can never share a
cached *template.Template and its func map.

Entire-Checkpoint: 597e5a60d968

### Prerequisites

- [ ] I have read and understood the [contributing guide][CONTRIBUTING.md].
- [ ] The commit message follows the [conventional commits][cc] guidelines.

### Description

### DIFF
--- a/.agents/skills/project-knowledge/references/testing.md
+++ b/.agents/skills/project-knowledge/references/testing.md
@@ -72,3 +72,13 @@ for usage. Gotchas baked into it, relevant to any future pty work:
 
 - One `bufio.Scanner` per pipe, ever - a second scanner on the same pipe loses buffered data
   (use a shared reader helper in Go tests).
+- Never decide "the records are done" from wall-clock silence (verified 2026-07-21). The serve
+  harness collected until a 500 ms gap with no record, which made every assertion a race against
+  render latency: under `go test ./...` with a cold build cache (i.e. CI), the first record of a
+  cycle regularly lands later than that, so `collect` returned empty and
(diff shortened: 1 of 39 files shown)


ihopenre-eng:
Hi Jan, thanks for shipping the fix here. Both advisories are published now (GHSA-6xj8-qv9j-xcjq and GHSA-fwjx-9p69-h25h) but neither has a CVE assigned yet. GitHub only lets repo maintainers and owners use the "Request CVE ID" button on the advisory pages, so it has to come from your side. Could you request one for each when you get a minute? It helps distros and vulnerability scanners track the fixes. Thanks again for the quick turnaround.

PR #7720, opened 2026-07-26
feat(jujutsu): expose change ID prefix and rest

Gravifer:
## Summary

This is a follow-up to #6671, which added `change_id_min_len` support to the Jujutsu segment. That change exposed the combined shortened Change ID; this PR additionally preserves Jujutsu's native distinction between the minimum unique prefix and the characters appended to satisfy the configured minimum.

- add `.ChangeIDPrefix` for the shortest unique prefix
- add `.ChangeIDRest` for the additional displayed portion required by `change_id_min_len`
- preserve `.ChangeID` as `ChangeIDPrefix + ChangeIDRest`
- leave the default segment template and visual output unchanged

It's useful to achieve the hightlighting of the shortest unambiguous prefix, like how jj displays it:
[image](https://github.com/JanDeDobbeleer/oh-my-posh/pull/6671#issuecomment-3164992201)

## Implementation

Both components are collected atomically by the existing single `jj log` process using `ShortestIdPrefix.prefix()` and `.rest()`. A trim-safe internal delimiter preserves an empty Rest for clean working copies, and malformed output is rejected without panicking or partially populating segment state.

No additional subprocess, version probe, ANSI parsing, configuration option, schema change, or dependency is added. No unrelated Git worktree or path behavior is included.

## Testing

Added coverage for:

- unique prefixes shorter than `change_id_min_len`
- unique prefixes already meeting the minimum
- clean and changed working copies
- unchanged combined `.ChangeID`
- command failures and malformed output
- trimmed empty-Rest output
- exact command template and a single `RunCommand` invocation
- unchanged default visual template

Validated with:

- `go test ./segments -run '^TestJujutsu' -count=1`
- `go test ./... -count=1`
- `go vet ./segments`
- Docusaurus production build
- `git diff --check`

### DIFF
--- a/.agents/skills/project-knowledge/references/codebase.md
+++ b/.agents/skills/project-knowledge/references/codebase.md
@@ -40,9 +40,10 @@
 - Segment writers gob-encode only exported fields. `segments.Base.env/options` are unexported and
   MUST survive a cache restore: overlay the restored data onto the writer initialized by
   `MapSegmentWithWriter`, never replace the writer.
-- `runtime/cmd.RunWithEnv` trims leading and trailing whitespace from command output (verified
-  2026-07-26). Segment machine protocols cannot encode meaningful trailing empty fields as blank
-  lines; use a non-whitespace delimiter and parse it defensively.
+- `runtime/cmd.RunWithEnv` applies `strings.TrimSpace` to the complete command output before
+  returning it (verified 2026-07-26). If a segment encodes an empty final field as a trailing blank
+  line, that field is lost. Use a record format that retains a final non-whitespace delimiter or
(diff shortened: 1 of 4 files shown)


Gravifer:
I also have some other ideas, and I will open some more focused PRs later. Hope it won't become too spammy 😮

Gravifer:
The theme that motivated this API split is documented in [Gravifer Tokyo](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/7723). It currently uses the backward-compatible combined `.ChangeID`; after this API is released, it can style the shortest unique prefix and the minimum-length remainder independently without changing the displayed identifier.

JanDeDobbeleer (approved):

Issue #4727, opened 2024-02-29
Highlight and always show the current *Git* directory name in `Path`


stevenxxiu:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### What would you like to see changed/added?

I'm thinking of migrating from *Starship*, as many important issues, like [Powerline prompt · Issue #528 · starship/starship](https://github.com/starship/starship/issues/528), seem rather dead. I also like the *Oh My Posh* feature of conditionally showing a segment based on an environment variable.

One feature I really need, is to highlight the *Git* directory's name in the path, and to always show it, if the path is shortened.

I only found [Oh My Posh: getting whether or not in a Git repo in other segments - Stack Overflow](https://stackoverflow.com/q/77389550), which suggests to me what I want isn't currently possible.

rockyoung:
![image](https://github.com/JanDeDobbeleer/oh-my-posh/assets/1207971/edd15110-b2d2-43fd-9919-ab67d94f3262)

is this what your want? if yes, then check my segment config:
```json
{
    "type": "path",
    "style": "plain",
    "foreground": "#ff80bb",
    // can also replace the \"\" to any icon unicode string you prefer
    "template": "{{if .Segments.Git.Ref}}<#ff0000>{{.Segments.Git.RepoName}}</>{{.Path | replace .Segments.Git.Dir \"\"}}{{else}}{{.Path}}{{end}}",
    "properties": {
        "style": "full"
    }
},
```

but note the "style" property must be "full" and better not change the "folder_separator" thing, or the `replace` may not effected cause the `.Segments.Git.Dir` may not be the same format as the `.Path` which will lead the replacement to be failed!

the drawback is could not use any built-in styles except for "full", but can only implement them by yourself with [Template Logic](https://ohmyposh.dev/docs/configuration/templates#template-logic) and [Functions](https://ohmyposh.dev/docs/configuration/templates#helper-functions)

stevenxxiu:
If `MB-Lyric/` there is a *Git* repo, then yes.

Although I'd rather not display the full path every time, as it can get rather long.

In my current *Starship* config, I have the path shortened. Ideally I'd want the shortening to focus on the parts before the *Git* repo, if it gets too long. Perhaps it could be implemented with what you linked, but I'm not sure how easy that is.

<https://ohmyposh.dev/docs/segments/path> already has many shortening styles, so surely a native implementation would be a lot simpler.

[...]

PR #7721, opened 2026-07-26
fix(path): support gitdir_format in linked worktrees

Gravifer:
## Summary

- apply `gitdir_format` when the repository root is discovered through either a `.git` directory or the `.git` indirection file used by linked worktrees
- preserve the existing path formatting behavior for standard Git worktrees
- clarify the linked-worktree behavior in the path segment documentation

## Implementation

The existing parent lookup already returns the working-tree root for both `.git` directories and files. This removes the directory-only guard and continues using that returned `ParentFolder`, without adding a Git subprocess or parsing the indirection file.

## Tests

- `go test ./segments -run '^TestSplitPath$' -count=1`
- Linux cross-compile of the segments test package
- `go test ./... -count=1`
- `go vet ./segments`
- `npm run build`
- `git diff --check`

Regression cases cover linked worktrees on both Unix-style and Windows paths. This is a focused follow-up to the original `gitdir_format` feature context in #4727; it does not add main-worktree discovery or repository-aware path shortening.

### DIFF
--- a/src/segments/path.go
+++ b/src/segments/path.go
@@ -993,7 +993,8 @@ func (pt *Path) makeFolderFormatMap() map[string]string {
 
 	if gitDirFormat := pt.options.String(GitDirFormat, ""); len(gitDirFormat) != 0 {
 		dir, err := pt.env.HasParentFilePath(".git", false)
-		if err == nil && dir.IsDir {
+		if err == nil {
+			// Linked worktrees use a .git file instead of a directory.
 			// Make it consistent with the modified parent.
 			parent := pt.join(pt.replaceMappedLocations(dir.ParentFolder))
 			folderFormatMap[parent] = gitDirFormat
--- a/src/segments/path_unix_test.go
+++ b/src/segments/path_unix_test.go
@@ -698,6 +698,18 @@ var testSplitPathCases = []testSplitPathCase{
 			{Name: "d", Path: "~/c/d"},
 		},
 	},
+	{
+		Case:         "Home directory - linked git worktree",
+		Root:         "~",
+		Relative:     "c/d",
+		GOOS:         runtime.DARWIN,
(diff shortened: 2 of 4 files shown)


Gravifer:
Follow-up context: I’ve prepared a [dependent branch](https://github.com/Gravifer/JanDeDobbeller.oh-my-posh/tree/codex/git-main-worktree) based on this PR.
It adds a lazy Git template property, `.MainWorktree`, which exposes the main-worktree path only when the current repository is a linked worktree. Git is invoked only when the property is referenced; the first lookup uses `git worktree list --porcelain -z`, and the result is cached per common repository for the shell session.
I’m intentionally not opening a follow-up PR until this one is resolved. Once it lands, I’ll rebase the branch onto main and revalidate it.

Gravifer:
The linked-worktree workflow in which this issue was discovered is demonstrated by [Gravifer Tokyo](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/7723). The theme does not itself use `gitdir_format`; this is real-world context for the linked-worktree behavior that exposed the inconsistency.

JanDeDobbeleer (approved):

JanDeDobbeleer:
@all-contributors please add @Gravifer for code,doc

This is pure magic 🪄! @holopin-bot @Gravifer wizard

Gravifer:
Update: the `.MainWorktree` follow-up is now open as [#7737](https://github.com/JanDeDobbeleer/oh-my-posh/pull/7737).

The property remains lazy, so the default Git segment pays no additional process cost. When referenced in a linked worktree, it runs one `git worktree list --porcelain -z` lookup and caches the result per common repository for the shell session. The hardened branch has passed focused and full Go tests, the documentation build, and real linked-worktree probes.

PR #7729, opened 2026-07-28
feat(jujutsu): expose commit ID prefix and rest

Gravifer:
## Summary

- add `.CommitIDPrefix` and `.CommitIDRest`
- preserve `.CommitID` as their combined displayed value
- add `commit_id_min_len`
- fetch Change ID, Commit ID, and working-copy status atomically in the existing single `jj log` process
- leave the default template and existing `.ChangeID` API unchanged

## Context

This is a focused successor to #6679 and builds on the flat, backward-compatible ID presentation introduced for Change IDs in #7720.

Credit to @nemith for the original #6679 exploration and design discussion. This branch intentionally preserves that historical lineage, while replacing the old broad and now-outdated implementation with one focused contribution against current Jujutsu and Oh My Posh APIs.

## Implementation

The Jujutsu query returns a versioned, NUL-delimited frame containing:

- Change ID unique prefix
- Change ID minimum-length remainder
- Commit ID unique prefix
- Commit ID minimum-length remainder
- working-copy diff summary

The frame's magic, field count, required prefixes, and terminal sentinel are validated before any fetched state is assigned. Malformed output follows the existing command-error behavior and leaves the fields empty.

Both `Rest` properties contain only the additional characters needed to satisfy their configured minimum length. They are not suffixes of the full canonical IDs and are not independently usable identifiers.

The strict theme schema and Jujutsu documentation include `commit_id_min_len`.

## Compatibility and performance

- `.ChangeID` remains unchanged.
- `.CommitID` follows the same flat `Prefix + Rest` convention.
- The default template still renders `.ChangeID`.
- Status fetching remains gated by `fetch_status`.
- Exactly one `jj log -r @` subprocess is used.

A 30-sample clean/changed timing probe found the additional template work within normal subprocess noise:

- clean: 337.124 ms baseline, 325.490 ms with Commit ID components
- changed: 201.084 ms baseline, 202.687 ms with Commit ID components

## Tests

Added coverage for:

- empty and nonempty Commit ID remainder
- Change ID and Commit ID combined-value invariants
- clean and changed working copies
- disabled status fetching and missing-command gating
- command failure
- incorrect framing magic/count
- missing, extra, and empty required fields
- terminal-sentinel behavior across output trimming
- exact JJ template and one-command execution
- documented default minimum lengths
- unchanged default segment template

Validation completed with targeted Jujutsu tests, `go test ./... -count=1`, the Docusaurus production build, schema parsing, a real JJ 0.42.0 framing probe, and `git diff --check`.

### DIFF
--- a/src/segments/jujutsu.go
+++ b/src/segments/jujutsu.go
@@ -16,8 +16,12 @@ const (
 
 	IgnoreWorkingCopy options.Option = "ignore_working_copy"
 	ChangeIDMinLen    options.Option = "change_id_min_len"
+	CommitIDMinLen    options.Option = "commit_id_min_len"
 	FetchAhead        options.Option = "fetch_ahead_counter"
 	AheadIcon         options.Option = "ahead_icon"
+
+	jujutsuStatusFrameHeader = "OMPJJ1:5"
+	jujutsuStatusFieldCount  = 5
 )
 
 type JujutsuStatus struct {
@@ -42,6 +46,9 @@ type Jujutsu struct {
 	ChangeID       string
 	ChangeIDPrefix string
 	ChangeIDRest   string
+	CommitID       string
+	CommitIDPrefix string
+	CommitIDRest   string
 	Scm
 }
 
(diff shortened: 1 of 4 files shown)


Gravifer:
After prototyping this, I realized the JJ Commit ID largely overlaps with information already available from the Git segment in the colocated Git+JJ workflow that motivated the work. Unlike the Change ID components in #7720, this doesn’t currently enable a concrete new viable presentation (since with Git one usually uses a 7-character prefix). I’m closing the draft rather than adding public API and query surface without a sufficiently distinct use case. We can revisit it if a JJ-only use case or broader demand emerges.

PR #7737, opened 2026-07-31
feat(git): expose main worktree path

Gravifer:
### Prerequisites

### Description

Adds `.MainWorktree` to the Git segment, exposing the absolute path to the main worktree when the current repository is a linked worktree.

The property is resolved lazily: no additional Git command runs unless a template references `.MainWorktree`. On first access, it uses `git worktree list --porcelain -z`. The result is cached for the shell session using the shared Git common directory, so sibling linked worktrees reuse the same lookup and repeated template access remains free.

The implementation:

- returns an empty string in the main worktree and for bare-main repositories;
- degrades to an empty string on unsupported Git versions, command failures, or malformed output;
- requires Git 2.36 or newer for NUL-delimited porcelain output;
- supports relative linked-worktree administration paths;
- converts returned Windows paths to shell-native paths under WSL;
- restores the required Git context lazily after live segment-cache restoration;
- keeps deterministic `--data` replay runtime-free;
- does not alter the default Git template or impose a command cost on users who do not reference the property.

This follows the linked-worktree discovery corrected in #7721. The optional PowerShell helper demonstrated by [Gravifer Tokyo](https://github.com/JanDeDobbeleer/oh-my-posh/discussions/7723) is a concrete workaround for the missing property; the native implementation moves that work behind lazy template access and repository-scoped session caching.

Tests cover normal and relative linked worktrees, sibling cache reuse, repeated and unused access, WSL path conversion, live segment-cache restoration, deterministic replay, bare repositories, malformed output, command errors, and the exact Git command arguments.


Gravifer:
The remaining red `report` check appears to expose a fork-PR limitation in the new Binary Size Check workflow rather than a failure in this change. All three binary builds succeeded; the reporting step fails only when its read-only fork token attempts to create the sticky comment (`403 Resource not accessible by integration`).

A safe way to preserve sticky reports for fork PRs would be to split measurement and publication: keep the existing `pull_request` build unprivileged, then use a separate `workflow_run` workflow from the default branch to download and strictly validate the size artifacts before posting the comment. The privileged workflow would never check out or execute PR content.

I also checked this run’s API payload: its `pull_requests` array is empty, so the reporter should resolve the open PR from the run’s head repository/branch and require an exact head-SHA match rather than trusting an uploaded PR number.

Because the privileged half cannot be exercised on upstream until its workflow exists on the default branch, would a focused follow-up PR with this design be welcome, or would you prefer the simpler job-summary fallback?

PR #7760, opened 2026-08-03
fix(runtime): honor COLUMNS fallback without a TTY

### LINKED ISSUE #7759: Claude statusline: right alignment ignores a successful COLUMNS fallback on Unix
### Code of Conduct

### What happened?

When `oh-my-posh claude` renders a config containing an ordinary prompt block with `alignment: right` on Unix, Claude Code supplies the statusline JSON through stdin. The terminal width probe runs `stty size` against that pipe and fails because stdin is not a TTY.

`runtime.TerminalWidth` then successfully parse  `COLUMNS`, but returns the recovered non-zero width together with the original `stty` error. `prompt.Engine.canWriteRightBlock` rejects any width returned with an error, so the right-aligned block is omitted.

Expected behavior: when `COLUMNS` contains a valid width, it should replace the failed terminal probe and right-aligned blocks should render using that width.

Actual behavior: the valid width is cached and returned, but the stale non-TTY error suppresses the right-aligned block for that render.

### Theme

Minimal reproduction:

```yaml
version: 3
final_space: false
blocks:
  - type: prompt
    alignment: left
    segments:
      - type: text
        style: plain
        template: LEFT
  - type: prompt
    alignment: right
    segments:
      - type: text
        style: plain
        template: RIGHT
```

Run it with piped Claude statusline data and an explicit width:

```shell
printf '%s' '{"session_id":"columns-fallback-test"}' \
  | COLUMNS=80 oh-my-posh claude --plain --config ./claude-right-alignment.omp.yaml
```

Linux (WSL2)

zsh; the failure is in the `oh-my-posh claude` subprocess and is independent of interactive shell prompt initialization.

### Log output

The debug path reports the failed Unix terminal-width probe even though `COLUMNS=80` is valid. The relevant source-level result is equivalent to `(80, "inappropriate ioctl for device")`; the right-block gate requires `(80, nil)`.

e-kulikov:
### Prerequisites

### Description

Fixes #7759.

Unix statusline renderers can receive their payload over stdin, so `stty size` cannot always read terminal dimensions. A valid `COLUMNS` fallback already supplies the width, but the previous code returned that width together with the stale terminal-probe error. Right-block rendering treats any terminal-width error as fatal and omitted `alignment: right` blocks.

This change makes successful positive `COLUMNS` parsing replace the failed probe, preserves both errors when neither width source succeeds, and adds table-driven regression coverage for valid, non-numeric, zero, and negative values.

Validation:

- `go test ./runtime -run '^TestResolveTerminalWidth$' -count=1`
- `go test ./runtime ./prompt ./cli -count=1`
- `go test ./...`
- `GOOS=windows GOARCH=amd64 go build ./...`
- `modernize --fix "./..."`
- `fieldalignment --fix "./..."`
- `go mod tidy`
- `gofmt -w .`
- `golangci-lint run`
- `git diff --check`

### DIFF
--- a/.agents/skills/project-knowledge/references/terminal.md
+++ b/.agents/skills/project-knowledge/references/terminal.md
@@ -28,3 +28,10 @@
 - On Windows there is no SIGPIPE; child lifecycle management must rely on fd closure / stdin EOF.
 - On native Linux, process spawns cost 11-16ms - daemon architectures that pay off on Windows can
   be a wash there (see the bash serve revert in [bash](bash.md)).
+
+## Statusline width detection
+
+- On Unix, `terminal-dimensions` runs `stty size` against the renderer's stdin. Claude Code sends
+  statusline JSON over that stdin, so it is not a TTY. A valid `COLUMNS` fallback must replace the
+  resulting `stty` error; returning a non-zero width with the stale error makes
+  `prompt.Engine.canWriteRightBlock` reject ordinary right-aligned blocks (verified 2026-08-03).
--- a/src/runtime/terminal_unix.go
+++ b/src/runtime/terminal_unix.go
@@ -3,6 +3,7 @@
 package runtime
 
 import (
+	"errors"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


JanDeDobbeleer:
@all-contributors please add @e-kulikov for code

This is pure magic 🪄! @holopin-bot @e-kulikov wizard

PR #7781, opened 2026-08-06
fix(pwsh): update cursor_style directly inside a PSSession

### LINKED ISSUE #7780: `cursor_style` One Line Of Buffer Deleted During Mode Change Inside PSSession
### Code of Conduct

### What happened?

Not sure if this should be supported or how to handle? Or if it should be a feature. Just thought someone might find it too.

When using `Enter-PSSession -ComputerName host`, the cursor style doesn't change and upon each mode change the prompt moves up one line and erases the previous line.

So I've switched back to using, `[Console]::Write` inside my VI mode script handler directly. This does allow the cursor to change shape.

### Theme

night-owl with the `vimode` enabled and `cursor_style` config.

Windows

powershell

### Log output

```Shell
I don't think I can capture this inside a PSSession?
```

JanDeDobbeleer:
### Prerequisites

### Description

Fixes #7780.

Inside a `PSSession` (`Enter-PSSession -ComputerName host`), `cursor_style` stopped changing the cursor shape, and every vi-mode change erased the previous prompt line.

**Root cause:** vi-mode changes call `Enable-PoshVIMode`'s `ViModeChangeHandler`, which triggers a full prompt repaint via `[Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()`. Repaint geometry is computed from `ExtraPromptLineCount`/`RawUI.CursorPosition`. Inside a PSSession, PowerShell's remoting client-host proxy prepends a `[hostname]: ` prefix to the rendered prompt before painting it locally, so the repaint's geometry no longer matches the actual on-screen layout — the repaint starts one row too high and erases the previous line. The embedded DECSCUSR cursor sequence also never reaches the real terminal as raw VT through that proxy, so the cursor shape never changes either. This matches the reporter's own workaround of calling `[Console]::Write` directly, which bypasses the proxy.

**Fix:** `Enable-PoshVIMode` now detects `$PSSenderInfo` (non-null only inside a PSRemoting server session). When set, it precomputes both cursor-style sequences once (via a new lightweight `print cursor` CLI type that renders only the `cursor_style` template, added in `src/prompt/cursor.go` / wired into `src/cli/print.go`) and caches them. On each vi-mode change it then writes the cached sequence directly with `[Console]::Write` instead of calling `InvokePrompt()`, which updates the cursor shape without triggering a repaint, avoiding the broken geometry entirely. Outside of PSSession, behavior is unchanged.

### Changes

- `src/prompt/cursor.go` (new): `Engine.CursorStyle()` renders just the `cursor_style` escape sequence.
- `src/prompt/engine.go`: add `CURSOR` print-type constant.
- `src/cli/print.go`: wire up `print cursor`.
- `src/shell/scripts/omp.ps1`: `Enable-PoshVIMode` branches on `$PSSenderInfo`, caching/writing the cursor sequence directly inside a PSSession; cache is cleared on vi-mode disable.
- Tests added for the new `CursorStyle()` method and the `print cursor` CLI type.

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01QPQGxXuxvNWEEkQPXMD8jE

---
_Generated by [Claude Code](https://claude.ai/code/session_01QPQGxXuxvNWEEkQPXMD8jE)_

### DIFF
--- a/src/cli/print.go
+++ b/src/cli/print.go
@@ -42,7 +42,7 @@ func init() {
 
 func createPrintCmd() *cmdtree.Command {
 	printCmd := &cmdtree.Command{
-		Use:   "print [debug|primary|secondary|transient|transient-right|right|tooltip|valid|error|preview]",
+		Use:   "print [debug|primary|secondary|transient|transient-right|right|tooltip|valid|error|preview|cursor]",
 		Short: "Print the prompt/context",
 		Long:  "Print one of the prompts based on the location/use-case.",
 		ValidArgs: []string{
@@ -56,6 +56,7 @@ func createPrintCmd() *cmdtree.Command {
 			prompt.VALID,
 			prompt.ERROR,
 			prompt.PREVIEW,
+			prompt.CURSOR,
 		},
 		Args: NoArgsOrOneValidArg,
 		Run: func(cmd *cmdtree.Command, args []string) {
@@ -133,6 +134,8 @@ func createPrintCmd() *cmdtree.Command {
 				fmt.Print(eng.ExtraPrompt(prompt.Error))
 			case prompt.PREVIEW:
 				fmt.Print(eng.Preview())
+			case prompt.CURSOR:
+				fmt.Print(eng.CursorStyle())
 			default:
 				_ = cmd.Help()
 			}
--- a/src/cli/print_test.go
+++ b/src/cli/print_test.go
@@ -13,3 +13,9 @@ func TestPrintCommandTransientRight(t *testing.T) {
 
 	assert.Contains(t, cmd.ValidArgs, prompt.TRANSIENT_RIGHT)
 }
+
+func TestPrintCommandCursor(t *testing.T) {
+	cmd := createPrintCmd()
+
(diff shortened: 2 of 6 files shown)


JanDeDobbeleer:
Follow-up on my own initial approach: a self-review caught that gating the fix on `$PSSenderInfo` was wrong. That variable is only set inside a remote/server-side runspace, but PSReadLine's vi-mode handler always runs in the local client process (it reads raw keystrokes off the real console, independent of which runspace is current for prompt evaluation) — so the check never fired where the handler actually executes.

Pushed a correction (6e1ab5b) that checks `$Host.IsRunspacePushed` inside the handler instead — the client-side, per-invocation signal for "a PSSession is currently entered" via `Enter-PSSession`/`Exit-PSSession` (`IHostSupportsInteractiveSession`). This also lets the behavior track sessions being entered/exited mid-shell rather than being fixed at `Enable-PoshVIMode` time.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #7803, opened 2026-08-13
fix(git): interpret bare HEAD through the shared resolver and stop overwriting .Upstream with the remote list

### LINKED ISSUE #7799: Bare repositories: .Ref/.HEAD mishandle detached HEAD and reftables, and .Upstream holds the raw remote list
### Code of Conduct

### What happened?

`getBareRepoInfo()` (`src/segments/git.go:425-438`) is the entire `IsBare == true` render path, and it
has two independent defects:

1. **HEAD is misinterpreted, not just misread.** It hand-rolls a three-line HEAD parse that only
   understands `ref: refs/heads/…`. A reftables repository renders `.Ref` as `.invalid` instead of the
   real branch name; a detached bare HEAD renders the full 40-character object ID with `.Detached`
   false and `.Hash`/`.ShortHash` empty. The non-bare path already interprets all of this correctly —
   the bare path just never reuses it.
2. **`.Upstream` carries the remote list, not the upstream.** `src/segments/git.go:434` assigns
   `g.Upstream = g.getGitCommandOutput("remote")` — the newline-separated, alphabetically-sorted output
   of `git remote`. A repository with two remotes renders `.Upstream` as `origin\nupstream`, putting a
   line break in the middle of the prompt.

Expected: in a bare repository, `.Ref`/`.HEAD`/`.Detached`/`.Hash` behave identically to a non-bare
repository (reftables, detached HEAD, everything), and `.Upstream` is either the branch's real upstream
or empty — never the remote list.

### Theme

```yaml
version: 3
final_space: false
blocks:
  - type: prompt
    alignment: left
    segments:
      - type: git
        style: plain
        properties:
          fetch_bare_info: true
          fetch_upstream_icon: true
        template: 'Ref=[{{ .Ref }}] Up=[{{ .Upstream }}]'
```

```shell
# bare clone with two remotes configured, rendered from its root
oh-my-posh print primary --plain --config ./bare.omp.yaml --pwd /path/to/bare-root --shell zsh
# Ref=[] Up=[origin
# upstream]
```

Linux, macOS, Windows

zsh

### Log output

```
[DEBUG] map.go:debugf:71 → fetch_bare_info: true
[TRACE] terminal.go:FileContent(<bare>/config) - 80µs   # bare = true, two remotes
[TRACE] git.go:isBareRepo() - 198µs
```

### Note
This depends on #7798 — the `.Ref` fix here reads HEAD through the same directory that issue's fix
corrects.

e-kulikov:
Fixes #7799.

### Prerequisites

- [x] I have read and understood the contributing guide.
- [x] The commit message follows the conventional commits guidelines.
- [x] Tests for the changes have been added.
- [x] Docs have been added/updated where relevant.

### Description

`getBareRepoInfo()` hand-rolled a HEAD parse that only understood `ref: refs/heads/…` — a reftables
repository rendered `.Ref` as `.invalid`, and a detached bare HEAD rendered the full 40-character
object ID with `.Detached` false and `.Hash`/`.ShortHash` empty, even though the non-bare path already
handles all of this correctly. Separately, `.Upstream` was assigned the raw, newline-separated output
of `git remote` — with two remotes configured, that put a literal line break in the middle of the
prompt.

This PR routes bare HEAD interpretation through the same resolver the non-bare path already uses, and
replaces the `.Upstream` assignment with a real upstream/remote-selection lookup, resolving to the
branch's actual upstream (or empty, when none is configured) instead of the remote list. Most of the
directory-selection machinery this originally needed is no longer necessary, since #7798 already makes
the current checkout's git directory correct in every layout this PR touches.

Verified with new tests covering reftables, detached HEAD, single- and multi-remote bare repositories,
and the existing non-bare behaviour, confirmed unchanged.

### This is part of a stacked series

```text
next (includes #7801)
 └─ fix/git-worktree-common-dir                     → #7825
     └─ fix/git-bare-head-upstream    (this PR)     → #7799
         └─ feat/statusline-payload-pwd              → #7800
```

This branch adds one commit on top of `fix/git-worktree-common-dir` (#7825) and depends on it directly —
this PR's simplified HEAD resolution assumes that PR's directory-role fix has already landed. Until
#7825 merges, this PR's diff includes its commit too; it will then shrink to this PR's own commit.


JanDeDobbeleer:
Same with this one, we can reopen.

e-kulikov:
Recreated as #7830, based on `main` with the branch rebased and my work squashed into a single commit, as requested.

This PR closed automatically when `next` was deleted, and `gh pr reopen` can't restore it because the base branch no longer exists. Since `next` had already landed in `main`, the replacement targets `main` directly rather than waiting for `next` to come back.

Same code, no functional change: the commit's patch-id is unchanged. It still stacks on the directory-role fix, now #7828.

PR #7864, opened 2026-09-11
fix(git): update git stash count for reftable

BigBear0812:
- Update the stash count to use the git rev-list command to get the number of stashes instead of checking log/refs/stash file. This file is not used when using the reftable backend.
- Remove the old git stash test which only tested the number of items in the log/refs/stash file.
- Create a new test that simulates using git to create changes and stash them. This will better mimic real use cases and is compliant no matter the backend.

### Prerequisites

### Description

This fixes the git stash count when working with git repositories using the reftable backend. This will happen most often with larger / older repositories which might have many commits. This should be compliant no matter what backend the repository is using.


src/segments/git.go
```diff
@@ -371,12 +371,14 @@ func (g *Git) StashCount() int {
 		return g.stashCount
 	}
 
-	stashContent := g.fileContent(g.scmDir, "logs/refs/stash")
-	if stashContent == "" {
+	stashCountStr := g.getGitCommandOutput("rev-list", "--walk-reflogs", "--ignore-missing", "--count", "refs/stash")
```
JanDeDobbeleer: I want to avoid using git commands as much as possible. Yes, it fixes things, but it's too slow.

JanDeDobbeleer (changes requested):

src/segments/git.go
BigBear0812: That is totally fair. I have updated the method in the `git.go` file to use the original method whenever there is a `logs/refs/stash` file. When that is not present (such as when using reftable) only then will it use the git command. Since more repos still use the standard files back-end this will not be an issue for them.

I looked into other ways to achieve this and that would require building a binary parser for the .ref files referenced in `tables.list` and checking each existing one for stash references. It would also need to determine if each one of those stashes were still active or if they had been deleted.  While this is technically possible I doubt it would be any faster than the way git is implementing it. 

I also restored the original test for stashes to make sure that code gets tested explicitly. I modified my new test to specifically test the case of using the reftable backend.

